PR #61, opened 2016-07-13
Merging persistent shocks

mnwhite:
Finished work on the "persistent shocks" module, ConsPersistentShockModel.py.  Contains two models: an "explicit permanent income" model that is identical to ConsIndShock except that permanent income is tracked as a state variable explicitly rather than normalized out; and a "persistent shock" model that generalizes the permanent income process to an AR1 (in logs) rather than random walk. Latter model is identical to former model when PermIncCorr = 1.0.

Cubic spline consumption function and value function construction both work.  Solver works roughly equally fast whether or not there is an artificial borrowing constraint.

Item for improvement: updatePermIncGrid tries to make a grid of permanent income levels that spans the distribution of pLvl that agents will have when they are in each period.  However, it can only handle cycles=0 (infinite horizon) and cycles=1 (ordinary lifecycle), not cycles>1.  This is because when the "same" period is experienced multiple times, the distribution of pLvl will be different on each pass; unclear how to handle this.  Also should be generalized to allow variation in pLvl for "newborns" and to handle PermIncCorr (currently assumes its 1.0, generating a too-wide distribution in the persistent shock model).  Also can't handle permanent income growth in infinite horizon because there isn't a closed form solution for that.  All of these things can be fixed if we abandon analytic forms for distribution of pLvl and just simulate it and use "empirical" distribution to make a grid.


mnwhite:
This got "re-merged" due to a lost history issue.

PR #63, opened 2016-07-18
Medical shocks model

mnwhite:
This pull request adds a consumption-saving-medical care model in the new module ConsMedModel.py.  Utility is the sum of two CRRA terms (with different coefficients) for consumption and medical care, and there is an iid multiplicative shock to utility from medical care each period.  As currently calibrated in the parameters file, most "medical need" shocks are extremely close to zero, but the upper tail is very high.

Solver is demonstrated on an infinite horizon problem, and (as in other models), can use cubic spline interpolation for the policy function and can construct the value function if requested; arbitrary borrowing constraints are also allowed.  Unlike other models, cubic spline interpolation does *not* enable the user to use a much smaller number of gridpoints and get an accurate solution-- there is slight wiggliness in the cFunc (but not the xFunc) when the number of gridpoints is small.  This might be due to the convexity-concavity of the cFunc.  On the upside, adding cubic spline interpolation (holding fixed the number of gridpoints) does not increase solution time by that much, relative to other models.

There are two remaining areas for improvement (along with robustness testing).  First, *none* of the functions currently have an extrapolation-bounding function, but I believe these exist in this model.  I *think* the xFunc eventually approaches a line emanating from -hLvl(pLvl) with a slope that depends on MedShk; the slope is the same as standard MPCmin when MedShk=0 and decreases in MedShk.  For the cFromxFunc, it should approach c = x, as consumption dominates medical care when expenditure is very large (assuming CRRA < CRRAmed).  Unsure about vFunc.

Second, the bottoms of functions are wacky in this model.  Ordinarily, cFunc, (pseudo inverse) vPfunc, and (pseudo inverse) vFunc all become approach linear functions as mLvl --> mLvlMin.  With two goods, however, medical care dominates consumption when expenditure is very small, so MPC --> 0 as mLvl --> mLvlMin, while the MPM has some  positive minimum.  This generates some weirdness in vFunc and vPfunc at the very bottom, so a new approach for the bottom segment (based on Med rather than cLvl) might be warranted.

Finally, the solver probably doesn't work if CRRAmed >= CRRA, but that would be an unrealistic parameterization.


dclow:
Two immediate issues:
1) This branch has conflicts with master about documentation and interpolation
2) There are also several lines that go over (sometimes well over) 100 characters. Since this is mentioned in our style guidelines, I think it's something we should stick to when committing and merging ourselves, when possible.  Can you fix these lines?

I'm going over the code for more substantive issues now

mnwhite:
Reject changes from my branch re: documentation; accept interpolation changes from this branch.  HARKinterpolation.py is behind master by 1 commit on this branch, but that commit was just deleting BicubicInterp... which is also in this branch.

I'll look for very long lines and fix them (next week?), but I strongly dislike the 100 character rule.

mnwhite:
^^ ...which is also *deleted* in this branch.

dclow:
Thanks Matt.  Tiny question: you mentioned cubic spline interpolation works in your pull request.  But in-line documentation says it does not.  I assume the in-line documentation is wrong?

mnwhite:
Shoot, I thought I removed that. Will fix shortly.

dclow:
No problem, I can fix it (assuming it just involves deleting that line the documentation)

mnwhite:
Yeah, it might be repeated in a few places.  If you ctrl+F for "cubic" you
should find all possible mentions.  There is one spot where I mention
future implementation of bicubic interpolation (of the x --> c function),
but I think that's safe to leave.

     -- mnw

dclow:
Merged in pull request, with a few very minor changes.

I will add some basic unit tests for this model after I merge in my changes to the Testing folder.

PR #64, opened 2016-07-19
Update HARKmanual.tex

JohnYeabinMoon:
I modified some. Please check it:

1. I think the simple description of CRRAutility is needed; 
2. add the definition of attributes
3. Examples in chapter 3.4 are not working. I guess Nagents are missing both in MyConsumer and OtherConsumer.


mnwhite:
Yes.  I meant to fix this in commit e035312 (following an email exchange with you), but forgot to include Nagents.  Merging now.

PR #66, opened 2016-07-21
Update HARKmanual.tex

JohnYeabinMoon:
I add
1. a short description of dir() method, which shows all the attributes. 
2. example code generating the plot for the perfect foresight consumption function 

I have a question:
As far as I know, 'instance' and 'object' are synonymous in Python. In the manual both terms are used. Do we use the terms interchangeably? Or, does each have specific purpose?

### DIFF
--- a/Documentation/HARKmanual.tex
+++ b/Documentation/HARKmanual.tex
@@ -280,7 +280,7 @@ The \texttt{ConsIndShockModel} module defines the class \texttt{PerfForesightCon
 
 \vspace{0.25cm}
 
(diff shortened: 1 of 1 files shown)


mnwhite:
I used those words mostly interchangeably in the manual and in the inline documentation, but with two thoughts in mind:

1) I took care to use the word "instance" early on to differentiate the concept of the class itself versus an object of that class.  The plain English meaning of "instance" is helpful here.

2) I might sometimes use the word object to mean something more general than "a Python object" in the programming sense.  I.e. it could be just a float or a function or a None; it's a "something".

dclow:
I think, to be technical, every instance is an object but not every object is an instance.  A Python class is an object (I think everything in Python is an object.  @mnwhite , floats, functions, and None are all objects in Python -- right?)  But a Python class is not generally an instance.

More usefully, I think the two words should be used in different contexts.  "Instance" should really only be used when you are talking about a class, and want to talk about a Python object that is a member of that class.  "Object" should be used as shorthand for "something in Python."

mnwhite:
I think we agree here?  If the manual contains falsehoods or confuses meanings, we should change it.  But I think I followed David's description when writing it.

JohnYeabinMoon:
Thanks for clarifications. Good guideline, David. I will read the manual again and let you know guys if the usage were not clear.

PR #67, opened 2016-07-21
Update HARKcore.py

JohnYeabinMoon:
I guess I found one minor typo. Please check


dclow:
Good catch, thanks!

mnwhite:
David: Did you mean to close this, or to accept the merge?

Also, I think "to" or "from" both work in this case, with slightly different meanings.  We do pass "the solution to the period t+1 problem" to "the problem for period t".

JohnYeabinMoon:
Oh, I see. Thank you!

PR #68, opened 2016-07-21
Update ConsumptionSavingModels.tex

JohnYeabinMoon:
Correct me if I am wrong. 
PerfForesightConsumerType needs six parameters to use solve method here. Nagent is missing parameters here.  Without Nagent, error message shows up. If it were true, did we need to put Nagent in the first table?

Sorry for keep questioning on minor details.

### DIFF
--- a/Documentation/ConsumptionSavingModels.tex
+++ b/Documentation/ConsumptionSavingModels.tex
@@ -48,7 +48,7 @@ $\Gamma$ & Permanent income growth factor & \texttt{PermGroFac}
 \end{tabular}
 \end{table}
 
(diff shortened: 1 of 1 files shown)


mnwhite:
You are correct that this needs fixing; please add Nagents to the table (as the bottom entry).  Further, the end of the sentence you changed should read:

"...must have in order to create an instance and use its \texttt{solve} method."

Technically, there are only five parameters needed to use the solve() method.  The sixth, Nagents, is irrelevant when solving the model and matters only when simulating.  However, Nagents is used by the __init__ method to set a_init and p_init, so it must be passed to *make* an instance, but not to *solve* it.  You can add a footnote to this effect if you want.

JohnYeabinMoon:
Ok. I think we need to add some explanations on this. I have a little hazy understanding on solve method. I will review it and let you know!

PR #70, opened 2016-07-28
Hark Demo

dclow:
Created a very short and simple HARK demo, with *extremely* detailed comments.  It is meant to be a gentle introduction to HARK for someone who doesn't know much about Python or structural modeling.   The demo is just one file, "Demo.py", located in HARK/ConsumptionSaving, that uses HARK to explore the differences between the MPC out of income and the MPC out of credit.


mnwhite:
This is very good.  Merging now, then making 3 tiny changes.

PR #71, opened 2016-07-31
Added verbose option to AgentType.solve

mnwhite:
The solve method of AgentType now has an optional argument named verbose
(default False).  When True, solve displays progress of the solution
method (timing, convergence progress) after each cycle of periods.  Also
eliminated AgentType.isSameThing (now incorporated into solveAgent),
which Alex in particular hated.

New option is useful for testing infinite horizon models that a) take a
while to solve and/or b) don't necessarily converge because it's still
in development.  I often found myself adding lines like this anyway,
then deleting them before a commit; this makes testing easier.

### DIFF
--- a/HARKcore.py
+++ b/HARKcore.py
@@ -10,6 +10,7 @@ problem by finding a general equilibrium dynamic rule.
 from HARKutilities import getArgNames, NullFunc
 from copy import deepcopy
 import numpy as np
+from time import clock
   
 def distanceMetric(thing_A,thing_B):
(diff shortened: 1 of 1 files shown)


dclow:
Looks great!  Sorry this took me so long

Only minor comment is since (for example) 3 evaluates to True, the user can pass verbose = 3, and progress will be printed rather some kind of error being thrown.  But I think this is the behavior we want so I'm merging this in

PR #72, opened 2016-08-02
Update Demo.py

JohnYeabinMoon:
This is extremely helpful. Thank you guys. I corrected minor details. Please disregard if you intentionally used 'credit_change' in 'FirstDiffMPC_Income' function.

### DIFF
--- a/ConsumptionSaving/Demo.py
+++ b/ConsumptionSaving/Demo.py
@@ -149,11 +149,15 @@ Now that we have the solutions to the 2 different problems, we can compare them
 
 ## First, declare useful functions to plot later
 
+# Declare how much we want to increase income by
+income_change =  .001
+
+
 def FirstDiffMPC_Income(x):
     # Approximate the MPC out of income by giving the agent a tiny bit more income,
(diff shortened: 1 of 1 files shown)


dclow:
I actually did mean to use credit change there.  My intent was to change income by the same amount as we change credit, so that any differences between the two MPC approximations are (hopefully) due to actual differences in the theoretical MPCs, rather than in the quality of the approximation or anything.

I agree this is confusing though.  What if we changed 

# Declare how much we want to increase income by
income_change = .001

 to

# Declare how much we want to increase income by
# Change income by the same amount we change credit, so that the two MPC 
# approximations are comparable
income_change = credit_change 

?

dclow:
Whoops, I put in hashtags for Python comments.  But github.com interpreted that as a Markdown command to make the font bold... so just imagine the bold words are Python comments

JohnYeabinMoon:
It makes more sense to me!

dclow:
I don't have write access to this PR so I can't make changes to it myself.

I've made the changes we agreed to and issued a new PR, so I'm closing this one

PR #76, opened 2016-08-05
Incorporated Johns comments about income_change

dclow:
Tiny change to the HARKDemo implementing the change we agreed to in John's PR for the demo.

I'm issuing this as a new pull request because I do not have write access to John's repo, so I cannot change anything in the original PR

Assigning this to Matt because I can't assign it to John (I think because he doesn't have write access?) 
Should be very quick to merge in, I think....


mnwhite:
Sorry about delay, was away on vacation.  Merging now.

Quick thought / suggestion to this: We could also add a comparison of first differences MPC (wrt transitory income or market resources) vs analytical MPC using cFunc.derivative().  Not sure if that's relevant, but it sort of points to the idea that HARK has stuff like that.

PR #77, opened 2016-08-15
Legend in figures produced by plotFuncs

dclow:
Added a new argument to plotFuncs in HARKutilities.py so that keywords can be passed to plt.legend, if desired.  Also, not coincidentally, added legends to the figures produced by ConsumptionSaving/Demo.py


mnwhite:
I did not see this pull request.  My apologies.

PR #78, opened 2016-08-26
Added class ConstantFunction

mnwhite:
Trivial addition to HARKinterpolation.py: ConstantFunction.  For representing functions that return the same value no matter what the
input (or how many inputs).  Yes, this is actually useful.

### DIFF
--- a/HARKinterpolation.py
+++ b/HARKinterpolation.py
@@ -537,7 +537,13 @@ class HARKinterpolator4D(HARKobject):
         
 class ConstantFunction(HARKobject):
     '''
-    A class for representing trivial functions that return the same real output for any input.
+    A class for representing trivial functions that return the same real output for any input.  This
+    is convenient for models where an object might be a (non-trivial) function, but in some variations
+    that object is just a constant number.  Rather than needing to make a (Bi/Tri/Quad)-
+    LinearInterpolation with trivial state grids and the same f_value in every entry, ConstantFunction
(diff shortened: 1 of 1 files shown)


dclow:
Two minor comments:

1) I certainly believe you that this is a useful class, but it's hard for me to see that myself.  Can you include in the inline documentation a reason or two why this would be useful?

2) Because derivativeW/derivativeX/derivativeY/derivativeZ/derivativeXX are all the same, rather than all defining them explicitly to be the same thing, I would prefer:

self.derivativeW = self.derivative
self.derivativeX  = self.derivative
self.derivativeY  = self.derivative, etc.

This makes it far more explicit that all the functions do the same thing.  It also makes testing / debugging easier, etc, and all the other things that typically come from avoiding the duplication of code.

Do you agree?  If so, can you make this change?

mnwhite:
Yes, I can make these changes.  It's mostly used as a convenience feature
so that you don't have to define a trivial N-dimensional multilinear
interpolation just to make a function that always returns the same value
(and makes it faster to evaluate).

This comes up in my insurance equilibrium model.  You can imagine that the
(effective) premium for a contract might depend on a consumer's permanent
income (e.g. ACA subsidies) and/or market resources (in a "perfect
individual rating" counterfactual), so the solver/simulator needs to handle
premiums that vary in this way.  However, in many cases the premium is just
a single number, and it's much easier to specify premium =
ConstantFunction(5.2) than to make a bilinear interpolation that's constant.

PR #79, opened 2016-09-02
LivPrb can vary across Markov states

mnwhite:
Updated the Markov model to allow the survival probability LivPrb to
vary between discrete Markov states.  This is useful for having states
represent discrete health levels (so that someone in poor health is more
likely to die than someone in good health).

Moreover, it's a back-door route to do the Krussell-Smith model of
infinitely lived dynastic agents who occasionally change their discount
factor due to a generational change (DiscFac --> 1.0, LivPrb substitutes
for DiscFac).  Must turn off mortality in simulation for this to work.


dclow:
I think eventually this needs a bequest motive added in, but otherwise looks good

PR #82, opened 2016-09-24
Demos for fed mod

llorracc-git:
Changed some variable names and methods

### DIFF
--- a/ConsumptionSaving/China.py
+++ b/ConsumptionSaving/China.py
@@ -1,0 +1,28 @@
+# -*- coding: utf-8 -*-
+"""
+Created on Sat Sep 24 15:35:35 2016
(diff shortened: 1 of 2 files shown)


mnwhite:
1) Should these demos be in /ConsumptionSaving, or in a new folder called /DemosForFed?  This is an *application* of our models for specific purpose(s), not model files.  I think having a file called China.py in this folder is a bit confusing.

2) Should we hold off on merging this until China.py is complete?  As is, we have a Python module that breaks the first line of "code"... because the code is just placeholder notes!

dclow:
I've changed the code a bit since this PR was issued, so I've updated the code to reflect the changes in this PR manually

PR #86, opened 2016-10-12
Robust hark parallel

dclow:
The demos for the Fed Board imported cstwMPC, which imported HARKParallel, which would not work on any machine that did not have joblib and dill installed.

We want the demos to work for anyone, so I modified HARKParallel to _try_ to import joblib and dill.  If it can't, it will instead define the functions it tried to import to raise appropriate error messages


mnwhite:
This looks good.  I really hope joblib and dill become part of standard Anaconda in the future.  The stuff in HARKparallel really ought to be in HARKcore and HARKestimation.

PR #87, opened 2016-10-12
Added an input check to preSolve in HARKcore to verify that elements …

dclow:
…of time_vary are lists.  Modified ConsIndShockModel.py so that preSolve would do this input check in addition to updateSolutionTerminal(), rather than overwriting it.  Also, added much more input checking to ConsMarkovModel to make sure arrays and lists are the right shape and length.


mnwhite:
Modifying some of your comments in a moment.  You say (e.g.) that PermGroFac_list should be a list of arrays when passed to the init method of MarkovSolver, but that's incorrect.  The attribute PermGroFac of a MarkovConsumerType is a list of arrays, but *only one array* is passed to init a MarkovSolver.

PR #89, opened 2016-10-16
Simulation overhaul

mnwhite:
This is a big one.  The simulation routines for all models (other than FashionVictim) have been revised to use a default/universal simulator that's part of AgentType.  To use the simulator, an AgentType subclass must define a field called poststate_vars that names the variables that represent "post-decision state variables" during the simulation.  Mathematically, these are a minimum-dimensional sufficient statistic for the agent's state at the _end_ of a period or _just before_ a period begins.  E.g. aNrmNow and pLvlNow for the IndShockConsumerType.

Generally, simulating a period follows the steps simMortality, getShocks, getStates, get Controls, getPostStates, [advance time].  Subclasses overwrite these methods, except simMortality (overwrite simDeath and simBirth instead).  The simulator is an overlapping generations simulator: it correctly handles simulated agents being at different points in their cycle; agents who die are replaced by newborn agents.  Variables like cNrmNow are always of length AgentCount and represent the _current / most recent values_ for that variable.

A "history" of a simulation variable is only created if that variable is named in that instance's track_vars attribute.  In this sense, an AgentType now acts more like a Market when simulated.  If the string "varname" is an element of track_vars, then the attribute varname_hist will be created by the initializeSim method and filled in during simulation.

Shocks are no longer pre-specified by methods like makeIncShkHist (etc), but instead drawn on the fly.  In the near future, we should restore the ability of the various agent types to make and use pre-specified shock histories.

Some "classic" variable names have been changed: sim_periods is now T_sim, Nagents is AgentCount, T_total is T_cycle, etc.  More details can be found in my email to the HARKteam from September 2 ("Impending changes to...").

All files have been updated to use the new simulation routines (including the /DemosForFed stuff), and everything seems to run correctly.


mnwhite:
PS: Additions to the HARKmanual are warranted after this gets included in master.

dclow:
Looks great.  Very clean and user-friendly!  Should be easy to build off too.

I have a few minor comments, but one major one I want to clear up first.  There are a bunch of lines like this one from ConsIndShockModel.py:

DiePrb = DiePrb_by_t_cycle[self.t_cycle-1] # Time has already advanced, so look back one

My question is: has time already advanced??  simOnePeriod in HARKcore is defined as follows:

    def simOnePeriod(self):
        '''
        Simulates one period for this type.  Calls the methods simMortality(), getShocks(),
        getStates(), getControls(), and getPostStates(); these should be defined for model subclasses.
        
        Parameters
        ----------
        None
        
        Returns
        -------
        None
        '''

        self.simMortality()
        self.getShocks()
        self.getStates()
        self.getControls()
        self.getPostStates()
        
        # Advance time for all agents
        self.t_age = self.t_age + 1 # Age all consumers by one period
        self.t_cycle = self.t_cycle + 1 # Age all consumers within their cycle
        self.t_cycle[self.t_cycle == self.T_cycle] = 0 # Resetting to zero for those who have reached the end

so it seems time does not advance until after the call to (e.g.) self.simMortality().  I'm pretty sure I must be missing something here.  But e.g. in the first simulated period of the LifecycleExample in ConsIndShockModel.py, all the simulated agents have DiePrb = .9 because they all have self.t_cycle-1 == -1... is that the behavior we want?

mnwhite:
Here's what's happening.  At the very beginning of a simulation, agents are created using initializeSim (which calls simBirth, requesting a newborn agent for every index); this can be overridden so that the agents at the start of a simulation aren't like newborns.  At the very beginning of the first simulated period, simMortality calls simDeath, which looks at LivPrb[-1] for all agents.  As you say, this is the survival probability at the end of the *last non-terminal period*, which should be irrelevant to simulated newborns.

BUT: All that's happening is that a (1.0-LivPrb[-1]) proportion of initial newborns are replaced with other newborns!  That is, in the very first simulation period, we re-draw initial states of some newborns; this is completely harmless in most cases.  The only time it could affect anything is when initializeSim overwrites the initial newborns with some other distribution of states, so that initial newborns are different from newborns created at any other time.  Even then, death/replacement probability is often quite low, so moving (say) 1-2% away from the intended initial distribution is largely irrelevant... particularly given that we often throw out the first X periods of a simulation to let things settle in.

However, (1.0-LivPrb[-1]) can be quite large in some models, like those that explicitly model the lifecycle (so it might be the death probability for a 119 year old).  If initializeSim overwrites initial newborns, won't this lead to a large deviation from the intended initial distribution of states?  No.  In lifecycle models, you *don't* want an initial population of only newborns, because it takes a *lot* of simulated periods to smooth out the age distribution and get it close to the long run steady state age distribution.  Instead, the initial population is set to have *the steady state distribution*, which can be calculated as a (normalized) eigenvector of the "age transition matrix".  Thus there will be relatively few initial agents who have t_cycle=0 and thus might get re-drawn using a high death probability, so we're back to the realm of "mostly harmless".

See cstwMPCnew.calcStationaryAgeDstn for more.  This function will be slightly revised and moved to be a method of PerfForesightConsumerType in the near future.

dclow:
Can you tell me where the initial population is set to have the steady state distribution of age?
This is what I can find for simBirth, in PerfForesightConsumerType (which does not seem to be overwritten in IndShockConsumerType)

    def simBirth(self,which_agents):
        '''
        Makes new consumers for the given indices.  Initialized variables include aNrm and pLvl, as
        well as time variables t_age and t_cycle.  Normalized assets and permanent income levels
        are drawn from lognormal distributions given by aNrmInitMean and aNrmInitStd (etc).
        
        Parameters
        ----------
        which_agents : np.array(Bool)
            Boolean array of size self.AgentCount indicating which agents should be "born".
        
        Returns
        -------
        None
        '''
        # Get and store states for newly born agents
        N = np.sum(which_agents) # Number of new consumers to make
        self.aNrmNow[which_agents] = drawLognormal(N,mu=self.aNrmInitMean,
                                                   sigma=self.aNrmInitStd,
                                                   seed=self.RNG.randint(0,2**31-1))

        pLvlInitMeanNow = self.pLvlInitMean + np.log(self.PermGroFacAgg**self.t_sim) # Account for newer cohorts having higher permanent income

        self.pLvlNow[which_agents] = drawLognormal(N,mu=pLvlInitMeanNow,
                                                   sigma=self.pLvlInitStd,
                                                   seed=self.RNG.randint(0,2**31-1))
        self.t_age[which_agents]   = 0 # How many periods since each agent was born
        self.t_cycle[which_agents] = 0 # Which period of the cycle each agent is currently in

mnwhite:
It's not.  This functionality is currently only in cstwMPCnew, and I plan to move it over lately.  You *can* simulate a lifecycle model starting from all newborns, it will just take a long time to even out the age distribution.

mnwhite:
Lately --> later.

dclow:
Gotchya.

Should I think of LivPrb as the _end of period_ survival probability?  Which is why, when simulating mortality at the beginning of the period, using _last_ period's LivPrb is the correct thing to do?

mnwhite:
Yes, that's the way it ends up being specified in HARK.  When I'm solving
the problem for period t, I need to think about the survival probability
*between* period t and t+1.  That's the information the solver needs to
solve the one-period problem, so LivPrb[t] has to be an end-of-period
survival probability for period t in order for the correct inputs to be
passed to the solver.

When I'm simulating period t+1, the very first thing I do is to find out
who died between period t and period t+1, so I look at the end-of-period
survival probability for period t.

mnwhite:
The exact ordering of when a period "ends" and the next one "begins" is
tricky and not obvious.  It's strange for me to say that agents die
"between" periods t and t+1.  I mean, isn't every moment in *some* period.
This is why we need to be specific about the class of models that can use
this "default simulator", and the assumptions we're making about
chicken-and-egg things.  Future HARKmanual update!

edmundcrawley:
Hi Matt,
You say, "In the near future, we should restore the ability of the various agent types to make and use pre-specified shock histories." I have some code for impulse-response graphs that relies on the ability to do this and which doesn't work with the simulation overhaul. Would you like me to take a look at how best to restore this functionality?
Related to this, I think it would be useful to have some impulse-response functionality in HARK. Has there been thought put into this? I have some code that could form a start for this (although it doesn't work now...)

mnwhite:
My initial thought is that there's a short term and a long term fix to restore pre-specified shock histories.  First, a quick delineation of two meanings of "pre-specified shock histories".

1) The shock history is randomly determined before the simulation is run.  That is, we draw all of the shocks for all periods, so that when we simulate we "draw" shocks by just pulling up row t_sim of a stored shock array.

2) The shock history is manually set before the simulation is run.  That is, we are not interested in a random set of shocks, but in some specific set/history, as in an impulse response analysis.

The short term fix for you is to simply make a subclass of the consumer type you're using and overwrite the getShocks method as something like:

`def getShocks(self):
    t = self.t_sim
    self.PermShkNow = self.PermShkNow_hist[t,:]
    self.TranShkNow = self.TranShkNow_hist[t,:]
    `
And then specify the attributes PermShkNow_hist and TranShkNow_hist to be T_sim x AgentCount arrays of whatever shocks you want.  Maybe these are constructed through some methods or just manually constructed for a specific purpose.  Everything should work normally for you.  This is why I called the method getShocks rather than drawShocks-- maybe we draw them, maybe we just look them up in an array.

The longer term fix is to incorporate this as part of AgentType itself, which will both speed up the simulation code and enable things like you're doing.  Here's a spitball of a design:

- Subclasses of AgentType have an attribute called shock_vars, which contains a list of strings naming the attributes that represent stochastic shocks during simulation.  E.g. shock_vars = ['PermShkNow', 'TranShkNow']

- AgentType has a method called makeShockHistory() that initializes the simulation, then runs a loop on a subset of simOnePeriod that only runs simMortality and getShocks each period, as well as advancing "time variables" t_age and t_cycle.  It generates the history of the variables named in shock_vars and stores them in X_hist attributes.  At the end, it sets the attribute read_shocks to True (automatically False when AgentType is created).

- The simOnePeriod method gets a simple conditional: if self.read_shocks, then self.readShocks(), else self.getShocks().  Thus the simulator will draw the shocks if a history has not been created, or simply read them from the X_hist attributes if it has.  The readShocks() method would live in AgentType and be simply:

`
def readShocks(self):
    t = self.t_sim
    for shock_name in self.shock_vars:
        setattr(self,shock_name) = getattr(self,shock_name + '_hist')[t,:]
`

Then you could make other methods that specified the shock history in some non-random way (like for impulse response analysis).  There are other changes I want to make to the new simulator that will speed up other aspects, but those are another topic.

mnwhite:
Gah, carriage returns don't work in the code block, I guess.

dclow:
Matt -- I'd like to see if we can improve this timing convention, for three reasons:
1) I find indexing shocks simulated in period t by (t-1) to be highly counterintuitive
2) The workaround for newborns in getShocks for IndShockConsumerType is kludgy, and I'm sure this kludge will have to be repeated elsewhere if we don't do something about it
3) You have a convincing argument above that simulating mortality for newborns is only slightly wrong.  But I'd rather the code be right than only slightly wrong.

My first instinct is to think of shocks as happening at the end of the period.  We'd stop indexing shocks simulated in period t by (t-1), and we'd change simOnePeriod to:
    def simOnePeriod(self):
        self.getStates()
        self.getControls()
        self.getPostStates()
        self.simMortality()
        self.getShocks()

This has the added benefit of not simulating shocks for newborns.

This seems like it would work, perhaps with a few modifications.  I expect you'll hate this suggestion, though I'm not sure I can predict why.  I think it might be because, for reasons I don't understand, it really is critical to think of these shocks as happening *between* periods.  If that is the case, I'd rather be much more explicit about it.  For example, we could change simOnePeriod to:


    def simOnePeriod(self):
        self.getStates()
        self.getControls()
        self.getPostStates()
        
        # Advance time for all agents
        self.t_age = self.t_age + 1 # Age all consumers by one period
        self.t_cycle = self.t_cycle + 1 # Age all consumers within their cycle
        self.t_cycle[self.t_cycle == self.T_cycle] = 0 # Resetting to zero for those who have reached the end

We'd create a new method, call it simBtwnPeriods:
    def simBtwnPeriods(self):
        self.simMortality()
        self.getShocks()

and we'd change simulate() to:
    def simulate(self,sim_periods=None):

        orig_time = self.time_flow
        self.timeFwd()
        if sim_periods is None:
            sim_periods = self.T_sim
            
        for t in range(sim_periods):
            self.simBtwnPeriods()
            self.simOnePeriod()

If we really want to keep simulating shocks for newborns.  If we don't (I don't), we could change the last bit to:

        for t in range(sim_periods):
            if t > 0:
                self.simBtwnPeriods()
            self.simOnePeriod()

Thoughts?

mnwhite:
This doesn't work, or rather it needs more specification.  First, shocks in period t *are not* indexed by t-1 in HARK.  You might be referring to lines 1766 and 1767 in ConsIndShockModel, which refer to IncomeDstn[t-1] and PermGroFac[t-1].  Remember, those are *next period* income distributions, the information needed to solve the period.  So they refer to period t-1+1=t.  Efforts to fix the time indexing for simulating will mangle it or make it really confusing for solving.

Second, you need to simulate shocks for newborns, or at least give them *some* value of the shock variables, otherwise getStates will automatically fail.  Note lines 1786-7, where I have commented out that newborns get shocks of 1.0.

Third, your system doesn't match the indexing notation used in most/all models.  See e.g. this old version of cstwMPC: http://www.econ2.jhu.edu/people/ccarroll/papers/cstwMPC.pdf , page 10 (or our models PDF, if I wrote it correctly).  The shocks that determine income in period t are themselves indexed by t, not t-1.  In the system you're describing, the shocks that we call \psi_t would actually be created in period t-1 (and stored in that row of a history array).

Fourth, the system of killing and replacing agents *before* we record their outcomes for that period is destroying information.  Those agents took actions and achieved some states, but we just throw it out!  This doesn't match how our current applications work.

I don't think it's important to think of shocks as happening "between" periods.  I think it's fine to have shocks happen at the beginning of the period, which is the way the models are written.  There are no perfect answers to this issue, which is why I think it's important to be specific about what we're doing and what the order of events is.

Now that I look at getShocks for the Markov model, I'm wondering if there's an off-by-one error for which Markov array is used.  I haven't actually tested a time-varying Markov array yet.

dclow:
Okay, here's an idea that might make us both happy (and maybe Edmund too?).

We add some method to AgentType, call it createSimShocks(), or something.  createSimShocks() would create the shock processes for simulations, which would be distinct from the shock processes for solving.  For example, for IndShockConsumerType, createSimShocks() would create an attribute self.IncomeDstnSim with self.IncomeDstnSim[t] = self.IncomeDstn[t-1] for t>=1, and self.IncomeDstnSim[0] explicitly set to something we think is appropriate (I'd prefer getting the shocks to 1, you might prefer doing something else)

This has three big advantages:
(1) During simulation, you would index shocks in period t by t, not t-1 (which it seems to me is exactly what lines in 1766 and 1767 in ConsIndShockModel do, even though I agree it is the correct thing to do).
(2) It would explicitly deal with newborns, and let them be handled in different ways for different purposes as whoever codes it sees fit
(3) For the many possible cases in which you might want perceived and actual shock probabilities to differ (e.g. the Chinese_Growth demo, or Edmund's impulse response functions), this would provide a quick, easy, and explicit way to do it

It's worth emphasizing that this way of coding it up would be very specific about what is being done and why, and we both agree we need to be very specific with this code

mnwhite:
I strongly disagree that the shocks for period t are indexed by t-1 in my branch.  Just because it pulls information from IncomeDstn[t-1] does not mean its indexed by t-1.  If you add "TranShkNow" to track_vars and simulate, you will see that the transitory shocks for period t are stored in row t of TranShkNow_hist.

I don't totally understand your proposal because you didn't specify what goes in IncomeDstnSim.  From the text, I *think* you meant that each element of IncomeDstnSim would have a discrete distribution with probabilities, permanent shocks, and transitory shocks, just like IncomeDstn.  Then the simulator would get to draw from IncomeDstnSim[t] rather than IncomeDstn[t-1] when distributing shocks to agents.  If that's what you meant, I have two issues with this:

1) IncomeDstnSim[t] would have to be constructed by looking at IncomeDstn[t-1], so all you're doing is shifting this (allegedly) problematic indexing to another method, while adding a layer of complexity.  I think this issue is better addressed through a very clear comment block near line 1766 that explains why we're drawing from IncomeDstn[t-1].

2) It doesn't work.  The t index in your proposed IncomeDstnSim[t] *seems* to refer to a period in the cycle, i.e. t_cycle?  So if there are T_cycle periods in the cycle, then IncomeDstnSim should have T_cycle elements (or maybe T_cycle+1 if there is a terminal period).  In an infinite horizon model with the same one period cycle repeated over and over, IncomeDstnSim would only have 1 element, and by the way you've described it would be a degenerate distribution with income shocks of 1; that is, agents would *never* experience income shocks because you specified IncomeDstnSim[0] to be degenerate to account for newborns.  Likewise, a model with (say) 4 periods in the cycle with cycles=3 would not be treated correctly.  Conditional on survival, agents in that model live the period corresponding to t_cycle=0 three times; your system would assign them shocks of 1 all three times, when you actually only wanted to give them in their very first period.

I could instead interpret the t IncomeDstnSim[t] as referring to how many periods the simulated agent has been alive (i.e. periods since last mortality/replacement event), that is t_age.  Then IncomeDstnSim would have to have T_age elements, but this is problematic for several reasons.  First, we never mandate that T_age is finite; we could allow an infinite horizon model where agents really are allowed to live forever if they keep getting lucky.  IncomeDstnSim would thus be infinitely long.  Second, it would be incredibly computationally inefficient when T_age is large: we would be looking up the income distribution separately for every t_age, rather than every t_cycle, ignoring that we know that (potentially) every t_age should be treated the same except for 0.  This also doesn't work with impulse response-type simulations.

If you instead meant for the t in IncomeDstnSim[t] to refer to t_sim, the absolute measure of time in the simulation (shared across all agents of this type, so its a float, not an array), that also doesn't work.  This would enable impulse response analysis, but would be completely wrong in many cases, as it completely ignores that simulated agents might have different ages or be at different points in the cycle of periods in the same absolute simulated period t_sim.

I really think the ideal solution here is to be really explicit about why we have to draw shocks from IncomeDstn[t-1].  Moreover, I think we should just set shocks for newborns to be 1, as I currently have commented out (I don't know why I left the version where we just pretend that newborns should get shocks as if it were the beginning of the second period; it's stupid).  In combination with my long response to Edmund above, where we give users the ability to specify whatever history of shocks they want (maybe drawing from IncomeDstn, maybe impulse response, maybe something else), this is the most straightforward approach that works.

dclow:
Sorry, I was sloppy in my suggestion.  You would need something like:

IncomeDstnNow    = self.getIncomeDstnSim()

where getIncomeDstnSim is a class method defined as (for example):

def getIncomeDstnSim(self):
    if self.t_sim == 120: #time period some exogenous, unforeseen shock we want happens in
        return ((income distribution of the unforeseen shock))
    else:
        if self.t_age>0:
            return self.IncomeDstn[self.t_cycle - 1] #comments explaining this
        elif self.t_age == 0: #deal with newborns
            return ((degenerate income distribution))

The more I think about it, this doesn't seem all that different than what you're proposing.  I guess the biggest difference is that the task of using the appropriate income distribution is outsourced to a special-purpose method, rather than being done directly in getShocks()

dclow:
A few other less substantive comments:

(1) I'd argue track_vars should be renamed something to tie it much more explicitly to simulation.  track_simulated_vars, or track_vars_in_sim, or something.

(2) simDeath is not declared in HARKcore, but everything else is (simBirth, simMortality, etc).  Why?

(3) I like your reasoning for calling e.g. getShocks rather than simShocks.  So why not getMortality rather than simMortality (or maybe getBirth and getDeath rather than simBirth and SimDeath?)

(4) You removed preSolve from ConsMarkovModel, which invokes the call to self.checkMarkovInputs().  Now there seems to be no call to self.checkMarkovInputs(), but there should be.  Am I right?

mnwhite:
The structure you proposed above (3 days ago) also doesn't work, or would be extremely slow.  self.t_age is an array, not a float; you can't do if self.t_age == n.

Responses to smaller points:

(1) Maybe, but:
(a) This name matches the one for the Market class, and the concept is identical.
(b) I'd prefer a short, memorable name because this is something the user *will* be interacting with on an instance by instance basis.  That is, something like time_vary is class-specific, with all instances sharing the same time_vary; the user never needs to think about time_vary unless he's writing a new solver/agent class.  But track_vars is specific to an instance of an AgentType-- the user needs to say which variables he wants to track *depending on his purpose*.

(2) There is a very good reason for this, which I can expla-- hey, look behind you! A three-headed monkey!

(3) English, mostly.  getMortality didn't seem like a sensible thing to write, because "mortality" means both "the process of death/birth" and "the condition of being mortal".  If I write simMortality, it points to the former; if I write getMortality, it sounds like I'm immortal and am attaining mortality.  I'm not attached to this, and I think it should be changed to getMortality if we implement something like the pre-specified shock history for replacement events.  That would enable easier analysis of "meteor of death" scenarios for first year grad macro students.

(4) Gah, mistake by me.  I was manually combining my private version of HARK with the latest master.  These got omitted by mistake, I will fix.

mnwhite:
I propose that I should:

1) Rectify the smaller issues raised by DCL, including simMortality --> getMortality in anticipation of near future functionality.

2) Actually add the pre-specified shock functionality, which will include the "simulate shocks as normal and write it down" style, as well as a basic version of impulse-response.

3) Run a does-Edmund-hate-it test, which would include him extending the impulse-response method to be fancy and have options.

I'll come back to the pre-spec age/death stuff later.  I think my HARK work after this should focus on updating the manual and models doc.

dclow:
Sounds good to me.  You've convinced me that if I want the more substantive changes I should actually code them up and verify they work well first.  I don't think there's any reason to hold this up until that happens, since if that ever does happen it would be easy to do later.

compumetrika:
Was looking into an auto-docs update, and noticed that this pull request hasn't been executed yet. Is this going to be merged soon? I.e. "any day now," or further in the future?

mnwhite:
I returned to other research and haven't had time to return to this.  Maybe
I will this weekend.

compumetrika:
OK, sounds good. If it is happening reasonably soon I'll update the docs right after the merge.

mnwhite:
Ok, I've done the stuff I said I would do above, except for writing a baby version of impulse response.  I'll ping Edmund and see if he can port his code over easily to this version.  It should be simple: all he has to do is store the shock histories in PermShkNow_hist and TranShkNow_hist and set read_shocks = True.

PR #107, opened 2017-01-30
Rewrote LinearInterp from scratch

mnwhite:
LinearInterp was previously mostly an extension of
scipy.interpolate.UnivariateSpline.  It has now been rewritten as a
fully custom class for HARK.  This was done because UnivariateSpline
uses 6-7x more memory than is necessary.  This is not a problem in
isolation, but can lead to memory constraints being binding in very
large models-- those with tens of thousands of LinearInterps.

All code runs identically as before these changes.  Old version remains
as LinearInterpOLD, for testing.  Should be deleted.  Some applications
run 5% faster, others are up to 14% slower than before.  We can fix
speed later by adding an option to the class that makes it use 50% more
memory.

Also fixed all classes where distance_criteria was not a static
attribute.

### DIFF
--- a/HARKinterpolation.py
+++ b/HARKinterpolation.py
@@ -596,6 +596,133 @@ class ConstantFunction(HARKobject):
     derivativeZ = derivative
     derivativeW = derivative
     derivativeXX= derivative
+    
+    
+class LinearInterp(HARKinterpolator1D):
+    '''
+    A "from scratch" 1D linear interpolation class.  Allows for linear or decay
+    extrapolation (approaching a limiting linear function from below).    
+    '''
+    distance_criteria = ['x_list','y_list']
+    
+    def __init__(self,x_list,y_list,intercept_limit=None,slope_limit=None,lower_extrap=False):
+        '''
+        The interpolation constructor to make a new linear spline interpolation.
+        
+        Parameters
+        ----------
+        x_list : np.array
+            List of x values composing the grid.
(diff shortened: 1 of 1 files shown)


dclow:
Please let me know what you think of my suggested changes, Matt.  Most obviously, there's now just one method _evalOrDer that does anything.  I've also dropped the _isscalar test for i, and now just have the line i      = np.maximum(np.searchsorted(self.x_list[:-1],x),1), because np.maximum returns a float if x is a float.  I really think this change works, but it's possible I broke something and missed it.  The code also no longer initializes y or dydx as numpy arrays of type float.  I believe by changing 1 to 1. in the evaluation of y, the y array will always be a float now anyway.  There is no such guarantee for dydx, I guess, so if we really want to make sure it is an array of floats, some code might need to be added there.

I'd especially be grateful if you could test this with other parts of HARK you understand better than I do, e.g. maybe cstwMPC, to make sure it doesn't break anything there.  This push is not necessarily supposed to be accepted as-is, but just as a suggestion and perhaps a starting point for something else.  [Though if it works well, great!]

mnwhite:
The changes you just made are compatible with the rest of HARK.  Methods like _eval should never be accessed externally, only through HARKinterpolator1D.__call__ (etc).  I support this code reduction, and we should implement it for the other classes.

dclow:
Awesome!  Sorry again this took me so long

PR #108, opened 2017-02-11
Fix to approxLognormal

mnwhite:
approxLognormal previously didn't work properly at very extreme inputs.
For example, trying approxLognormal(N=20,mu=-26,sigma=15) results in
many of the discrete values being 0.  In some cases, all but the top
point were 0.  This commit fixes this issue by adding an alternate
version for extremely low shocks.  The math has been simplified a bit,
and a typo in the docstring removed.


mnwhite:
There might still be an unresolved opposite version of this problem, but I haven't explored it.  I think it might come up if the user specifies tail_N very large, so that the highest discrete points have probabilities on the order of 10^-16.

dclow:
Looks good!

(I'm even ahead *cough* of schedule *cough*)

PR #110, opened 2017-04-19
Tiny fix to decay interpolation in linear interpolation 

dclow:
Changed one line of code in HARKinterpolation.py, in  the calculation of x_temp in the _evalOrDer method of LinearInterp.  Previous code didn't work.

This should be a super simple PR.  Let me know if not.


mnwhite:
Ah yes.  And for fixes like this, I think we can agree to ignore the PR protocol and go straight to a commit.  This is small and critical for functionality.

dclow:
Agreed on ignoring the PR protocol for tiny commits like this from now on.  Thanks!

Issue #111, opened 2017-05-09
Necessity to check the conditions for nondegenerate solutions in ConsIndShockModel


derinaksit:
ConsIndShockModel should be updated in a way that it checks the conditions for nondegenerate solutions and reports whether a solution exists under the given parameter values. For example, if the user asks HARK to solve a buffer stock model for which the Finite Value of Autarky condition fails, the code should report back a warning like "under the given parameter values, it is known that no nondegenerate solution exists" with a link to the relevant theoretical literature. I will prepare a pull request to add this feature.

mnwhite:
Yes, this would be good.  Per our in-person conversation, this should be a method for the perfect foresight type and the ind-shock type (which extends the former). Because it will be inherited by child classes but might not be applicable/correct, it would be wise to put in dummy methods for child classes that raise an error when called.  See ConsIndShockModel.KinkedRconsumerType.makeEulerErrorFunc for an example.

mnwhite:
Addressed in #115

PR #115, opened 2017-05-29
Checking the conditions for nondegenerate solutions 

derinaksit:
Addition of a method that checks 6 conditions, which may be sufficient for a
nondegenerate solution with the relevant reference to the theoretical literature.


mnwhite:
I will take a look at this later this month, after I get back from travel.  This PR concerns Issue #111 posted by Derin a week or two ago, following discussion with CDC and me.

mnwhite:
I will take a look at this and merge into master this weekend.  Apologies for being slow.

mnwhite:
1) It looks like you've put the entirety of the checkConditions method in PerfForesightConsumerType.  This won't work, as a PerfForesightConsumerType does not have the attribute PermShkDstn, UnempPrb (etc), but this information is necessary to compute the GIC, WRIC, and FVAC.  The RIC, AIC, and FHWC can be evaluated by a PerfForesightConsumerType, but the other conditions do not exist.

Please restructure the code to evaluate one condition at a time in a block, and display results as appropriate.  For example:

# Evaluate and report on the finite human wealth condition
FHWC=self.PermGroFac[0]/self.Rfree
if verbose:
    blah blah
elif FHWC >= 1.:
    blah blah

Do this for the RIC, AIC, and FHWC in PerfForesightConsumerType.checkConditions.  For IndShockConsumerType.checkConditions, the first line should be PerfForesightConsumerType.checkConditions(self,verbose), which will execute the stuff you wrote for the RIC, AIC, and FHWC.  Below that, evaluate and report on the GIC, WRIC, and FVAC.  That way the method will run on both PerfForesightConsumerTypes and IndShockConsumerTypes.  Your "not implemented" version for KinkedRConsumerType is correct.

2) Please rename checkCondition --> checkConditions.  Multiple conditions are being checked.

3) For style: the assertion "This method only checks for the conditions under infinite horizon models with a 1 period cycle." should appear at the top of the file, not as an else clause on the entire function.  As is, the entire function has an extra indentation, but the same thing can be achieved by beginning the method with:

if self.cycles != 0 or self.T_cycle > 1:
    print('blah blah')
    return

Then the method will just end, and you don't need an else clause that creates an extra indentation.

mnwhite:
Ugh, I need to learn how to format these comments.  I think GitHub interprets things in markdown format; that giant heading is supposed to lead with a hash to begin a comment:

`#Evaluate and report on the finite human wealth condition`

derinaksit:
I made the necessary corrections that you pointed out and I believe it works now. Let me know if you notice any other problems or if you prefer an alternative feedback mechanism.

mnwhite:
The method still doesn't work because:

1) You refer to T_total rather than T_cycle; see the code snippet in my previous comment.

2) PerfForesightConsumerType.checkConditions still tries to use information about the distribution of permanent shocks, but this does not exist.  The point of splitting up the 6 condition checks into 3 and 3 between the two types is that the income distribution *does not exist* for the perfect foresight type.  You correctly split up the two batches of conditions, but your code still tries to make calculations based on PermShkDstn.  Just delete those lines from PerfForesightConsumerType.checkConditions.

Also two very tiny fixes: verbose should have a default of False, so that checkConditions can be called without an argument.  Initializing line for both types' method:

def checkConditions(self,verbose=False)

And the docstrings should have None as the Return, not feedback.  The Returns are literally the output of the function, so the user knows what X will be like if s/he does X = thing.method(args).

derinaksit:
Sorry, forgot to delete those lines with PermShkDstn under PerfForesightConsumerType previously. I fixed it along with the T_cycle replacement for T_total and the two tiny fixes.

mnwhite:
Great, thank you.

PR #117, opened 2017-06-20
Cons agg shock fix

mnwhite:
This PR fixes a few small accounting errors in the aggregate shocks model, and eliminates the loop over MaggNow values.  You might think that further vectorization / array-ization would speed things up... but no: code is somehow slower now.  Discussed these changes yesterday morning with Edmund Crawley.

Also: One line fix to HARKsimulation that is *necessary* after updating to numpy v1.12.


mnwhite:
Closing this PR because it has been superceded by #125

PR #125, opened 2017-07-28
ConsAggMarkov

mnwhite:
This PR expands and supercedes #117 , from which it branches off.  See that PR for more description of what's in this branch.

I have added a "Markov aggregate shocks" model to ConsAggShockModel.py.  In this model, consumers face idiosyncratic and aggregate shocks like in ConsAggShock, but the macroeconomic state is also characterized by a discrete Markov state.  As written, only the objects IncomeDstn, Afunc, and PermGroFacAgg vary with the discrete state, as these are (at least partially) "macroeconomic" inputs.  This can later be extended to allow technology parameters CapShare and DeprFac to vary with the state, which would cause Rfunc and wFunc to vary across states for the solver.

The new AgentType is supported by CobbDouglasMarkovEconomy, the Markov extension of CobbDouglasEconomy.  This Market subclass knows how to produce a Markov state history that has sufficient coverage of each discrete state to allow a decent update of the aggregate saving rule by state.  I have tested it with up to 21 states, which takes ~45 minutes to solve.

The new model has several examples at the bottom of ConsAggShockModel.py: a simple two-state version, an even simpler KS-style example (minus serially correlated idiosyncratic unemployment), and a "poly state" example that can have an arbitrary number of states.

The basic ConsAggShockModel now handles aggregate permanent growth, which was a serious oversight before.  Construction of the Mgrid takes into account aggregate growth (in calculating the perfect foresight steady state); the Markov version calculates long run average aggregate growth and uses that to construct Mgrid.


mnwhite:
Forgot to mention: I also improved combineIndepDstns to handle multivariate inputs.  This greatly simplifies the method that adds the aggregate shocks to the idiosyncratic shocks in IncomeDstn.  Made some related fixes to pLvl in simulation.

mnwhite:
This PR is now incorporated into ESC's PR for the StickyE project.  Closing.

PR #126, opened 2017-08-04
Fixed a bug in makeMarkovApproxToNormal in HARKutilities.py.  Before,…

dclow:
Came across a bug in makeMarkovApproxToNormal, so (I think) I fixed it.  Before, the weights in alpha could go below 0 or above 1, which obviously caused issues.

I also created another function, makeMarkovApproxToNormalByMonteCarlo, largely as a way to check makeMarkovApproxToNormal.  Now, the two functions seem to give very comparable output, but I will let you verify this, and check this code any other way you want.

makeMarkovApproxToNormalByMonteCarlo was really just intended to check makeMarkovApproxToNormal; I'm not sure it needs to stay in HARK once we are sure makeMarkovApproxToNormal works.  I leave that decision up to you.

### DIFF
--- a/HARKutilities.py
+++ b/HARKutilities.py
@@ -646,19 +646,69 @@ def makeMarkovApproxToNormal(x_grid,mu,sigma,K=351,bound=3.5):
     bot = x_grid[sample_pos-1]
     top = x_grid[sample_pos]
     alpha = (sample-bot)/(top-bot)
+
+    # Keep the weights (alpha) in bounds    
+    alpha_clipped = np.clip(alpha,0.,1.)    
     
     # Loop through each x_grid point and add up the probability that each nearby
     # draw contributes to it (accounting for distance)
     for j in range(1,x_n):
         c = sample_pos == j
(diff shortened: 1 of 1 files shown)


mnwhite:
Yes, you are correct in finding this bug.  I'm not a huge fan of this function, as it was just sort of thrown together when I needed to make a Markov approximation to an AR1.

Edmund Crawley has added the function TauchenAR1 (which should be renamed makeTauchenAR1) to a branch on his fork that we're currently working on.  For most purposes for which I intended to use makeMarkovApproxToNormal, makeTauchenAR1 is better suited.  The only advantage to my function is that you can pick a weird grid of values if you want (and it does one row at a time), but that can easily be added as an option to makeTauchenAR1

mnwhite:
You misplaced the comma or parentheses in the third part of the assertion.  Method crashes.

`(np.isclose(np.sum(p_vec)),1.)` should be `(np.isclose(np.sum(p_vec),1.))`

dclow:
Should be fixed now

Issue #124, opened 2017-07-21
Model(s) overhaul: ExplicitPermInc and PersistentShock


mnwhite:
We should overhaul / redo the ExplicitPermInc and PersistentShock solvers/models to represent the consumption function in terms of normalized variables.  That is, cFunc will take in mNrm values and spit out cNrm values, rather than the current system of taking in mLvl and spitting out cLvl.  This will simplify the simulation methods, as fewer need to be overwritten.

Moreover, this will likely result in more accurate solutions that are more easily understood graphically.  At high levels of permanent income, far above the center of the distribution, adjacent 1D "layers" of the consumption function diverge more and more from each other, as the differences between consecutive pLvlGrid values become large.  The shape of the "above" and "below" 1D cFuncs for a pLvl between grid values might be significantly different, reducing accuracy of the interpolation.

Making this switch will make it easy to graphically verify that ExplicitPermInc is solving correctly: plotting the cFunc at various pLvl values *should* perfectly align with each other.  They might not perfectly align because the pLvlGrid is bounded, so the top and bottom 1D cFuncs are subject to extrapolation error.  In the PersistentShock model, we can more easily see that high pLvls are associated with lower normalized consumption, as consumers expect their productivity to revert to the mean (and vice versa for low pLvls).

Most broadly, the inheritance structure of these models should not be ConsIndShock --> ExplicitPermInc --> PersistentShock, but rather ConsIndShock --> ConsGenIncProcess, with ExplicitPermInc and PersistentShock being child models of the new ConsGenIncProcess.  All of these will use the same solver, with the only difference being how the attribute (and solver input) PermIncNextFunc is constructed.  This function is a generalized income process that will be used as:

pLvlNext = PermIncNextFunc(pLvlNow)*PermShkNext

The ExplicitPermIncConsumerType class will construct its PermIncNextFunc(s) as linear functions (using primitive attribute PermGroFac), while PersistentShock will construct AR1-type functions based on PermGroFac, PermIncCorr, and pLvlInitMean.  The baseline GenIncProcessConsumerType will have no constructor for PermIncNextFunc.

The ConsGenIncProcess solver will let us better model how the drop in permanent income at retirement depends on pLvl, i.e. the SS replacement rate (approximating the AIME calculation).  As is, we use Cagetti's method of imposing different permanent income drops based on *education*.  The new way will better reflect reality.

mnwhite:
[...]

PR #128, opened 2017-08-21
Persistent shock model is now general income process model

mnwhite:
This PR addresses issue #124 concerning ConsPersistentShockModel.py.  It replaces that module with ConsGenIncProcessModel.py, and extends the changes to ConsMedModel.py (the original module is preserved in ConsMedModelOLD).  Both the "explicit permanent income" and "persistent shock" models now use the exact same solver, with an input called pLvlNextFunc that gives next period's expected permanent income as a function of this period's.  In ExplicitPermInc, that's just a linear function (constant permanent income growth factor); in PersistentShock, it's an AR1-style function (which now has its own class).  This change adds the capability to have oddball pLvlNextFuncs, with the most obvious use being an approximation to the SS benefit structure.

These models still can't handle cycles>1 for the same reason as before: it's not clear how to construct the pLvlGrid for a particular period when it should have different values on each cycle.  It would also create a problem in the persistent shock model, as the AR1 would want to regress back to a different mean level for each cycle.  To get around this, users will have to explicitly list out the repeating cycle of periods in the inputs they pass to the constructor (which is trivial in Python) and use cycles=1.  Long run permanent income growth also won't work in the persistent shock model with cycles=0.

I wrote in #124 that we should redo the representations of policy and value functions to use normalized levels, and convert when called.  I realized this won't work because of pLvl=0, which would make undefined cNrms (etc).

Before approving and merging to master, we should delete ConsMedModelOLD.py and ConsPersistentShockModel.py, and remove a few obsoleted lines from ConsumerParameters.py.  I'm leaving these in for now for testing and comparison purposes.

### DIFF
--- a/ConsumptionSaving/ConsGenIncProcessModel.py
+++ b/ConsumptionSaving/ConsGenIncProcessModel.py
@@ -262,7 +262,7 @@ class ConsGenIncProcessSolver(ConsIndShockSetup):
     period permanent income (subject to shocks).
     '''
     def __init__(self,solution_next,IncomeDstn,LivPrb,DiscFac,CRRA,Rfree,
-                      PermIncNextFunc,BoroCnstArt,aXtraGrid,pLvlGrid,vFuncBool,CubicBool):
+                      pLvlNextFunc,BoroCnstArt,aXtraGrid,pLvlGrid,vFuncBool,CubicBool):
         '''
         Constructor for a new solver for a one period problem with idiosyncratic
         shocks to permanent and transitory income, with permanent income tracked
@@ -286,7 +286,7 @@ class ConsGenIncProcessSolver(ConsIndShockSetup):
             Coefficient of relative risk aversion.
         Rfree : float
             Risk free interest factor on end-of-period assets.
-        PermIncNextFunc : float
+        pLvlNextFunc : float
             Expected permanent income next period as a function of current pLvl.
         BoroCnstArt: float or None
             Borrowing constraint for the minimum allowable assets to end the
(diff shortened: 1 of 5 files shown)


mnwhite:
After discussing with CDC and making some additions, this PR should be ready to merge.

llorracc:
Right, $P_{0}$ is nonzero.  So $P_{t}$ will be nonzero so long as $P_{t-1}$ is nonzero.  

I think @mnwhite was just trying to use the fact that we rule out $P_{0}=0$ as a way to illustrate clearly the structure of what we are doing.  Maybe it is obvious, but it helps make the obvious crystal clear.

mnwhite:
Sort of. Let me try to unpack my original comment more.

In ConsGenIncProcessModel, the consumption function is a
LinearInterpOnInterp1D (actually a VariableLowerBoundFunc2D on that, but
let's stick to the important part). It's composed of a list of
pLvL-specific 1D interpolators over mLvl. At some point, I made a
comment-to-self to rewrite that as 1D interpolators on mNrm, and then have
an extra "conversion" step when it's evaluated at (mLvl,pLvl).

In the top post of this thread, I noted that this won't work, because the
pLvlGrid *might* include pLvl=0, either because there actually *is* some
way in some specification to end up with pLvl=0, or to have a proper
bounding cFunc for when pLvl *approaches but does not reach* 0, through a
terrible series of shocks. You can't "convert to normalized market
resources" when pLvl=0, so it can't work.

Random thought I had the other day: at some point I had a demo for
TractableBufferStock that shows it being solved "the slow way" using
ConsMarkovModel vs the "the fast" way. We can actually use ConsIndShock and
have the "zero" permanent income shock be arbitrarily close to 0, but still
positive.

nomadj1s:
@mnwhite Thanks for the explanation. One additional follow up: the model nests the unit root case, which is solved in `ConsIndShock`. In that special case, even though we can normalize, implicitly, is pLvl=0 or close to zero possible? And if so, why isn't it a problem in the unit root case?

mnwhite:
In ConsIndShock, pLvl=0 would be a problem, as it would make mNrm infinite
and thus break the code. For ExplicitPermInc (the unit root example for
ConsGenIncProcess that solves the ConsIndShockModel without normalization),
pLvl=0 is allowed to be in the grid, and it's allowed to actually happen
during both solution and simulation. No normalization is ever performed (as
far as I know), so there will never be a division by zero.

It might be that there's a very small adjustment to the pLvlNextFunc that
needs to happen to account for pLvl=0 (e.g. if it's taking log(pLvl) as
part of its math), but that's like 2 lines.

PR #133, opened 2017-11-02
Port ConsIndShock into OpenCL

mnwhite:
As a demonstration of how to call OpenCL code from HARK, I have ported the ConsIndShock model (both solver and simulator) into OpenCL.  The new AgentType class is constructed by passing a list of ConsIndShockConsumerType instances, whose data are reformatted into OpenCL buffers.  The solve and simulate methods have been remapped to call the appropriate OpenCL kernels rather than use the core AgentType functionality.

The code for this branch will not work "out of the box", but needs an additional package to be installed.  I have added a file to the documentation folder on how to set up opencl4py and check which devices are available on a machine.


edmundcrawley:
A few notes from ESCs attempt to follow the documentation and run this:

1. The first link in the documentation needs to be updated to: https://developer.amd.com/amd-accelerated-parallel-processing-app-sdk/
2. opencl4py has a download button that doesn't seem very helpful (It is a .gz file, but I couldn't unzip it with WinZip, maybe I was being dim). In the end I went to the GitHub page and downloaded from there.
3. In the documentation
 print(platforms.dump devices) 
should read 
print(platforms.dump devices())
4. It wasn't clear that I needed to run os.environ["PYOPENCL CTX"] = "0:0,1,2" before going through the devices with ctx.devices[n].extensions

mnwhite:
Addressed those issues, and expanded comments in the example files.  I forgot to mention it in the commit message, but I also made use of the Queue.finish() method to wait for the command queue to clear.  This is the "there's probably a function for this" that I referred to above.

mnwhite:
I'm closing this PR (but please don't delete the branch!) because it's from way before HARK was package-ified; files need to rearranged and tested in new structure.  This PR was ready for merge and had been reviewed by CDC, but just wasn't merged.

PR #151, opened 2018-04-25
explore reticulate, reduce rpy2 dependence

ganong123:


ganong123:
Created by accident. Deleting.

PR #152, opened 2018-04-27
Add if __name__ == '__main__' statement

edmundcrawley:
Matt,

I've made a small change to cstwMPC.py to add a statement:

if __name__ == '__main__':

at the start of the part of the file where it actually begins doing stuff (as supposed to defining stuff). Obviously I've also indented all the code below that line.

Do you know of a reason not to do this? I sometimes call cstwMPC in other code, so need this there to load the relevant objects without wasting time running all the code.

Edmund


mnwhite:
This is fine.  Replying more fully in private correspondence.

Issue #85, opened 2016-09-29
Add setup.py so module can be installed.


theoldfather:

mnwhite:
This has been discussed and an early version was developed.  I believe we decided to hold off on this for a while, but now I forget our reasoning.  Leaving open for others to fill in details.

brainwane:
I'm working on this in https://github.com/brainwane/HARK/tree/installability which I should have linked here much earlier - apologies! I should have it merged into `master` later today.

brainwane:
We have some early developer releases available for `pip install` at http://pypi.org/project/econ-ark and `conda install` at https://anaconda.org/sumanah-changeset/econ-ark .

mnwhite:
Before merging the installability branch into master, can we fix the Models vs Demos issue?

brainwane:
Thank you for that very good point, @mnwhite!

The Models versus Demos issue @mnwhite mentioned is: in the `installability` branch, we've changed the structure of the code, with several commits by @jasonaowen -- some of which he'd like to do over, reverting how he separated out Demos.

@compumetrika asked when we were planning on merging the `installability` branch back into `master`, and I originally suggested doing it this week. But @mnwhite makes a good point: it would be nice to have only one giant breaking change in `master` instead of two in a row. We do have some time pressure here, since we want to get this all merged and tested, and cut a new release, before Tuesday, 19 June (when Chris and Matt do their next public demo).

So here's how I propose to do this:

* Today, I'll make a separate pull request with other cleanup that's immediately applicable to the `master` branch.
* Tomorrow, I'll further prepare `installability` to make it more ready to merge into `master`.
* Next week, starting Monday, Jason will address the Models vs Demos issue on `installability`.
* After he's done that, ideally in the middle of next week, we make a pull request to merge `installability` into `master`, and get that reviewed and merged.

mnwhite:
(Non-code note: I'm back stateside.  Chris is doing the Bank of England and Venice presentations solo!)

PR #160, opened 2018-06-14
Remove (some) unused imports

jasonaowen:
In support of finding dependencies and transitive dependencies that make packaging HARK more difficult, remove some of the import statements that bring in dependencies that are not used in the code, as detected by flake8:

```bash
git ls-files '*.py' | xargs flake8 --select=F401
```

In particular, some of the apparently-unused imports not removed are due to usages flake8 can't detect, such as the fact that the HARKutilities.warnings import has side effects, or the `exec()` use of `copy.copy`.

Issue #85 Add setup.py so module can be installed


mnwhite:
This is fine.  It's not relevant to this PR, but I'd like to remove Try-Alternate-Parameter-Values.py entirely.  I don't know why it was added.

Also, those uses of exec were written literally the first week I started programming HARK. And I had extremely little Python experience before that.  They can all be replaced with better style.

jasonaowen:
@mnwhite totally! I just wanted to make clear why I restricted my scope, and why, if you run that shell script now, you'd still see warnings. I didn't mean to call you out! :smiley:

PR #161, opened 2018-06-14
Make HARK pip- and conda-installable

### LINKED ISSUE #85: Add setup.py so module can be installed.

jasonaowen:
Make the HARK installable with pip and conda.

Part of this work is restructuring how the source code is laid out in the repository. All the Python files in the root were moved into `HARK/`, so that the generated package has a single, top-level module named `HARK`, and everything is named relative to that: `HARK.core`, `HARK.ConsumptionSaving.ConsMarkovModel`, and so on.

This branch has been rewritten to address the problem on the previous WIP branch of having many of the models named "Demos", such as `Demos.ConsumptionSaving.ConsMarkovModel`; they should all be named correctly now. Some demo code is still included in this package; we plan on removing it later.

As of this writing, we have not yet made a new package and uploaded it to PyPI; we expect to do so after this PR is merged (and possibly once the Python 3 changes are merged; we need to discuss that).

Resolves #85 Add setup.py so module can be installed

Co-authored-by: @brainwane


mnwhite:
Ok, this is going to take me a while to look through.  On the face of it, this is what we talked about.  The only small thing is that I kind of want all of the material in HARK.core to just be top level HARK. That stuff is the absolute kernel of HARK.

This could definitely be done later, but it seems like something we don't want to change on people once its in the master.

Most of HARKparallel can now also be moved into that (the parallel Nelder-Mead algorithm should go in HARK.estimation).  But that's way, way lower priority.

jasonaowen:
> Ok, this is going to take me a while to look through.

@mnwhite it's a big change! I tried to break it down into small commits; I'd suggest reviewing the commits one-by-one rather than the all-changes view.

>  The only small thing is that I kind of want all of the material in HARK.core to just be top level HARK. That stuff is the absolute kernel of HARK.
> 
> This could definitely be done later, but it seems like something we don't want to change on people once its in the master.

That should be a reasonable change; I can work on that tomorrow. I think that means adding `from core import *` to `HARK/__init__.py`, and then changing all the imports elsewhere to refer to `HARK.thing` instead of `HARK.core.thing` (or `from HARK import thing` instead of `from HARK.core import thing`). Does that sound reasonable, @npalmer-professional and @mnwhite?

mnwhite:
That's exactly it, yes.

I just looked at Zulip for the first time in a while.  I had no idea we were trying to get this done this week.  When I said "a while", I meant "I'll try to get to it in a few weeks".  I'm going to have to waive my right to provide further input on this.

npalmer-professional:
@brainwane @jasonaowen @mnwhite Yeah this will be a doozy of a commit. I'll start reviewing it tomorrow, avoiding the HARKcore thing that Jason will be working on.

jasonaowen:
I added a commit to expose everything in `HARK.core` as just `HARK`, as we discussed last night: 4be101c.

jasonaowen:
Here's the testing plan we're using for this pull request, courtesy of @mnwhite and @npalmer-professional:

> Here is the order of priorities, from DO IT FIRST to DO IT LAST
> 1) all of the HARKblah.py modules except HARKparallel.py
> 2) ConsIndShockModel.py and ConsumerParameters.py
> 3) ConsAggShockModel.py
> 4) all other /ConsumptionSaving model files
> 5) HARKparallel.py
> 6) files in /ConsumptionSaving/Demos/
> 7) FashionVictim

We want to compare the output from running these files on `master` with the output of running them on this branch. As many of them include randomization at some level, it is not important that the numbers be identical, but it is important that the graphs look visually similar.

jasonaowen:
`ConsAggShockModel.py` is a long-running simulation; for the purpose of testing these changes, we can disable some steps to reduce the amount of work (and amount of time) taken:

```diff
diff --git a/HARK/ConsumptionSaving/ConsAggShockModel.py b/HARK/ConsumptionSaving/ConsAggShockModel.py
index aee59f9..b7f5c95 100644
--- a/HARK/ConsumptionSaving/ConsAggShockModel.py
+++ b/HARK/ConsumptionSaving/ConsAggShockModel.py
@@ -1759,9 +1759,9 @@ if __name__ == '__main__':
     solve_agg_shocks_market = True # Solve for the equilibrium aggregate saving rule in a CobbDouglasEconomy
 
     solve_markov_micro = False # Solve an AggShockMarkovConsumerType's microeconomic problem
-    solve_markov_market = True # Solve for the equilibrium aggregate saving rule in a CobbDouglasMarkovEconomy
+    solve_markov_market = False # Solve for the equilibrium aggregate saving rule in a CobbDouglasMarkovEconomy
     solve_krusell_smith = True # Solve a simple Krusell-Smith-style two state, two shock model
-    solve_poly_state = True # Solve a CobbDouglasEconomy with many states, potentially utilizing the "state jumper"
+    solve_poly_state = False # Solve a CobbDouglasEconomy with many states, potentially utilizing the "state jumper"
 
     ################ EXAMPLE IMPLEMENTATIONS OF AggShockConsumerType ##########
 
```

(via @npalmer-professional in [chat](https://chat.opentechstrategies.com/#narrow/stream/16-Econ-ARK/subject/Python3/near/43533))

jasonaowen:
I note that, in both `master` and in this branch, `ConsMarkovModel.py` has an error partway through:

```
Solving a Markov consumer took 0.3596 seconds.
Consumption functions for each discrete state:
divide by zero encountered in double_scalars
Solving an "unemployment immunity" consumer took 0.4267 seconds.
Consumption functions for each discrete state:
All-NaN slice encountered
Solving a serially correlated growth consumer took 0.3753 seconds.
Consumption functions for each discrete state:
Traceback (most recent call last):
  File "ConsMarkovModel.py", line 1468, in <module>
    MarkovEconomy.makeMkvShkHist() # Simulate a history of markov shocks
  File "ConsMarkovModel.py", line 1235, in makeMkvShkHist
    Events      = np.arange(self.PermShkAggDstn[0].size) # just a list of integers
AttributeError: 'MarkovSmallOpenEconomy' object has no attribute 'PermShkAggDstn'
```

Because this behavior is the same between these branches, I am inclined to say fixing it is out of scope; if you like, I would be happy to open an issue, and if it gets fixed before this gets merged, I would rebase this branch on top of the fix.

jasonaowen:
I've done comparison testing of the following:

- [x] all of the HARKblah.py modules except HARKparallel.py
- [x] `ConsIndShockModel.py` and `ConsumerParameters.py`
- [x] `ConsAggShockModel.py`

As well as some of the ConsumptionSaving models:
- `ConsGenIncProcessModel.py`
- `ConsMarkovModel.py`
- `ConsMedModel.py`

So far, everything looks good!

edmundcrawley:
Yes, I agree with @mnwhite  that we can remove the problem classes and example from ConsMarkovModel as they have been superseded by classes in ConsAggShockModel. I have made this change in a branch and will issue a pull request shortly.

mgerring:
Finished testing, everything looks good. Complete notes are below.

First, the version currently on pypi doesn't work. Instead of running `pip install econ-ark` you need to run `python setup.py install` from the root folder of the installability branch.

Second, I used the suggestions for speeding up `ConsAggShockModel`.

Full notes:

```
HARKcore/core:
Same except HARKcore becomes HARK.core

HARKestimation/estimation:
"

HARKinterpolation/interpolation
"

HARKsimulation/simulation
"

HARKutilities/utilities
"


ConsumptionSaving/
	ConsIndShockModel:
		Identical

	ConsumerParameters:
		Both run with no output

	ConsAggShockModel:
		Identical

	ConsGenIncProcessModel:
		Identical

	ConsMarkovModel:
		Both report the same error:
```
Consumption functions for each discrete state:
Traceback (most recent call last):
  File "ConsMarkovModel.py", line 1468, in <module>
    MarkovEconomy.makeMkvShkHist() # Simulate a history of markov shocks
  File "ConsMarkovModel.py", line 1235, in makeMkvShkHist
    Events      = np.arange(self.PermShkAggDstn[0].size) # just a list of integers
AttributeError: 'MarkovSmallOpenEconomy' object has no attribute 'PermShkAggDstn'
```

	ConsMedModel:
		Identical

	ConsPrefShockModel:
		Identical

	ConsRepAgentModel:
		Identical

	RepAgentModel:
		Identical

	TractableBufferStockModel:
		Identical

	
	/ConsumptionSaving/Demos
		ChineseGrowth
			Identical

		Fagereng_demo.py
			Identical

		MPC_credit_vs_MPC_income.py
			Identical

		NonDurables_During_Great_Recession
			Identical

	FashionVictim/FashionVictimModel
		Identical
```

HARK/cstwMPC/cstwMPCold.py
```diff
@@ -637,7 +630,7 @@ def sensitivityAnalysis(parameter,values,is_time_vary):
         # If you want this to work, you must edit TractableBufferStockModel slightly.
         # See comments around line 34 in that module for instructions.
         if Params.do_tractable:
-            from TractableBufferStockModel import TractableConsumerType
+            from HARK.ConsumptionSaving.TractableBufferStockModel import TractableConsumerType
```
mgerring: I don't think there's anything wrong with this, but it isn't idiomatic Python. The comment doesn't really explain why it's a conditional import.

mnwhite:
Yeah, that's an old version of code for a research application. We can
probably just delete the whole thing at this point. I think there's one
small section that I haven't moved to the current version.

That little conditional import wasn't even part of the paper. It was just a
few lines I threw in when we wanted to show that you could swap one
(sophisticated) model for another (ludicrously simple) model of consumers
and it would still work.

Man, there is a lot of old stuff in here.

     -- mnw

PR #170, opened 2018-07-06
Add MPC to simulation in ConsGenIncProcessModel

mnwhite:
The MPC is calculated and stored as an attribute (MPCnow) in some models, but this was omitted in ConsGenIncProcessModel.  As it turns out, this functionality is necessary for an exercise/notebook that is being prepared for NBER SI.


mnwhite:
CDC: This was meant for you to merge in back in July.  It's a three line fix for the notebook you requested for NBER.

PR #172, opened 2018-07-14
Update build process to handle Python 2 and 3 compatibility

mgerring:


mnwhite:
Is this ready to merge?  It looks like a quick fix after the Py3 update, but I've been out of the loop in Austin.

mnwhite:
Should have checked Zulip first.  Merging now.

PR #174, opened 2018-07-19
Added a "replicate" file to major modules

npalmer-professional:
Added a 'replicate' file to major modules, so user can easily obtain a copy of the code on their local hard drive to directly play around with.

Usage:

    from HARK.[example-module-name].replicate import replicate
    replicate()

The replication function will ask the user for a location to place a copy of that module, with a default of a sub-directory with the same name as the module, in the home directory. The function will not overwrite any existing file. 

This is a first pass to get this replication functionality into HARK.

### DIFF
--- a/HARK/ConsumptionSaving/replicate.py
+++ b/HARK/ConsumptionSaving/replicate.py
@@ -1,0 +1,104 @@
+'''
+This module contains simple code to copy the current sub-module to a location on 
+your hard drive, as specified by you. The purpose of this code is to provide the 
+user with a simple way to access a *copy* of code that usually sits deep in the 
+Econ-ARK package structure, for purposes of tinkering and experimenting directly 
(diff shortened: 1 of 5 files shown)


npalmer-professional:
**Important note:** in running tests I realized that "current HARK," as of today, is not working for at least SolvingMicroDSOPs.StructuralEstimation.main() 

This appears to be due to changes between [this commit](https://github.com/econ-ark/HARK/commit/4b340003d36c48d33c4e2e6adc7b25437a8cb7ee)  (PR#172) and the [next commit](https://github.com/econ-ark/HARK/commit/902008460fbb1d4c2986b10234645727f2629199). 

I *believe* this PR is based on the most recent HARK code (I believe I merged upstream right before issuing the PR), thus will have the error included -- thus this commit should *not* be merged. I am closing it and will re-open after we've figured out what is going on. 

I'll also raise an issue in main HARK.

Issue #155, opened 2018-06-05
ConsIndShock: cFunc extends below mNrmMin


mnwhite:
A student at Goethe University flagged this issue for us.

The consumption function in ConsIndShockModel exists below mNrmMin, but it should not.

Explanation: cFunc is a LowerEnvelope instance containing cFuncCnst and cFuncUnc.  cFuncCnst hits zero and ceases to exist at mNrmMin, but cFuncUnc continues further down to BoroCnstNat.  LowerEnvelope uses nanmax rather than max, so it throws out the nan from cFuncCnst and thinks it should use cFuncUnc below mNrmMin.

Solution: Add optional boolean input for LowerEnvelope (and related classes) to indicate whether to use max/min or nanmax/nanmin.  Default should be nanmax.

TimMunday:
I will have a PR coming on this in the next couple of days.

mnwhite:
Awesome, let me know if you need help / get stuck.  Hopefully the description above is enough.

Just to clarify, the optional input here would be for the `__init__` method, not `_eval` (or whatever it's called).  That is, the boolean is an object attribute, not an optional input when you want to query the interpolation.

PR #193, opened 2018-09-04
NanBool introduced

TimMunday:
Introduces a parameter NanBool, a boolean which if True, when calculating the  lower envelope during the construction of the consumption function, NA's are omitted. If cFunc is extending below mNrmMin, set NanBool to False to rectify. Default is True.

Solves #155 .

NanBool is a consumer parameter and is introduced to ConsIndShockModel. Therefore it also appears in the models which inherit ConsIndShockModel's characteristics.

It is also introduced into the ConsumerParameters module: it is imported as True if one uses the default parameter libraries.

(DemArk will need to be updated with a few lines setting NanBool to True whenever the the notebooks spell out the whole library instead of importing the defaults. PR to follow shortly).

### DIFF
--- a/HARK/ConsumptionSaving/ConsAggShockModel.py
+++ b/HARK/ConsumptionSaving/ConsAggShockModel.py
@@ -83,7 +83,7 @@ class AggShockConsumerType(IndShockConsumerType):
         # Add consumer-type specific objects, copying to create independent versions
         self.time_vary = deepcopy(IndShockConsumerType.time_vary_)
         self.time_inv = deepcopy(IndShockConsumerType.time_inv_)
-        self.delFromTimeInv('Rfree','vFuncBool','CubicBool','NanBool')
+        self.delFromTimeInv('Rfree','vFuncBool','CubicBool')
         self.poststate_vars = IndShockConsumerType.poststate_vars_
         self.solveOnePeriod = solveConsAggShock
         self.update()
--- a/HARK/ConsumptionSaving/ConsAggShockModel.py
+++ b/HARK/ConsumptionSaving/ConsAggShockModel.py
@@ -83,7 +83,7 @@ class AggShockConsumerType(IndShockConsumerType):
         # Add consumer-type specific objects, copying to create independent versions
         self.time_vary = deepcopy(IndShockConsumerType.time_vary_)
         self.time_inv = deepcopy(IndShockConsumerType.time_inv_)
-        self.delFromTimeInv('Rfree','vFuncBool','CubicBool')
+        self.delFromTimeInv('Rfree','vFuncBool','CubicBool','NanBool')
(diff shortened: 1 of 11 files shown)


TimMunday:
An very short example showing that it works is here[https://github.com/TimMunday/WorkARK/blob/master/ConsIndShock%20lower%20bound.ipynb](url)

mnwhite:
Good job, but you actually did *too much* in this case.

1) Main thing: NanBool should not be an input for AgentType subclasses, only for LowerEnvelope and UpperEnvelope (and their 2D and 3D analogues).  There is *no* situation in which a user would want to do

`MyType = IndShockConsumerType(params,params,params,NanBool=True)`

That says, "make an `IndShockConsumerType`, but make its `cFunc`s wrong below `mNrmMin`."  You're fixing a bug; there's no reason to give the user the option of reintroducing the bug.

2) In ConsIndShockModel.py, the only line you needed to change was 906, to be:

`cFuncNow = LowerEnvelope(cFuncNowUnc, self.cFuncNowCnst, NanBool = False)`

That's it.  None of the other model files need to change at all.

3) To fit with the style of other inputs in HARK.interpolation, please rename to `nan_bool` or `ignore_nan` or something like that.

ADDITIONAL COMMENTS:

4) Your code works, but it can be done with many fewer lines.  Rather than have an if-else at eval time, just put this in `__init__`:

`if ignore_nan:`
~~`self.compare = np.nanmin`
~~`self.argcompare = np.nanargmin`
`else:`
~~`self.compare = np.min`
~~`self.argcompare = np.argmin`

Then just have the eval method use `self.compare` rather than `np.nanmin`.  Each instance will take ~8 bytes more memory (I forget how big a function pointer is), but that's ok.

5) The `_evalAndDer` methods need to be fixed as well.  If you use the snippet I just posted, you can use `self.argcompare`.  Looking at the code, I now see that `_evalAndDer` for `UpperEnvelope` is bugged anyway.  As written, `LowerEnvelope._evalAndDer` ignores NaNs, but `UpperEnvelope` uses them... and turns them into infinities (which will be returned) as the output `y`.  Whoops.

6) This is really small, but I also just noticed that the `_der` methods should call the private/internal method `_evalAndDer`, not the public `eval_with_derivative`.  My bad.

TimMunday:
Apologies! I got a bit carried away it seems...
I'll rework and come back to you.

mnwhite:
No worries: you got a grand tour of HARK!

mnwhite:
Great. I am away this weekend, but will look at this Sunday night (and
hopefully a backlog of other PRs).

mnwhite:
I will look at this later today, around 5pm eastern.

mnwhite:
I forgot to merge this way back.  So much old work on HARK...

Issue #156, opened 2018-06-05
BilinearInterp: Indexing error in BilinearInterp._derY


mnwhite:
Derivative with respect to Y (second input) method has a math error in BilinearInterp, and possibly other related classes.  The last two lines where dfdy is actually calculated have incorrect indices.  Should be:

`dfdy = (
              ((1-alpha)*self.f_values[x_pos-1,y_pos]
            +  alpha*self.f_values[x_pos,y_pos]) -
              ((1-alpha)*self.f_values[x_pos-1,y_pos-1]
            +  alpha*self.f_values[x_pos,y_pos-1]))/(self.y_list[y_pos] - self.y_list[y_pos-1])`

Other interpolation classes' derivative methods should also be checked.

PR #194, opened 2018-09-09
BilinearInterpBug

TimMunday:
Solves #156 

Also checked the trilinear and quadlinear interpolation methods. They are correct.


mnwhite:
Perfect, that's it.  I'm a little embarrassed I didn't take the 2 minutes to do this myself-- I found the error in my private fork while working on a project, made the issue to remind myself to fix it here... then just didn't.

Thanks!

mnwhite:
Any call to BilinearInterp.derivativeY would be wrong. There were two
errors in the element indices, now fixed.

llorracc:
For release notes: Fixed indexing bug in bilinear interpolation

PR #195, opened 2018-09-21
Remove relative import

jvasile:
HARK uses the "from . import foo as bar" construction, but I don't see how the relative import adds any useful semantics as opposed to just doing "import foo as bar".  The relative import threw an error:

> Traceback (most recent call last):
>   File "ConsIndShockModel.py", line 2559, in <module>
>     main()
>   File "ConsIndShockModel.py", line 2413, in main
>     from . import ConsumerParameters as Params
> ValueError: Attempted relative import in non-package

This patch just does a normal absolute import using the module system so we avoid the relative import and avoid throwing that error.  I made this change for every relative import in main() in ConsumptionSaving.

Most of the files in that ConsumptionSaving directory use the relative import at the top level, though I don't understand why.  if we need it for some scenario, I don't see what it is.

### DIFF
--- a/HARK/ConsumptionSaving/ConsAggShockModel.py
+++ b/HARK/ConsumptionSaving/ConsAggShockModel.py
@@ -1753,7 +1753,7 @@ class AggShocksDynamicRule(HARKobject):
 ###############################################################################
 
 def main():
-    from . import ConsumerParameters as Params
+    import ConsumerParameters as Params
     from time import clock
     from HARK.utilities import plotFuncs
     mystr = lambda number : "{:.4f}".format(number)
--- a/HARK/ConsumptionSaving/ConsGenIncProcessModel.py
+++ b/HARK/ConsumptionSaving/ConsGenIncProcessModel.py
@@ -1276,7 +1276,7 @@ class PersistentShockConsumerType(GenIncProcessConsumerType):
 ###############################################################################
 
 def main():
-    from . import ConsumerParameters as Params
+    import ConsumerParameters as Params
     from HARK.utilities import plotFuncs
     from time import clock
     import matplotlib.pyplot as plt
--- a/HARK/ConsumptionSaving/ConsIndShockModel.py
+++ b/HARK/ConsumptionSaving/ConsIndShockModel.py
@@ -2410,7 +2410,7 @@ def constructAssetsGrid(parameters):
 ####################################################################################################
 
 def main():
-    from . import ConsumerParameters as Params
+    import ConsumerParameters as Params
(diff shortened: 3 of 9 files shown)


mnwhite:
Somehow just saw this PR.  I'm not sure why that's there, and I don't think I wrote those imports that way.  Did they get changed during package-ification?  @npalmer-professional ?

jameysharp:
I've tested 5b6c097bca16d0ebfafbdc357b3ea266e3bf72ef and I still get `ValueError: Attempted relative import in non-package` when running any of the changed source files via either of these approaches:

```
python HARK/ConsumptionSaving/ConsAggShockModel.py

cd HARK/ConsumptionSaving
python ConsAggShockModel.py
```

This patch _does_ make running these as a module work:

```
python -m HARK.ConsumptionSaving.ConsAggShockModel
```

but only if I run that command from inside the `HARK/ConsumptionSaving` directory. From anywhere else, I get errors like `ImportError: No module named ConsumerParameters`.

One possibility is to go all-in on this approach:

1. In every module that has `if __name__ == '__main__'`, rewrite _every_ import statement to use absolute imports, including those at the top of the file that aren't inside `def main()`. This fixes running the scripts as e.g. `python ConsAggShockModel.py`.

2. Make sure imported modules are referenced from the root, like `import HARK.ConsumptionSaving.ConsumerParameters as Params`. This fixes running the scripts from working directories other than `HARK/ConsumptionSaving`, assuming that either the `econ-ark` package is installed (possibly with `pip install -e .`) or the user has manually adjusted `$PYTHONPATH` appropriately.

However, it seems to me that the relative imports are nice for code readability and maintainability, so that wouldn't be my first choice.

The option I'd suggest instead resembles how Django's [`django-admin`](https://docs.djangoproject.com/en/2.1/ref/django-admin/) command works.

Specifically, when you `pip install` the project, you'd get a program installed on your path, let's call it `econ-ark-demo` or something. That would take a subcommand indicating which demo you want to run, and it would look for a specially-named function or class in each package to find the requested demo. (Django effectively uses `<app>/management/commands/*.py` to find the subcommands, for each configured app in a project.)

Some useful things this hypothetical program could do:

- programatically list all the available demos
- report help/documentation on each demo
- allow each model to have multiple demos for the user to choose between, instead of having to edit boolean flags in the source code to turn individual demos on and off
- offer a GUI for doing all of the above
- potentially, allow running demos from other projects that build on the econ-ark framework (although I haven't worked out the details)

mnwhite:
This PR was addressed by CDC and I a week or two ago.  Closing.

PR #197, opened 2018-09-26
Removes multiprocessing from FluctuationsTwoAsset.py

edmundcrawley:
Due to a bug that we have been unable to resolve that occurs only in Python 3 to do with multiprocessing, this pull request removes the multiprocessing from FluctuationsTwoAsset.py. The code is slowed down, but still runs in around 10-25 minutes.

I will open an issue to reinstate multiprocessing to this code, probably via joblib and dill.

On a separate note, it doesn't seem right to me that this code lives on a separate fork that is likely to get out of date while the rest of HARK moves forward. I think there has been some talk about separating core functionality of HARK from optional add-ins. I think the BayerLuetticke folder would fit into that second category, especially as it is not well integrated into the rest of HARK.


mnwhite:
I strongly agree with Edmund on his "separate note" point, to the extent that the stuff on this fork is about a particular research project/question.  If the work being done here primarily concerns solving (and simulating) a model, with the purpose/reason/meaning behind *why* you would want to do that left open, then it should eventually be merged into the master branch.

Also, Edmund: Feel free to merge these PRs into Econ-ARK/HARK when they're *not* going to the master branch.  Changes to master should get a second set of eyes, changes to other branches can be self-PR'ed.  You do have commit power to the main fork, yes?

PR #201, opened 2018-10-23
[WIP/RFC] A simple logit-style discrete choice model

pkofod:
Names, directories, style violations and more are very much up for critique. (it's referenced as DARK or dark in some places, but that should be changed to match the agent type's name once that is settled).

I created this tiny example of solving Rust's (1987) model of capital replacement. I did so because it's the baby version / precursor on the discrete side of things for stuff like G2EGM, DCEGM, and whatever models out there use extreme value type I taste shocks to either smooth out the problem or to get non-degenerate policies for maximum likelihood estimation purposes.

I'm attempting to use the HARK features such as `distance_criteria` etc, but please tell me more places where I can use the features HARK has to offer.

As it stands, it's quite bare. It's just solving the model. If @llorracc / @mnwhite / others agree that such a model should be here, it would make sense to expand with simulation of data and estimation (will be very straight forward to do).

The model is very simple, but that has the added value that the example will be quite easy to follow for new-comers to Econ-ARK. This is why I am requesting suggestions above for incorporating HARK features where possible/advantageous.

As an added bonus, this made me realize that there were some bugs in the `discreteUpperEnvelope` function that is also in the G2EGM PR (#200), so if this PR is shot down, we'll at least have that :)

***Todo***
- [ ] add doc strings
- [ ] add coments
- [ ] add simulate method
- [ ] add estimate method
- [ ] small description of the model, assumptions, etc... it's all in Rust (1987)


pkofod:
Tried the Jupytext plugin to Jupyter as suggested by Chris. It is super easy to set up, and quite nice to work with in practice. So far I've not run into any annoyances. Certainly, it's better than `add`ing `.ipynb` files to the repo.

pkofod:
I'm going to archive this. We can do such a model at a later point, but all relevant code that carries actual functionality has been further developed elsewhere.

In the future, I will re-implement this specific model with the discrete choice solver being written.

PR #207, opened 2018-11-05
Add scipy to requirements.txt

pkofod:
scipy is required in HARK (.utilities) so it should be in requirements.txt

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,3 +5,4 @@ numpydoc
 future
 joblib
 dill
+scipy


mnwhite:
Kind of surprised this was missing.  Nathan tested installing HARK on a fresh install of Python, but maybe he always did it with Anaconda?  Merging now.

pkofod:
> Kind of surprised this was missing. Nathan tested installing HARK on a fresh install of Python, but maybe he always did it with Anaconda? Merging now.

I was surprised as you, and came to the same conclusion: this has probably not been tested on a bare Python install before (which is how I realized it was missing btw :) )

Issue #175, opened 2018-07-20
Changes to ConsIndShockModel in commit "9020084" have broken SolvingMicroDSOPs and possibly other applications


npalmer-professional:
If you run the following under the current codebase as of today (most recent commit is [43952d1](https://github.com/econ-ark/HARK/commit/43952d127a432d473e2509582393af5119f9d363) at the time of this writing):

    import HARK.SolvingMicroDSOPs.StructEstimation as struct
    struct.main() 

You will get this error: 

    TypeError: unsupported operand type(s) for /: 'float' and 'NoneType'

This appears to be due to changes implemented in [this commit](https://github.com/econ-ark/HARK/commit/902008460fbb1d4c2986b10234645727f2629199). To check, you can pull the [previous commit](https://github.com/econ-ark/HARK/commit/4b340003d36c48d33c4e2e6adc7b25437a8cb7ee)  (PR#172) and the above example lines will run fine. 

I am pretty sure some of the whitespace change that shows up in the [possible problem commit](https://github.com/econ-ark/HARK/commit/902008460fbb1d4c2986b10234645727f2629199) are the reason for this.

I will look into this in the morning. 

Until then we should *not* update the HARK up on pip/conda, because those appear to be made using the last working [previous commit](https://github.com/econ-ark/HARK/commit/4b340003d36c48d33c4e2e6adc7b25437a8cb7ee)  (PR#172).

mnwhite:
Was this addressed before the latest version was pushed to pip?

mnwhite:
It looks like those commits were reverted, and might have been pushed in error.  I don't think they could be the cause of current errors?

npalmer-professional:
Solved. Could not find a good way to clean up the history of my "ReplicateMake" branch, so did the following: 

Did a force reset for `npalmer-professional:master` to `econ-ark:master` for the most recent common commit after the reverts. Then created a new `ReplicateMake-Take3` branch, manually moved the changes to this new branch, and rebased/squashed `ReplicateMake-Take3` into `npalmer-professional:master`. 

Then I compared the histories between `econ-ark:master` and `npalmer-professional:master`  using 
    git log --pretty=oneline --graph --decorate --all
and everything looked good. I then checked the diff between `econ-ark:master` and `npalmer-professional:master`. Since both the histories and diffs looked as expected, I issued pull request #176 .

PR #208, opened 2018-11-05
Setup basic travis support.

pkofod:
Since I don't have the `econ-ark` keys, I need someone to log in to https://travis-ci.org/ and click the button to enable github integration for the travis-ci service.

I put in the simplest test module from `Testing` just to show how it works. If you want this, then I'd hope it could be merged with only these basic tests, and then I can add the rest later. I had problems running the some of the other test modules (kind of shows why we need this), but I'd rather use this infrastructure to add tests for some of the discrete choice functions that I'm using in the other PRs, and make PRs for those. Then I can come back to fix some of the problems with the other test modules.

I have never used travis for python projects, so it took a bit of experimenting, but a `tests` folder with an `__init__.py` file seemed to be simplest.

To add more `unittest` modules, you simply need to place them in `tests/test_,,,.py` files, and `pytest` (which is run by travis) will find them.

*edit*
Oh, and the python versions are somewhat arbitrary, we can add more (but 3.7 needs a hacky solution to work).

Travis is free for open source projects, and they give you buildbots that should more than suffice for a project with HARK's PR activity.

This will help avoid #175 in the future!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,9 @@
+language: python
+python:
+  - 2.7
+  - 3.6
+cache: pip
+install:
+  - pip install -r requirements.txt
+script:
+  - pytest HARK/tests
--- a/HARK/tests/test_initial.py
+++ b/HARK/tests/test_initial.py
@@ -1,0 +1,64 @@
+"""
+This file implements unit tests to check HARK/utilities.py
+"""
+from __future__ import print_function, division
+from __future__ import absolute_import
+
+from builtins import str
+from builtins import zip
+from builtins import range
+from builtins import object
+
+import HARK.utilities
+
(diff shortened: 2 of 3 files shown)


pkofod:
When/If Travis is enabled, please tell me so I can try to refire it from this PR.

econ-ark:
OK, I think I have successfully activated Travis.  (It's travis-ci.com, though, not travis-ci.org).  Does this let you start doing what you need to do, or are there further steps?

pkofod:
Is there any way this could be merged?

mnwhite:
Merging now.  I haven't tested it, but there's no way it can break HARK given what's been added and changed.

PR #212, opened 2018-12-12
Update FashionVictim to work as HARK package

llorracc:
This is a tiny fix needed to bring FashionVictim up to date with the changes that happened when HARK became pip-installable.  No longer right to say "from . import" -- instead it need to be "from HARK.FashionVictim import".


llorracc:
For release notes: subsumed under the other notes about fixing improper imports

PR #213, opened 2018-12-17
Modify the SCF file to pull in extra data when requested (in SolvingMicroDSOPs)

emmakalish:


mnwhite:
1) This doesn't read in extra SCF data, it reads in an alternate SCF dataset.

2) This is not a standard approach.  If there exists a directory named 'Download' three levels above the SolvingMicroDSOPs directory, then the code automatically looks there for the data file, rather than its normal location.  Why?  People don't do work from their download folder, and this would make the example application not work if they just so happened to put it in a place that satisfies this arbitrary file structure.

llorracc:
Thanks for this PR; we've revamped the file management system in a way that accomplishes these goals.

PR #217, opened 2019-01-18
Merge PKM's basic discrete choice tools

mnwhite:
This PR supercedes #209 and was made while adding an extra commit to that.  See that thread for history and discussion.


llorracc:
For release notes: calcLogSum, calcChoiceProbs, and calcLogSumChoiceProbs added to HARK.interpolation

Issue #182, opened 2018-07-30
Error in ConsIndShockModel when run as non-package


edmundcrawley:
When you run ConsIndShockModel by typing:
run ConsIndShockModel
in Spyder, you get an error:

    from . import ConsumerParameters as Params
ValueError: Attempted relative import in non-package

This error occurs when the code at the very end of the file is run:

if __name__ == '__main__':
    main()

This code is not normally run when the file is used as a package. I'm told that perhaps I should not be running this code as a non-package now, but if that is the case these lines of code at the very end should probably be removed. If not removed, then the bug should be fixed.

Nathan suggests that OTS will have some useful input on this point.

mnwhite:
Fixed, forgot to close issue.

PR #224, opened 2019-02-14
Fixed imports in model files

mnwhite:
All of the consumption-saving model files used a style of import that didn't work when the file was run directly (rather than called as a module), even though they have a __main__ block (and main() function).  This has now been fixed.

Also mostly removed extraneous file RepAgentModel.py, which seems to be an old name of ConsRepAgentModel.py.  This file now simply imports all of ConsRepAgentModel and warns the user to use that instead.


mnwhite:
This addresses #182 , which was also raised by CDC today.  We can also consider deleting the main() calls of all of these files and only having them in the examples that we have users put in their working folder.

llorracc:
My purpose in asking for the fix was to make the various modules like ConsIndShockModel.py "portable" -- so that they would run even if moved to a different filepath.  In attempting that with the revised version, I found that the lines containing "import ConsumerParameters" still did not work (from a directory other than ConsumptionSaving), but that everything worked if "import ConsumerParameters" was replaced with "import HARK.ConsumptionSaving.ConsumerParameters".  I have done that replacement not only for ConsIndShockModel.py but also for all of the other .py files in the ConsumptionSaving directory, and pushed the changes back up to the QuickImportFix branch.

mnwhite:
Ah yes, I was running them in place.  As is, the "from . import blah"
construction did not work.

llorracc:
For release notes: "Removed improper local package imports"

PR #229, opened 2019-02-26
Remove Chinese Growth and NonDurables since these now exist as DemARKs.

pkofod:


llorracc:
For Release Notes: This is subsumed in another "for release notes" description

shaunagm:
@llorracc which other "for release notes" description is it subsumed under?

mnwhite:
The general release notes should be:

"All standalone demo files have been removed. The content that was in these
files can now be found in similarly named Jupyter notebooks in the DEMARK
repository. Some of these notebooks are also linked from econ-ark.org."

PR #235, opened 2019-03-07
Fix for nan_bool in Py2

mnwhite:
Py2 can't accept named inputs after an *args list, but Py3 can.  Recent PR with nan_bool functionality for UpperEnvelope (etc) worked in Py2 but not Py3.  Attempting fix.

Not ready to merge, issuing a PR in order to use Travis tests.


mnwhite:
Travis tests pass, but easy fix doesn't work.  Will have to undo that merge.

PR #236, opened 2019-03-07
Undo nan_bool merge

mnwhite:
Tim Munday's nan_bool feature on LowerEnvelope (etc), which had been merged into master by me, worked on Py3 but not Py2.  Trying to fix by rearranging order of inputs failed, so this PR reverts that PR.

We still do want nan_bool, but will need to change the syntax of LowerEnvelope (etc) *and everywhere it's implemented* from 

`def __init__(self,*functions):`

to

`def __init__(self,functions):`

This only requires throwing brackets around all of the argument lists where we make a LowerEnvelope (etc), but we need to be thorough.  Should also discuss how to handle the fact that we're changing syntax of a long-standing function class, so *other people*'s code will break next time they update HARK, if they were using this.


mnwhite:
Merging myself because this is undoing a merge I did recently.

PR #237, opened 2019-03-11
nanbool

TimMunday:
reintroduce nanbool

### DIFF
--- a/HARK/ConsumptionSaving/ConsIndShockModel.py
+++ b/HARK/ConsumptionSaving/ConsIndShockModel.py
@@ -893,7 +893,7 @@ class ConsIndShockSolverBasic(ConsIndShockSetup):
         cFuncNowUnc = interpolator(mNrm,cNrm)
 
         # Combine the constrained and unconstrained functions into the true consumption function
-        cFuncNow = LowerEnvelope(cFuncNowUnc,self.cFuncNowCnst)
+        cFuncNow = LowerEnvelope(cFuncNowUnc,self.cFuncNowCnst, nan_bool = False)
 
         # Make the marginal value function and the marginal marginal value function
         vPfuncNow = MargValueFunc(cFuncNow,self.CRRA)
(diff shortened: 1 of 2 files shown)


mnwhite:
I asked Tim to reissue this PR to a branch I just created.  I'm going to try to figure out how to make this work in both Py2 and Py3 before merging to master.  The problem stems from my (shortsighted) decision to specify UpperEnvelope (etc) as `__init__(self,*functions)`, passing an arbitrary number of functions, rather than `__init__(self,functions)`, passing one list or tuple of functions.  Fixing this would (very slightly) change the syntax of an existing HARK class, so need to act with caution.

shaunagm:
Hi @TimMunday, I'm going through PRs and adding a "needs release notes" label. Can you provide a one sentence summary of your changes, ideally by editing the issue itself (as opposed to leaving a comment), clearly labeled with "For release notes" or similar? Remember to specify if there are breaking changes in the PR.

I'm not going to leave a comment like this on every merged PR of yours, but you can find a full list (well, full once I'm done tagging) [here](https://github.com/econ-ark/HARK/pulls?utf8=✓&q=is%3Apr+label%3A"needs+release+notes"+author%3ATimMunday).

TimMunday:
Hi @shaunagm what do you mean by "editing the issue itself( as opposed to leaving a comment)" ? Sorry, I'm not as pro at github as I should be by now...

shaunagm:
As the person who opened the PR, you should be able to scroll up to the very top of the page and edit the original/first post by clicking on the "..." in the top right corner.  But if you can't figure that out, totally okay to just leave it in the comments here.  It's easier for me to collect the notes if they're all in one place, but it's really not a hardship to scoll down to the most recent comments if I need to.

mnwhite:
This was a PR to a non-master branch; I don't think we need to have release
notes for it.  I will write a release note when I issue the PR for this
branch into master, which will involve a (minor) breaking change.

shaunagm:
I meant to only tag merges into master but I guess I missed this one.
Thanks for catching it.

PR #238, opened 2019-03-14
set numpy floating point error level to ignore.

pkofod:
@llorracc this should do the job. It ignores the floating point exceptions. Currently we only warn anyway, so it's not like we're handling it.

The exceptions aren't really errors either as numpy calls them. They're well-defined operations that give useful answers. Things such as `-1/-np.inf` gives `0`. This is very useful in DCEGM to interpolate from the lowest value to the first values of the resource grid that is not `0`. It could be special handled, but this is implemented all the way down to the hardware level, so I don't see why we shouldn't just use it.

### DIFF
--- a/HARK/core.py
+++ b/HARK/core.py
@@ -21,6 +21,11 @@ import numpy as np
 from time import clock
 from .parallel import multiThreadCommands, multiThreadCommandsFake
 
+# Ignore floating point "errors". Numpy calls it "errors", but really it's excep-
+# tions with well-defined answers such as 1.0/0.0 that is np.inf, -1.0/0.0 that is
+# -np.inf, np.inf/np.inf is np.nan and so on.  
+np.seterr(all='ignore')
+
 def distanceMetric(thing_A,thing_B):
     '''
     A "universal distance" metric that can be used as a default in many settings.
@@ -932,12 +937,12 @@ class Market(HARKobject):
         self.act_T = act_T
         self.tolerance = tolerance
         self.max_loops = 1000
-        
-        self.print_parallel_error_once = True  
-            # Print the error associated with calling the parallel method 
+
+        self.print_parallel_error_once = True
+            # Print the error associated with calling the parallel method
             # "solveAgents" one time. If set to false, the error will never
             # print. See "solveAgents" for why this prints once or never.
-        
+
     def solveAgents(self):
         '''
         Solves the microeconomic problem for all AgentTypes in this market.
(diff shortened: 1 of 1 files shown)


mnwhite:
One line added, other line changes are whitespace removal.

Does this need to be in our other files where numpy is imported, or does this act on numpy everywhere in the code?

pkofod:
> One line added, other line changes are whitespace removal.

Ah yes, that's my editor stripping white space at the end.



> Does this need to be in our other files where numpy is imported, or does this act on numpy everywhere in the code?

I tried using the REMARK PR notebook with this branch, and the errors were gone even in functions defined outside of HARK. I wonder though if this also means that if you check out two packages, one being HARK and the other being, say, pandas, you'll also change the setting of pandas?

mnwhite:
Hm, yeah.  Is that behavior that we want to foist on people?

pkofod:
Then we can just turn it in and off at the beginning and end of a solve call?

mnwhite:
Yeah, maybe solve, simulate, and update?  The "major" AgentType operations
where lots of work is done inside of HARK?

pkofod:
> Yeah, maybe solve, simulate, and update? The "major" AgentType operations where lots of work is done inside of HARK?

I'll try that instead.

pkofod:
How's this @mnwhite ? It uses error states instead, so every code in that `with` block *should* have those error setings. I need to run the notebooks to make sure, but I'm fairly sure this works as intended.

This last commit sets the error state for `solve` and `simulate`.

A side-question: I tried to "hide" the white space changes. When you look at the diff, do they still show up?

mnwhite:
The whitespace changes show up on github.com's diff display, but I don't actually care.  My comment was a note to self that there was only one line added, explaining why I was about to merge a PR with X hundred "line changes" a few minutes after it was submitted (my finger was hovering over the Merge button).

shaunagm:
Hi @pkofod, I'm going through PRs and adding a "needs release notes" label. Can you provide a one sentence summary of your changes, ideally by editing the issue itself (as opposed to leaving a comment), clearly labeled with "For release notes" or similar? Remember to specify if there are breaking changes in the PR.

I'm not going to leave a comment like this on every merged PR of yours, but you can find a full list (well, full once I'm done tagging) [here](https://github.com/econ-ark/HARK/pulls?q=is%3Apr+label%3A"needs+release+notes"+author%3Apkofod).

llorracc:
For release notes: "Suppress some worrisome but meaningless error messages"

llorracc:
Release note: Removed inappropriately scary warnings about non-problems

PR #243, opened 2019-03-19
Delete old demo files

mnwhite:
We have two folders inside of /ConsumptionSaving that contain model demos.  To my knowledge, all of these have been turned into DemARK notebooks other than TryAlternativeParameters, which has no explanation or documentation.  I'm not even sure when or why it was put into HARK.

This commit removes these old files.


shaunagm:
Hi @mnwhite, I'm going through PRs and adding a "needs release notes" label.  Can you provide a one sentence summary of your changes, ideally by editing the issue itself (as opposed to leaving a comment), clearly labeled with "For release notes" or similar?  Remember to specify if there are breaking changes in the PR.

I'm not going to leave a comment like this on every merged PR of yours, but you can find a full list (well, full once I'm done tagging) [here](https://github.com/econ-ark/HARK/issues?q=is%3Aopen+label%3A"needs+release+notes"+author%3Amnwhite).

llorracc:
Release note: Moved old "demo" files to DemARK or other appropriate places

PR #251, opened 2019-04-18
Prepare release 0.10.0.dev2

shaunagm:


shaunagm:
@llorracc this is [up on PyPI now](https://pypi.org/project/econ-ark/0.10.0.dev2/) and you should be able to use it in remarks, etc.

PR #254, opened 2019-04-24
Add a comment to the construction of aNrmNow

### LINKED ISSUE #253: Artificial borrowing constraint
If we want to impose a constraint on borrowing, say `a >= 0`, we use `BoroCnstArt` in the input dictionary.

Unless I am mistaken, does this not mean that https://github.com/econ-ark/HARK/blob/bf35222d3af40cdbd544f443fe071fc896744cc4/HARK/ConsumptionSaving/ConsIndShockModel.py#L796 should actually read
```
aNrmNow = np.asarray(self.aXtraGrid) + max(self.BoroCnstNat, self.BoroCnstArt)
```
?

This took me quite a long time to debug (longer than I'd like to admit, but I really wasn't looking here), because I thought I had set the borrowing constraint to >=0 in my portfolio model, but only quite late did I realize that my problems arose because assets were allowed to become negative, with... "interesting" solutions as a result :)

cc @mnwhite

pkofod:
fixes #253

### DIFF
--- a/HARK/ConsumptionSaving/ConsIndShockModel.py
+++ b/HARK/ConsumptionSaving/ConsIndShockModel.py
@@ -793,6 +793,12 @@ class ConsIndShockSolverBasic(ConsIndShockSetup):
         aNrmNow : np.array
             A 1D array of end-of-period assets; also stored as attribute of self.
         '''
+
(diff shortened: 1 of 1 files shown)


pkofod:
Clarify how the specification of borrowing constraints affect the internal representation of the consumption function.

PR #256, opened 2019-04-25
Simple test of  initialization of IncomeDstn of IndShockConsumerType.

pkofod:
cc @mnwhite @llorracc 

The other tests are also quite simple, but see here for a very simple example for how to add a test to the `HARK/tests` folder.

The template is:
- import `unittest` for the `TestCase` class
- define at least one (but you can do more in one file) class that inherits from `unittest.TestCase`
- optionally define a `setUp` method to set up variables before each `test_*` method is called (see below) (you can also define a `tearDown` method to destroy objects after each test
- define methods that start with `test_` and do the test with an assertion. You can do multiple tests in each class by defining multiple methods

Add files, commit, push, and it should be tested on CI!


mnwhite:
Thank you, this is exactly what I need for today's sprint.

pkofod:
> Thank you, this is exactly what I need for today's sprint.

Great, that was the motivation. I added a few comments about the `setUp` and `tearDown' methods - they are optional. Of course, I should also point to the documentation for good measure https://docs.python.org/3/library/unittest.html

mnwhite:
Yeah, I started reading that this morning.  Possibly dumb question: Does
Travis know to treat every file of the form /HARK/tests/test_*.py as a
unittest?  Or do we have to name/list those files individually somewhere?

pkofod:
> Yeah, I started reading that this morning. Possibly dumb question: Does Travis know to treat every file of the form /HARK/tests/test_*.py as a unittest? Or do we have to name/list those files individually somewhere?

Yes, it will run all those files automatically.

pkofod:
I think this should just be merged.

mnwhite:
Will do this morning. As you said on the other thread, we need init tests
for all of our classes as well.

shaunagm:
@llorracc or @mnwhite (I suspect @pkofod has gone to bed) - can you write a release note here?  It's the only thing blocking the dev3 release.

llorracc:
Release note: Add some unit tests for IndShockConsumerType

PR #265, opened 2019-05-06
add version dunder in init fixes #221

### LINKED ISSUE #221: module HARK has no attribute __version__
Best practice for modules seems to be that they have an attribute that lets users determine the version number.  This should be done (or whatever substitute is appropriate for conda) on the next public release.

MridulS:


mnwhite:
That's it!

PR #270, opened 2019-05-06
Modify hyperlinks to use relative links feature from GitHub

tanyaschlusser:
Modify file links to use [relative links][1] rather than absolute ones, because the CONTRIBUTING.md link was directing to a branch that has since been deleted.

Also hyperlink to code files listed at the end of the README, and delete links to files that no longer exist.

[1]: https://help.github.com/en/articles/about-readmes#relative-links-and-image-paths-in-readme-files

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ GitHub repository:    https://github.com/econ-ark/HARK
 
 Online documentation: https://econ-ark.github.io/HARK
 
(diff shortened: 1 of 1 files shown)


mnwhite:
Merging in commits from today and letting Travis run (should be no problems, but Travis has been weird lately), then will merge.

PR #274, opened 2019-05-06
Linted Comparison and Model Tests

StephenSchroeder:
Also added venv and vscode settings.json to .gitignore


shaunagm:
Looks good to me!  Thanks @StephenSchroeder.

PR #275, opened 2019-05-06
Add notebook testing.

hameerabbasi:


hameerabbasi:
You can probably move notebooks into separate dirs and run `--nbval-lax` only on things that succeed. 😄

llorracc:
@hameerabbasi, so this failed Travis because one of the DemARK notebooks took more than 10 minutes to run.  I wrote that notebook and am not surprised that it took that long -- it does some pretty serious calculations that could easily take more than 10 mins (depending on hardware).  So, basically, it looks like your contribution worked correctly but we've run into exactly the issues we were talking about earlier, about our needing to flag specific notebooks for testing and leave out others.  

There's probably a configuration file where that can be done?  Do you know how to do it?  One file that certainly ought to work is [Uncertainty-and-the-Saving-Rate](https://github.com/econ-ark/DemARK/blob/master/notebooks/Uncertainty-and-the-Saving-Rate.ipynb)

PS.  I've just realized that most of the .py files will not work for Travis because when you plot a file from the command line it pops up an interactive window and waits for the user to close the window before proceeding.  I've fixed that for a couple of notebooks, but need to do it for the rest.

hameerabbasi:
Unfortunately, other than separating notebooks by directory or testing them one by one, I’m not aware of a way.

hameerabbasi:
You could run `pytest -- nbval-lax file` to see why it fails locally.

llorracc:
We should ask the AlphaCruncher guys for the script they ran to test the notebooks and REMARKs.

MridulS:
DemARK testing has been moved here https://travis-ci.org/econ-ark/DemARK

Closing this PR.

Thanks @hameerabbasi !

PR #278, opened 2019-05-06
Simple lint edits commits for dcegm.py

rsaavy:
it still gets
`dcegm.py:155:25: E712 comparison to False should be 'if cond is False:' or 'if not cond:'
dcegm.py:161:26: E712 comparison to False should be 'if cond is False:' or 'if not cond:'
dcegm.py:175:50: E712 comparison to False should be 'if cond is False:' or 'if not cond:'
`

### DIFF
--- a/HARK/dcegm.py
+++ b/HARK/dcegm.py
@@ -7,6 +7,7 @@ choice models with (or without) taste shocks" by Iskhakov et al. (2016)
 import numpy as np
 from HARK.interpolation import LinearInterp
 
+
 def calcSegments(x, v):
     """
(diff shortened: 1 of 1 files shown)


llorracc:
I think we'd better leave this for @pkofod's judgment tomorrow.  Thanks!

shaunagm:
I'm going to go ahead and merge it.  I'll flag it so we can follow up and fix the last E712 errors once we hear back from Patrick.

llorracc:
OK, sounds good.

pkofod:
Thanks! Wrt the other things, I'd have to check first that it works with numpy arrays.

PR #281, opened 2019-05-07
Linting ConsGenInc~.py

StephenSchroeder:
Just linted another file! I also missed adding a return to the def in ConsAggShockModel.py when I switched from a lambda so I fixed that as well.


mnwhite:
I read this over, and it looks good.  Merging momentarily.

Issue #190, opened 2018-08-31
Add input checks for interpolator (and other?) classes


mnwhite:
As is, the `__init__` methods for the interpolator classes in HARK.interpolation do absolutely no input checking, which can lead to bizarre behavior and errors when illegal / incorrect inputs are passed.  Input checks can be added fairly easily.  These could include:

1) Type checking: Throw an error if something is not an array that should be, etc.

2) Size checking: Make sure that passed arrays have the right size relative to each other.

3) Other stuff: E.g. for LinearInterp make sure that xSearchFunc (etc) are correct if specified.

pkofod:
For any potential "sprinters", here's what the problem is:
```
Python 3.6.8 |Anaconda, Inc.| (default, Dec 30 2018, 01:22:34) 
[GCC 7.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from HARK.interpolation import LinearInterp
>>> x = [1,2,3]
>>> y = [3, 4]
>>> linterp = LinearInterp(x, y)
>>> linterp(1.1)
array(3.1)
>>> linterp(2.1)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/pkofod/anaconda3/lib/python3.6/site-packages/HARK/interpolation.py", line 55, in __call__
    return (self._evaluate(z.flatten())).reshape(z.shape)
  File "/home/pkofod/anaconda3/lib/python3.6/site-packages/HARK/interpolation.py", line 816, in _evaluate
    return self._evalOrDer(x,True,False)[0]
  File "/home/pkofod/anaconda3/lib/python3.6/site-packages/HARK/interpolation.py", line 777, in _evalOrDer
    y = (1.-alpha)*self.y_list[i-1] + alpha*self.y_list[i]
IndexError: index 2 is out of bounds for axis 1 with size 2 
```
What happens is that the `LinearInterp` object is allowed to be constructed *even if* `x` and `y` are of different lengths. This causes hard to find errors, because a potential user may not realize why on earth `linterp(1.1)` works but `linterp(2.1)` doesn't. We should fail early and hard here.

Basically, to close this issue, someone should add a type check on the inputs and are the sizes correct. @mnwhite may want to chime in on the classes and methods where this might be relevant.

keithblaha:
> Type checking: Throw an error if something is not an array that should be, etc.

[...]

PR #282, opened 2019-05-08
Add non_empty argument validator

keithblaha:
This prototypes a decorator-based approach to help address some issues noted in https://github.com/econ-ark/HARK/issues/190

Decorators for all kinds of common cases could be introduced to validate arguments


mnwhite:
Merging after checks finish running.

mnwhite:
Release notes: Adds initial version of validators.py, which contains decorators for adding input validity tests to functions.  Currently not implemented on any functions.

PR #284, opened 2019-05-10
Moved checkConditions out of init

mnwhite:
IndShockConsumerType.__init__ was calling its checkConditions method, but this caused many subclasses to fail, as their checkConditions method either throws a notImplementedError or runs into an attribute error.

CDC wants checkConditions called automatically, so this is now done in preSolve().  All subclasses have had a trivial preSolve method added, which only calls updateSolutionTerminal.


llorracc:
In some future version of HARK we should set thing up so that the agent knows whether the most recently calculated conditions are still applicable (which will be true if no parameters have changed).  We should also refine which checks are done so that they are correctly tailored to the individual model being solved.

llorracc:
Release notes: Fix bug caused by misapplication of check_conditions

PR #285, opened 2019-05-12
Fix StickyE crash caused by missing track_vars in Market init method

llorracc:
do_mid and do_all were crashing with a message "TypeError: __init__() got an unexpected keyword argument 'track_vars'".  Eventually traced this down to the omission of track_vars from the list of objects in the init method of the Market class.  It's a bit mysterious to me why this didn't cause lots more problems earlier.  I wonder if we should add **kwds to the Market init method?

### DIFF
--- a/HARK/core.py
+++ b/HARK/core.py
@@ -886,7 +886,7 @@ class Market(HARKobject):
     layer on top of the "microeconomic" models of one or more AgentTypes.
     '''
     def __init__(self, agents=[], sow_vars=[], reap_vars=[], const_vars=[], rack_vars=[], dyn_vars=[],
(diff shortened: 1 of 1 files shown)


mnwhite:
Which call to __init__ was doing this?  The code runs correctly on a very recent version of HARK, so something changed recently to trigger this.

mnwhite:
Okay, I found the problem.  A delinting commit during PyCon put spaces between all argument names in core.py, and the contributor misclicked on that line, mistakenly changing track_vars to rack_vars.

I'll make a new branch to make this change, so that the merged branch doesn't have a misleading name.

llorracc:
Superceded by #286

PR #286, opened 2019-05-13
Fix typo in Market.__init__

mnwhite:
Delinting at PyCon created a typo in the argument list for Market, now fixed.  This addresses the same issue as just-closed PR from CDC #285

### DIFF
--- a/HARK/core.py
+++ b/HARK/core.py
@@ -885,7 +885,7 @@ class Market(HARKobject):
     dynamic general equilibrium models to solve the "macroeconomic" model as a
     layer on top of the "microeconomic" models of one or more AgentTypes.
     '''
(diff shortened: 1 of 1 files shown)


pkofod:
This is sort of why I suggested at some point (and didn't do it myself) that it can be a good idea to simply gather all the model types in a test set and just `init` them all. Just to check that stuff still can be initialized after a change. It's a crude check, but it is quite important that all models can be initiated after all :)

llorracc:
Release notes: Bug fix

mnwhite:
Adjustment: No release notes needed.  This was a one character typo
introduced at PyCon and corrected a few days later.  There's no need to
mention this in the release notes.

PR #287, opened 2019-05-15
Fixes multithreading problem from ipython kernel

llorracc:
See if this works on Windows

### DIFF
--- a/HARK/core.py
+++ b/HARK/core.py
@@ -21,8 +21,7 @@ import numpy as np
 from time import clock
 from .parallel import multiThreadCommands, multiThreadCommandsFake
 
-
-def distanceMetric(thing_A, thing_B):
+def distanceMetric(thing_A,thing_B):
     '''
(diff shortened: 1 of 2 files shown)


mnwhite:
Chris and I discovered that HARK.parallel.multithreadCommands did not work on some of our project code... but only in some Python environments.  On Windows, I could only generate this error on Python 3, but everything worked correctly on Python 2.  On Mac and/or Linux, Chris found that the error came up when running Python in a terminal, but not when running Spyder; he also tested various web servers.

After some digging, Chris found that recent versions of joblib (which multithreadCommands uses) changed the default backend of joblib from multiprocessing to loky.  Apparently *something* in loky does not play nicely with our class structure, as it can't (de)serialize at least some AgentType subclasses.

This branch fixes this by simply changing the backend argument on the call to Parallel.  It also fixes one typo in a comment.

llorracc:
Release notes: Fix multithreading problems by using Parallels(backend='multiprocessing')

PR #293, opened 2019-05-23
Add more init tests

pkofod:
Just some more model init tests, making sure that our standard parameter dicts can be used to init models.


pkofod:
Merge?

mnwhite:
Yes.

PR #298, opened 2019-05-30
Fix minor error in conda install instructions

shaunagm:


llorracc:
For future reference: Google search says "source activate" will work, but "conda activate" is preferred (partly because it's much faster).

shaunagm:
I made this PR because `source activate` didn't work for me while `conda activate` did, but perhaps there was some error on my part.

PR #300, opened 2019-05-31
Improved parallelNelderMead

mnwhite:
Made some fixes to HARK.parallel.parallelNelderMead.  This was all done in my private fork of HARK and extensively tested on a real application.

1) Renamed maxcores to maxthreads, to be more correct
2) Fixed various small comments
3) Applied fix if degree of parallelization is illegal
4) Fixed computation of x_dist (whoops)
5) Allowed initial simplex to be saved to disk
6) Adjusted save/load procedure to use comma as delimiter rather than space (and use one line per simplex point), to make it easier to copy-paste points from the simplex file while estimation is in progress.

Also deleted superfluous test of parallelNelderMead in the main block, and trimmed down the message that's printed if HARK.parallel is ever executed.


HARK/parallel.py
```diff
@@ -335,30 +351,34 @@ def saveNelderMeadData(name, simplex, fvals, iters, evals):
...
     -------
-    none
+    None
     '''
-    f = open(name + '.txt','w')
-    my_writer = csv.writer(f,delimiter=' ')
+    N = simplex.shape[0] # Number of points in simplex
+    K = simplex.shape[1] # Total number of parameters
+    
+    f = open(name + '.txt','wb')
```
MridulS: This could be something like

``` python
with open(name + '.txt','wb') as f:
    my_writer = csv.writer(f,delimiter=',')
    my_writer.writerow(simplex.shape)
    my_writer.writerow([iters, evals])
    for n in range(N):
        my_writer.writerow(simplex[n,:])
    my_writer.writerow(fvals)
```
reduces the probability of missing the open and close of the file in future iterations.

HARK/parallel.py
```diff
@@ -370,28 +390,39 @@ def loadNelderMeadData(name):
     evals : int
         The cumulative number of function evaluations in the search process.
     '''
+    # Open the Nelder-Mead progress file
     f = open(name + '.txt','rb')
```
MridulS: same thing here.

I can update this if you want :)

mnwhite:
Yes, just make sure the delimiter doesn't change between versions, on the
read or write end. I think csv.reader figures out the delimiter from
context anyway.

     -- mnw

     -- mnw

mnwhite:
Thank you.  I know I'm *supposed* to do this to avoid dangling file writes, but I don't.

PR #302, opened 2019-06-13
Update docstring of simDeath() in core.

pkofod:
Docstring doesn't match behavior.

### DIFF
--- a/HARK/core.py
+++ b/HARK/core.py
@@ -561,7 +561,7 @@ class AgentType(HARKobject):
         '''
         Determines which agents in the current population "die" or should be replaced.  Takes no
         inputs, returns a Boolean array of size self.AgentCount, which has True for agents who die
-        and False for those that survive. Returns all True by default, must be overwritten by a
+        and False for those that survive. Returns all False by default, must be overwritten by a
         subclass to have replacement events.
 
         Parameters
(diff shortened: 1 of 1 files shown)


mnwhite:
Good catch! But my typo is in the opposite direction: the docstring says what it's *supposed* to do, but the np.ones is supposed to be an np.zeros.  We can also consider removing that print statement and letting "no death and replacement" be the default.

pkofod:
> But my typo is in the opposite direction: the docstring says what it's _supposed_ to do, but the np.ones is supposed to be an np.zeros.

I thought that might be the case, but I couldn't determine if you intended for everyone to die by default and thereby force people to respect the printed message, or if it was a typo the other way. In that case, I will change the behavior! (and add a test), and remove the printed message. Then noone dies, unless the users specifies it.

llorracc:
Sounds like "what to do" is now agreed on this?  Or is more discussion needed?

mnwhite:
I think Patrick already did it.  Let me take a look tonight and see if it's
ready to merge (probably is).

pkofod:
I removed the warning, and changed the default.

shaunagm:
@pkofod, can you write a release note for this?

pkofod:
"Fix unintended behavior in default simDeath(). Previously, all agents would die off in the first period, but they were meant to always survive."

This change is technically breaking. Before, there would be a warning that the users should define their own method and all agents would die. Now we simply assume that agents live forever, and people have to define a method to avoid this.

PR #321, opened 2019-06-19
Remove annoying message in checkConditions when model assumptions for…

### LINKED ISSUE #319: IndShockConsumerType init method prints unnecessary message from checkConditions
IndShockConsumerType now calls checkConditions() via update() when it is initialized.  checkConditions knows that it is only useful if T_cycle==1 and cycles==0, and does nothing otherwise... except print a message to screen that says, "This method only checks for the conditions for infinite horizon models with a 1 period cycle."

This was fine when checkConditions was not called internally, only from the command line as a teaching tool, but is weird (and unhelpful) now that it's run automatically.  The user will not know what "this method" is.  Someone running a lifecycle model does not want to see this message pop up every single time.

pkofod:
… checks to be valid are not met.

Fixes #319

### DIFF
--- a/HARK/ConsumptionSaving/ConsIndShockModel.py
+++ b/HARK/ConsumptionSaving/ConsIndShockModel.py
@@ -1496,8 +1496,8 @@ class PerfForesightConsumerType(AgentType):
         self.verbose        = verbose
         self.quiet          = quiet
         self.solveOnePeriod = solvePerfForesight # solver for perfect foresight model
-        
-        
+
+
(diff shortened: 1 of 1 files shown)


shaunagm:
@pkofod can you write a release note for this?  The title of the PR is not quite clear enough for me to understand what this does.

pkofod:
"Make checkConditions() less verbose when the checks are not actually performed by converting a print statement to an inline comment."

PR #323, opened 2019-06-20
Updatepresolve

### LINKED ISSUE #222: After income process parameters change, warn user to recompute .updateIncomeProcess()
It is inevitably confusing to new users that when they change some component of the income process like UnempPrb, the IndShockConsumerType consumer continues to use the old income process corresponding to the original parameterization.  We should at least have a warning to the user that they need to invoke [instance].updateIncomeProcess() to recompute the new income process.

pkofod:
Fixes #222 at the expense of a few more calculations each time `solve` is called. Advanced users can easily create a `MyIndShkType` and overwrite `preSolve` with a new 3-line method that doesn't update between `solve()` calls. The cost is probably negligible anyway, compared to solving the model and any post-processing you might do after solving a model.

### DIFF
--- a/HARK/ConsumptionSaving/ConsIndShockModel.py
+++ b/HARK/ConsumptionSaving/ConsIndShockModel.py
@@ -1496,8 +1496,8 @@ class PerfForesightConsumerType(AgentType):
         self.verbose        = verbose
         self.quiet          = quiet
         self.solveOnePeriod = solvePerfForesight # solver for perfect foresight model
-        
-        
+
+
     def preSolve(self):
         self.updateSolutionTerminal()
 
@@ -1711,10 +1711,10 @@ class PerfForesightConsumerType(AgentType):
         None
         '''
         if self.cycles!=0 or self.T_cycle > 1:
-            if verbose == True: 
+            if verbose == True:
                 print('This method only checks for the conditions for infinite horizon models with a 1 period cycle')
             return
-        
+
         violated = False
 
         #Evaluate and report on the return impatience condition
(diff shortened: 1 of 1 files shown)


mnwhite:
This has been one of the long-demanded items (as a one line fix). Gentle
intro notebook should be updated to reflect this change. That's part of my
notebook overhaul.

pkofod:
> This has been one of the long-demanded items (as a one line fix). Gentle intro notebook should be updated to reflect this change. That's part of my notebook overhaul.

Just verifying: you agree that this should be done and it should be done like this, right?

mnwhite:
Yes, please.

pkofod:
Great. Merge when you're ready :)

llorracc:
Ultimately, we will want to define a distribution object which "stores" the parameters needed to produce its approximation (like, that it is a lognormal with mean 0.02 and standard deviation of 0.1 produced by the equiprobable method of approximation using seven nodes).  That way, the current parameters can be compared to the stored one, and the distribution computed only if it has changed.  A quicker but uglier fix would be just to have the agent object store this info (instead of having a separate distribution object for it).  

@albop

pkofod:
> Ultimately, we will want to define a distribution object which "stores" the parameters needed to produce its approximation (like, that it is a lognormal with mean 0.02 and standard deviation of 0.1 produced by the equiprobable method of approximation using seven nodes). That way, the current parameters can be compared to the stored one, and the distribution computed only if it has changed. A quicker but uglier fix would be just to have the agent object store this info (instead of having a separate distribution object for it).

I agree that it could be done to make this slightly more efficient, but I'm wondering if the extra complexity is really worth it? I have strong doubts that re-calculating these things will be significant in the overall picture. We can try a specific implementation at some point to see if the extra complexity is worth it.

shaunagm:
@pkofod Can you write a release note?  Thanks.

pkofod:
Uhm yes. Something like.

"Added a call to updateIncomeProcess() in preSolve() to avoid solutions being based on wrong income process specifications if some parameters change between two solve() calls."

PR #324, opened 2019-06-20
Add a method (empty) for preSolve called checkRestrictions that can b…

### LINKED ISSUE #204: Restricting parameter space (throwing errors, warning, etcs)
If someone (well, that me right now, but in general) wants to add a model somewhere in econ-ark and implement a solution method that only works for some parameter combinations (say some relation between the discount factor and a risk free return). Is there an agreed-upon way to catch and handle such exceptions in econ-ark?

pkofod:
…e overwritten in classes inheriting from AgentType to check for illegal parameter values.

Fixes #204 by providing infrastructure for checking parameter restrictions.

This way of calling `AgentType.preSolve(self)` should be fine right @mnwhite ? Currently, all the classes that defined a `preSolve` failed to check for length time varying things for example, unless `AgentType.preSolve` was called somewhere I cannot see. The idea is basically that if you have any important restrictions such as values of CRRA, DiscFac, or something else, you write your own `checkRestrictions` method and raise an exception if the input fails to meet the requirements.

This way still allow users to construct the models, `checkRestrictions` is simply guarding calls to `solve()`.


pkofod:
Would you look at that. I caught a mistake thanks to the initialization tests  :) (I forgot to import AgentType)

llorracc:
@pkofod Thanks for adding this.  But it looks like you changed two things at once: Both this and the PR that I just merged modified the same place in IndShockConsumerType, so it is giving a "conflict" error.  

If you could resolve the conflicts and let me know, I will merge.

pkofod:
> If you could resolve the conflicts and let me know, I will merge.

Ah yes, that happens. I'll resolve the conflicts!

shaunagm:
Can you add a release note, @pkofod?

PR #325, opened 2019-06-20
Bayer luetticke iworld1991

iworld1991:
- Debugged  DCT-Copula notebook with the fix of indices-unraveling 
- I plot dct points using surface and full grids at scatter so that it is clear to see where they overlap.
- Include additional working version of the graphs for me to educate myself.
- I made a separate notebook called DCT simply doing the experiments using arbitrary c values at grids to see how DCT works and I plan to apply to the main notebook once I am done with experimenting.


llorracc:
BayerLuetticke has been updated subsequent to this PR, so a new PR needs to be constructed based on the revised BayerLuetticke

PR #327, opened 2019-06-21
Remove initialization of TranShkNow

### LINKED ISSUE #326: Default income shock for newborns
cc @mnwhite 

So, when we simulate an `IndShockConsumerType` and newborns arrive, we have to set initial permanent and transitory income shocks.

Current, it's sort of a weird situation. Transitory income shocks are initially set to 1.0 (the mean) but permanent shocks are set to a random draw from the permanent income distribution. So why/what/when?

It seems to me that this code was introduced #38 where we have these lines 
https://github.com/econ-ark/HARK/blame/9d0b1f4b2c883c562f60624df08636257c785562/ConsumptionSaving/ConsIndShockModel.py#L1566

As you can see, both shocks are initialized to 1.0, and then we update the shocks from there according to their distribution.

Then, in a commit that has no PR associated to it, the code changes to
https://github.com/econ-ark/HARK/commit/b4b473cc0c68ecadad096aee8a1d872a36a4db76#diff-b74561b6ba933aa9634d9b609ac72f36R2023

Now, both shocks come from their distribution from the first period.

And then, in a commit related to StickyE, and in the commit message it says
https://github.com/econ-ark/HARK/commit/fce7d45bb400aa6a56e404a813b7586fb0abb7b5#diff-b74561b6ba933aa9634d9b609ac72f36R1858

```
The SOE-Markov model is working in StickyE.  Had to give newborns 1.0
for TranShkNow, else change in log consumption was crazy.  Added
IdentityFunction to HARKinterpolation, which is a copy-paste from a
private MNW branch.  Will clean up the main file to make it a bit more
elegant.  Everything should be set for HA-DSGE versions, will do
tomorrow.
```
And transitory shocks are once again initialized to `1.0`.

My reading of this history is that at some point it was initialized to the mean, then it was changed to a draw (I guess to get a more interesting default data set), but then it was reverted because of the specific use-case in StickyE. This seems like something that should have only happened in the StickyE-specific folder/code. Or is there a more general reason that we want to initialize everyone with the same transitory income?

pkofod:
fixes #326 

I think it makes no sense to hard-code the initial transitory income shock. If it is super important from a practical point of view, we should simply make the very first distribution a discrete distribution with one point (`1.0` ofc). Otherwise the simulations don't really match the model specification.


mnwhite:
The problem goes the other direction, in my view. Lifecycle models in HARK
have no input for what the distribution of income shocks at t=0 should be.
The object ThisType.IncomeDstn[0] is what the *solver* needs to know for
t=0, which is the distribution of income shocks at the start of t=1.

Permanent income shocks at t=0 can be handled by the distribution specified
with the pLvlBlahInit variables-- there is no difference between initial
pLvl distribution and "shocks" that occur between birth and the first time
they could be "observed", so they're lumped together.

Transitory income shocks are more difficult, because we have no alternate
way to specify it. Moreover, there are some papers (including by us) that
assume no transitory shock in the very first period. Among other things,
this simplifies ex ante welfare analysis because you only consider the
value function at exactly one state.

In the big overhaul of HARK, this is one of the timing quirks that needs to
be worked out. We would like to be able to specify things so that shock
distributions at the beginning of t=0 are put into an instance like any
other period, but our solver framework knows that it's not actually useful
for solving the problem.

pkofod:
> The object ThisType.IncomeDstn[0] is what the *solver* needs to know for t=0, which is the distribution of income shocks at the start of t=1.

I see you point. The model is solved conditional on the initial condition.

Then why don't we introduce a distribution of initial (t_age = 0) income shocks that is only used in these lines - that is it is only used for simulation. It's not necessary for `solve()` to run. The default can be IncomeDstnNow[0] for permanent income shocks and degenerate `0` or `1` for the transitory shocks. Optionally, they can be given as `T_sim` length arrays such that newborns are endowed with initial conditions the evolve as `t` in the simulations.

I think an explicit "initial shock distribution for simulation" would be both useful and easy to implement.

pkofod:
In the talk today, we agree on a solution as sketched above, so I will move forward with that. I'll keep the default as a degenerate "draw" of 1, but allow users to more flexibly control it with a distribution if they want.

llorracc:
@pkofod what is the status of this?

PR #337, opened 2019-07-04
new installation instruction

janrosa1:


mnwhite:
I'm merging this momentarily.  Sorry it sat for so long.

PR #338, opened 2019-07-05
Comparing c functions for different number of basis functions(accuracy levels). 

iworld1991:
This PR includes three major changes since last meeting. 

1. In the DCT-Copula-Illustration.ipynb, I finally managed to plot the DCT approximated consumption function at all the grids(basically inverse DCT back).  This is done in both 3d graph (fornon-adjusters and adjusters), and addtional 2d graphs, which is the same idea as page31/32 in Ralph's slides. Importantly, I plot the comparisons for a range of accuracy levels (thus different number of basis functions used).

2. I created a new notebook called DCT-Copula-Illustration-OneAsset for redoing the two-asset experiments. However, it seems to me that the one-asset state reduction does not invoke DCT operation, or at least does not do it in the same way.  So DCT is less of a focus there.  

3. The DCT.ipynb now can serve as an independent  illustration of DCT with any arbitrary consumption function at full grids.


iworld1991:
ddd

Issue #340, opened 2019-07-06
BayerLuetticke notebooks should import instead of copy-paste


llorracc:
The notebooks in HARK/BayerLuetticke/ were created by copying and pasting the relevant bits of code from the source files [FluctuationsOneAssetIOUsBond.py](https://github.com/econ-ark/HARK/blob/master/HARK/BayerLuetticke/BayerLuetticke_code/OneAssetCode-HANK/FluctuationsOneAssetIOUsBond.py) and [FluctuationsTwoAssets.py](https://github.com/econ-ark/HARK/blob/master/HARK/BayerLuetticke/BayerLuetticke_code/TwoAssetCode/FluctuationsTwoAsset.py).  Instead, they should import the necessary tools and use them that way, so bugs fixed in the core code will automatically fix problems in the notebooks.

PR #342, opened 2019-07-06
make BayerLuetticke importable and rewrite Assets notebook

MridulS:
This is still work in progress
TwoAsset notebook is updated.

- [X] Make BayerLuetticke a module so its importable
- [X] Moved all notebooks in a new folder and reorganise code
- [x] Edit notebooks, remove code, import functions

The current folder structure is 
```
- BayerLuetticke
   -- notebooks
   -- OneAssetCode_HANK
   -- OneAssetCode_KS
   -- TwoAsset
```

@llorracc thoughts?

Fixes https://github.com/econ-ark/HARK/pull/340


llorracc:
So the OneAssetCode_HANK and OneAssetCode_KS directories are virtually identical, and were created at the last minute before a presentation when I realized for some reason that the two experiments could not be run sequentially.  This is one of the many things that would be fixed by making the code modular.  So, a better structure would just be 

```
- BayerLuetticke
   -- notebooks
   -- code
       -- OneAsset
       -- TwoAsset
```
or even

```
- BayerLuetticke
   -- notebooks
   -- code
       -- Assets
          -- One
          -- Two
```
which leaves room for "Three" or "Many" later

MridulS:
@llorracc What is the difference between TwoAssets and TwoAssets-uncertainty notebook? They seem to have the exact same content.

llorracc:
I created the -uncertainty notebook just to preserve the figures and other outputs so I could show them to an audience which would not have time to wait while they were generated.  The -uncertainty notebook should be deleted now.

MridulS:
@llorracc Should I merge in HANK-MP, HANK-TFP, HANK notebooks as only the shock variable is changing and we can have that in 3 different cells in the notebook?

llorracc:
Yes, the three HANK notebooks should be merged -- the reason there are three of them, though, is that I found that doing each experiment after doing one of the others generated screwy results that were not right and not what you got if you only did one experiment at a time.  3 copies were a quick-and-dirty way around that, and it would be better by far to have them all in one notebook, but be sure that you get the same results from running the experiments sequentially as running each alone.  (I tried to figure out what variables to clear to accomplish this but in the 2 minutes I spent on it couldn't figure out the answer).

MridulS:
Correct me if I am wrong but I think there is something seems off in the original notebooks. For example in 
https://github.com/econ-ark/HARK/blob/master/HARK/BayerLuetticke/OneAsset-HANK.ipynb the variables like `aggrshock`, `rhos`, `sigmas` are changed after the class `FluctuationsOneAssetIOUs` is already instantiated and state reduction is already done. So when the SGU_Solver is called it doesn't take any the changes to the variables in account. 

Variables should be set before doing the state reduction.
```
EX2SS=pickle.load(open("EX2SS.p", "rb"))
EX2SS['par']['aggrshock']           = 'Uncertainty'
EX2SS['par']['rhoS']    = 0.84    # Persistence of variance
EX2SS['par']['sigmaS']  = 0.54    # STD of variance shocks

EX2SR=FluctuationsOneAssetIOUs(**EX2SS)

SR=EX2SR.StateReduc()

SGUresult=SGU_solver(SR['Xss'],SR['Yss'],SR['Gamma_state'],SR['Gamma_control'],SR['InvGamma'],SR['Copula'],
                         SR['par'],SR['mpar'],SR['grid'],SR['targets'],SR['P_H'],SR['aggrshock'],SR['oc'])

plot_IRF(SR['mpar'],SR['par'],SGUresult['gx'],SGUresult['hx'],SR['joint_distr'],
             SR['Gamma_state'],SR['grid'],SR['targets'],SR['os'],SR['oc'],SR['Output'])

```

Thoughts?

llorracc:
Could well be -- will try to look at it later today.

llorracc:
Looks to me like you're right about this.  

@lofie21 , @edmundcrawley, could you take a look and see if Mridul is right about this?  Like, shouldn't the 

SR=EX2SR.StateReduc()

step be after the 

EX2SS['par']['aggrshock']           = 'Uncertainty'
EX2SS['par']['rhoS']    = 0.84    # Persistence of variance
EX2SS['par']['sigmaS']  = 0.54    # STD of variance shocks

steps?  Or, does the state reduction occur across experiments as well as within them?

edmundcrawley:
Yes, Mridul is right. The variables need to be set before the state
reduction.

MridulS:
I have updated both the notebooks and merged One-Assets-HANK-* and now the current structure of the folder is

```
- BayerLuetticke
   -- Assets
        --- One
        --- Two
   -- notebooks
```

Also I would suggest moving these notebooks out of this folder, maybe in the examples folder of HARK or even an independent repo. Ideally there should only be code inside the HARK/HARK folder.

shaunagm:
@MridulS or @llorracc - can one of you summarize this PR in one sentence for our release notes?

llorracc:
Release notes: BayerLuetticke HANK tools now importable as a module

PR #345, opened 2019-07-09
Finalizing the StE version 

iworld1991:
There are 24 commits ahead of the most recent master branch. I have also integrated the changes to other files in the master. 

** Before the Release**
-  Modified DCT-Copula-Illustration notebook so that it imports separately stored codes, instead of copying it.  
  -  The codes work just fine. But one thing not entirely clear to me is that I need to make the modification to the directory path so that I can run the following code. 

`from HARK.BayerLuetticke.Assets.Two.FluctuationsTwoAsset import FluctuationsTwoAsset`

-  No longer need to import a lot of libraries. So only import the necessary ones. 
-  Small edits to the formatting, fixing some magic numbers. 

**Most recent after meeting on July 16**
- Now the notebook takes user's input on accuracy level as well as trimmed population density. 
- Z-axis of certain graphs is re-scaled to show the variations. 
- It is confirmed that the 90% m and k is chosen contingent on h. The reason why now the scale of m and k are the same across h(0) to h(2) is that it just happens to be so for 90%. It picks different scale of m and k for, say 60% and other percentiles. 

**New ones**
- For each 3d graph, contour plots of joint density of m-k is plotted in the x-y plane. Redness is normalized so that they are comparable across h. Thus not surprisingly, high h value has very light red as there are very few agents in those values. 
- All labels are included. 3d surface plot has bugs in generating legends. Thus I used fake 2d line as auxiliaries.
- Instead of presenting 100% and 90% population density plots separately, I made the notebook to take user's input of the percentile to generate graphs. 
- I have not found the median value of h to normalize m and k yet. Will do it soon. 

**Old ones** 
- I copied the two files to the correct place, i.e. BayerLuetticke folder, thus overwriting the old ones; also deleted the two in the root folder. 
- Below each 3d graph, I plot the trimmed version where 90% agents are distributed. 
- Add plot of difference in approximation errors for adjusters and non-adjusters.  
- Deleted image plot and consumption function for adjusters and non-adjusters. 
- Improved formatting, title and labels.
- No more magic numbers. 

Finally, this is also a test of the new working flow. I allow the maintainers to edit the PR.


llorracc:
Release notes: Added notebook illustrating dimensionality reduction in Bayer and Luetticke

PR #350, opened 2019-07-13
Linting and resolving error messages

jdice:


shaunagm:
Sorry to take so long to review this, @jdice!  Travis says that your PR passes in Python 3.6 but not 2.7, and gives this specific error message:

Screenshot from 2019-07-17 14-21-10

It looks like `assertRaisesRegex` was added in 3.1, so that explains why it's failing in 2.7.  We have a few options here:

1) implement a check so the tests use `assertRaisesRegex` in 3.1+ and `assertRaisesRegexp` in 2.7
2) find an alternative to `assertRaisesRegex` that works in both 3.2 and 2.7
3) stop testing in 2.7, which is close to end of life (but not actually there yet)

Option 2 is the most elegant, but easier said than done.  Option 1 is the easiest.  Option 3 is oh so tempting but we probably shouldn't do it yet.

shaunagm:
Hi @jdice!  Would you like to rebase this pull request to incorporate changes that have happened in master in the meantime, and we can merge it?  Feel free to squash down my multiple commits while you're at it, if you like.

If you're not familiar with github, I can also handle this for you, just let me know what you prefer.

shaunagm:
Addressed by https://github.com/econ-ark/HARK/pull/374

PR #351, opened 2019-07-13
linted parallel.py

ingridbgl:


shaunagm:
I'm assigning this to Matt because Ingrid and I did a bit of code re-writing in order to pass the linter (see comment on the code itself).  Matt, can you check that the code still does what you want it to?  No rush.  :)

HARK/parallel.py
```diff
@@ -81,22 +83,24 @@ def multiThreadCommands(agent_list,command_list,num_jobs=None):
...
 
     # Default number of parallel jobs is the smaller of number of AgentTypes in
     # the input and the number of available cores.
     if num_jobs is None:
-        num_jobs = min(len(agent_list),multiprocessing.cpu_count())
+        num_jobs = min(len(agent_list), multiprocessing.cpu_count())
 
     # Send each command in command_list to each of the types in agent_list to be run
-    agent_list_out = Parallel(backend='multiprocessing',n_jobs=num_jobs)(delayed(runCommands)(*args) for args in zip(agent_list, len(agent_list)*[command_list]))
+    delayed_commands = [delayed(runCommands)(*args) for args in zip(agent_list, len(agent_list)*[command_list])]
```
shaunagm: This is where Ingrid and I rewrote the code a little bit - it was the only way to keep from breaking the line length limits.  I'm not 100% sure I didn't change the meaning of the code though.  Your original syntax here is very novel to me.

mnwhite: If no new line was added beyond that, then this won't do anything.  It will just create a list of stuff to do, and then never do it.

shaunagm: @mnwhite just to make sure - are you seeing line 96, with the code `agent_list_out = Parallel(backend='multiprocessing', n_jobs=num_jobs)(delayed_commands)`?  My understanding is that this runs the delayed commands, but you understand the code better than I.

llorracc:
This needs to be tested; @MridulS will test it in the context of the parallel problems with cstwMPC and either accept it or delete it.

MridulS:
This will work but instead of creating a new var and storing it in memory, we can use the following version for the linted version
``` python
    agent_list_out = Parallel(n_jobs=num_jobs)(
        delayed(runCommands)(*args)
        for args in zip(agent_list, len(agent_list) * [command_list])
    )
```

MridulS:
Closing in favour of https://github.com/econ-ark/HARK/pull/465

PR #352, opened 2019-07-15
rm OpenCL from master because flaky

llorracc:
The OpenCL stuff was put in the master branch temporarily for SciPy2019 in case somebody there could help figure out why it only works on a few configs.  We didn't find anybody to help with this, so this PR removes it again.  This code is now in limbo (on a branch OpenCL-20190710-Start) until/unless we find someone who can fix it.


shaunagm:
Just noting that the PR this PR reverts is #348

PR #354, opened 2019-07-16
Remove jupytext *.py files

llorracc:
Confusing situation was caused because @Mridul was not using jupytext and @llorracc was.

With jupytext active, the *.ipynb file is generated from the corresponding *.py file, so when @llorracc opened the "new" ipynb files the old one was autogenerated.

Immediate solution is to delete the *.py files, so that when the new *.ipynb files are opened a new *.py file is generated.  

Longer term lesson is to be sure everybody is using jupytext!


shaunagm:
I'm a little confused.  Is the .ipynb file generated from the corresponding .py file, or vice versa?

llorracc:
Release notes: None

PR #363, opened 2019-07-22
Undo June 20 commit to preSolve AgentType because it breaks things

llorracc:
@pkofod's commit on June 20 broke the KrusellSmith notebook among others -- this PR removes the lines that caused the problems (and also the functionality that they were intended to accomplish).  We need to revisit.


pkofod:
Is there an issue to reference? How did it break? I guess I can just add the lines back and check myself.

llorracc:
No, that was written by Pablo's student Jan.   I've also had trouble finding Karsten's thing; I'll email him.

PR #367, opened 2019-07-22
Also fix ConsAggShockModel

llorracc:
Missed a required fix in ConsAggShockModel like the others


llorracc:
Need to check for other instances in all Cons*.py files.

PR #377, opened 2019-07-27
move to default loky joblib backend, so lambda functions are picklable

MridulS:


llorracc:
This fixes a problem that prevented the Krusell-Smith DemARK from using multiprocessing.  

Release notes: Fix problems that sometimes prevented multiprocessing from working

PR #381, opened 2019-08-08
This adds all the Python classes to the Sphinx docs

ericholscher:
I imagine some of these should not be added,
but I'm not sure which ones :)


shaunagm:
@llorracc or @mnwhite - can you take a quick look and tell us if any of these classes shouldn't be in the docs?

shaunagm:
Merging this - we can remove unnecessary classes as needed.

PR #382, opened 2019-08-08
Add a submodule of DemARK and include a notebook in the doc build for it

ericholscher:
An example is online here: https://test-hark.readthedocs.io/en/add-submodule/submodules/DemARK/notebooks/KinkedRconsumerType.html

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "Documentation/submodules/DemARK"]
+	path = Documentation/submodules/DemARK
+	url = https://github.com/econ-ark/DemARK
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -11,7 +11,7 @@ sphinx:
 
 # Optionally build your docs in additional formats such as PDF and ePub
 formats:
-  - pdf
+  - []
 
 # Optionally set the version of Python and requirements required to build your docs
 python:
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -10,8 +10,7 @@ sphinx:
   configuration: Documentation/conf.py
 
 # Optionally build your docs in additional formats such as PDF and ePub
-formats:
-  - []
+formats: []
 
 # Optionally set the version of Python and requirements required to build your docs
 python:
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -18,3 +18,7 @@ python:
   version: 3.7
   install:
     - requirements: requirements.txt
+
+submodules:
+  include: all
+  recursive: true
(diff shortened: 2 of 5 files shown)


shaunagm:
This looks good to me.  @llorracc, any objections to merging it?

shaunagm:
Actually Eric, [this guide](https://github.com/econ-ark/HARK/blob/a55c0d5861b70e66bafb43513d3b3f21389490d1/Documentation/notebooks/README.rst) is not quite detailed enough for me.  Where do I put `ln -s ../../Examples/Gentle-Intro-To-HARK.ipynb .`?

ericholscher:
@shaunagm I've updated the guide with a bit more detail. 

> Or - was the goal here just to test that the notebook works, and we don't want to actually import the demarks/remarks, just the "gentle introduction to HARK" and similar?

I'm not sure what the best path forward is for the structure of the project's docs. This is mostly just an example of how you can include a submodule in your documentation. I generally like having submodules if the two things are truly logically separate, and generally like co-mingling the content if they are more logically similar. I don't have enough context to really know for this use case.

llorracc:
Thanks for this update -- seems like it should be very straightforward.  

I'd like to think of a way to set it up so that we can have some metadata embedded in each notebook that indicates whether it should be part of the documentation.  That way it will be very easy to add and remove things as new material becomes available and old material gets superceded.  

One more question: After we add new links, is there some procedure for recompiling the docs?  Or, do they recompile automatically nightly?  Or what?  

PS.  I've just made a PR with my first attempt to include LaTeX math in a docstring.  I tested it using your cool "Online Sphinx editor" and it seemed to work -- I guess that's the most reliable way to pretest whether an edit to a docstring is going to work?

shaunagm:
Thanks for the updates, Eric.

Chris, my understanding is that once we add the links, we'd need to push the changes to git, but that readthedocs would the automatically updating the docs once the changes are merged on Github.

(I just had to google to confirm that symbolic links could be tracked via git.  It makes sense now that I understand that, but part of my confusion was that I'd never encountered symlinks in git before.)

Final question from me, for Chris: do we want KinkedRConsumerType in the docs, or should we remove it before merging this request?

shaunagm:
I'm going to go ahead and merge this.  We're having a convo elsewhere about which notebooks to include, once we've got that figured out we can add more as needed - and probably remove KinkedRConsumerType, but I don't think it's a big deal to include it in our docs for a few days or weeks.

shaunagm:
Hi @ericholscher, I'm sorry, I'm still not getting this.  Can you walk me through the steps again?  Let's say I want to add [GenIncProcessModel.ipynb](https://github.com/econ-ark/DemARK/blob/ad40b284e3b9320daac0e4b5d75a3335072b4f12/notebooks/GenIncProcessModel.ipynb) from the DemARK repository to our documentation.  Where would I make the symlink, and what would the symlink be _to_?

shaunagm:
@ericholscher when you get a chance, can you provide a bit more instruction?

PR #388, opened 2019-09-17
Labor int marg

JackShiqiLi:
1. Finished Moving the Labor model parameters into the ConsumerParameters.py file and deleted the ConsumerParametersTM.py file.
2. Took the Lifecycle example from ConsIndShockModel and put it into the ConsLaborModel. (Is this what you want or did I misunderstood it?)
3. Tried to rewrite the getControls and getPostStates functions for the ConsLaborModel for simulation. But I was not able to fully understand the getControls function and the consumption function in ConsLaborModel. What I did was trying to mimic the getControls function in the ConsIndShockModel and got something similar. Please review carefully on this.


llorracc:
@JackShiqiLi I'm not sure what Matt said but I definitely do NOT want to _move_ the life cycle example from ConsIndShockModel to ConsLaborModel because some infrastructure has already been built expecting it to be in ConsIndShockModel.  

The preferred way to approach this would be to have a single version of the tool in which the user could specify a set of parameters, and any parameter that was not explicitly set would take on a default value.  If the default value of the elasticity of labor supply were zero, then the model would be mathematically identical to the original version.  If the user deliberately sets a labor supply elasticity different from zero, the modified code could be used.  

For the HARK 1.0 redesign, we need to think hard about how to generically handle this set of issues.  For the short run, though, I think having one function that solves both problems depending on labor supply elasticity is the way to go.

JackShiqiLi:
> @JackShiqiLi I'm not sure what Matt said but I definitely do NOT want to _move_ the life cycle example from ConsIndShockModel to ConsLaborModel because some infrastructure has already been built expecting it to be in ConsIndShockModel.
> 
> The preferred way to approach this would be to have a single version of the tool in which the user could specify a set of parameters, and any parameter that was not explicitly set would take on a default value. If the default value of the elasticity of labor supply were zero, then the model would be mathematically identical to the original version. If the user deliberately sets a labor supply elasticity different from zero, the modified code could be used.
> 
> For the HARK 1.0 redesign, we need to think hard about how to generically handle this set of issues. For the short run, though, I think having one function that solves both problems depending on labor supply elasticity is the way to go.

Thank you Carroll for the comment and guideline, I will look into this.

mnwhite:
I didn't say to move the lifecycle model from ConsIndShock to
ConsLaborIntMarg, I said to make a lifecycle example *like* the
ConsIndShock one. Models of labor supply are usually in a lifecycle
framework, not infinite horizon. We just want a functional example of that.

JackShiqiLi:
Sorry that I have misunderstood, I will finish it soon. Really sorry for that.

JackShiqiLi:
I have opened a new pull request to view it more clearly.

PR #391, opened 2019-09-19
Fix AdjustPrb for discrete portfolio choice.

### LINKED ISSUE #389: ConsPortfolioModel does not work with AdjustProb < 1
@pkofod:

@MridulS and I spent the morning going over the ConsPortfolioModel and making sure it all worked.  We've made a [DemARK](https://github.com/econ-ark/DemARK) called ConsPortfolioModelDoc.ipynb mostly based on the demo file you had provided.

But we could not get one thing to work:  You had an example of the version that allows the probability of adjustment to be less than one, but it was calibrated with the probability equal to one.  When we tried to run it with AdjustProb < 1, it crashed with an error that looked like it was trying to evaluate utility at consumption of zero.  

Am guessing this will be an easy one for you to fix, but it wasn't so easy for Mridul and me to track down the problem, so I've assigned it to you.

pkofod:
Fix #389 

So there were some leftover lines from a different way of constructing the re-curving value function interpolation.

A thing I noticed while trying this out to be sure it worked. If you set the probability to 0.5 and look at the share of periods where the agents can adjust, it'll be closer to 0.6 than 0.5. It could maybe be interesting to ask students to calculate the empirical share, and comment. (The answer is that when a new agent is replacing an old one whose lifecycle ended or who died, a new agent will always get the choice of their initial portfolio which implies a slightly higher "can adjust" ratio than 0.5).

### DIFF
--- a/HARK/ConsumptionSaving/ConsPortfolioModel.py
+++ b/HARK/ConsumptionSaving/ConsPortfolioModel.py
@@ -226,8 +226,6 @@ def _calcwFunc(AdjustPrb, AdjustCount, ShareNowCount, vFunc_adj, CRRA):
     integrand : function (lambda)
         Can be used to evaluate the integrand and the sent to a quadrature procedure.
     '''
-    if AdjustPrb < 1:
-        raise Exception('The code allowing probabilities of adjustment less than 1 is not working yet; release notes will indicate when we have a release where it work')
 
     # AdjustCount could in principle just be inferred from AdjustPrb instead of
     # cartying it arround FIXME / TODO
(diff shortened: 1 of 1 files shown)


HARK/ConsumptionSaving/ConsPortfolioModel.py
```diff
@@ -244,9 +242,6 @@ def _calcwFunc(AdjustPrb, AdjustCount, ShareNowCount, vFunc_adj, CRRA):
             evVals = AdjustPrb*vFunc_adj[0][ShareIndex](evalgrid) + (1-AdjustPrb)*vFunc_adj[1][ShareIndex](evalgrid)
             with np.errstate(divide='ignore', over='ignore', under='ignore', invalid='ignore'):
                 evValsNvrs = utility_inv(evVals,gam=CRRA)
-            evVals = np.insert(evVals, 0, 0.0)
-            evValsNvrs = np.insert(evVals, 0, 0.0)
-
```
MridulS: Yeah, this was exactly where the error was popping in from. If there was a divide by zero error which was ignored by `np.errstate` there was a mismatch between the length of `evVals` and `evValsNvrs`

PR #393, opened 2019-09-23
Inherit all attributes from Market Class in cstwMPC

MridulS:
Fixes https://github.com/llorracc/cstwMPC/issues/1


llorracc:
Release Note: Avoid bug caused when using parallel.py in Windows (by avoiding parallel.py)

Issue #359, opened 2019-07-22
"No solution" error is not reached in PerfForesightConsumerType for inf hum wealth


llorracc:
Adding these two cells to the Gentle Introduction - Perfect Foresight notebook does not result in the warning message that the model has no nondegenerate solution, but instead gives the error messages below.

NewExample.PermGroFac = [1.05]
NewExample.solve()

/dhlib/lib/python3.7/site-packages/HARK/interpolation.py:774: RuntimeWarning: divide by zero encountered in true_divide
  alpha  = (x-self.x_list[i-1])/(self.x_list[i]-self.x_list[i-1])
/dhlib/lib/python3.7/site-packages/HARK/interpolation.py:777: RuntimeWarning: invalid value encountered in multiply
  y = (1.-alpha)*self.y_list[i-1] + alpha*self.y_list[i]

JackShiqiLi:
I find that the method checkConditions were defined in the PerfForesightConsumerType but not called before SolveAgent() was called. I have Overwritten PreSolve() in PerfForesightConsumerType to allow for checkConsitions() to work in pull request #395 .

sbenthall:
Has this been fixed?

llorracc:
Yes.

PR #395, opened 2019-09-26
Tackle HARK issue #359

JackShiqiLi:
Overwritten preSolve() in PerfForesightConsumerType to call checkConditions().
Previously, checkConditions() were defined but not called before SolveAgent(), now setting PermGroFac such that PermGroFac/Rfree > 1 will output the message in checkConditions().
Problem: checkConditions() does not stop the solving process, so the RuntimeWarnings mentioned in issue #359 still appear. Maybe we should stop the solving process if the conditions are not met?


JackShiqiLi:
With a further look I found that my code was out of date, in the code here there is a preSolve() overwritten at line 1540, except that it does not call checkConditions(), so simply add checkConditions() to preSolve() will do the work.

llorracc:
It looks like this test is performed in all cases.  It only makes sense in the case of the infinite horizon solution of the model.  Could you add a test for whether the user has invoked it with the appropriate values of the various "cycles" settings to require an infinite horizon solution?

JackShiqiLi:
I think the test for the 'cycles' has already been written in the checkConditions() function at lines 1762 to 1765.
```python
    # This method only checks for the conditions for infinite horizon models
    # with a 1 period cycle. If these conditions are not met, we exit early.
    if self.cycles!=0 or self.T_cycle > 1:
        return
```
My trial in the Gentle-Intro-To-HARK-PerfForesightCRRA notebook shows that if you add another line `NewExample.cycles = 1` ( 1 can be replaced by any integers > 0) in addition to what you have mentioned in issue #359 , then a proper solution is reached with a graph. but if it sticks to 0 (the infinite horizon case), then the message 'The given type violates the Finite Human Wealth Condition; the Finite Human wealth factor value 1.01942' shows before the RuntimeWarnings.

llorracc:
Thanks for closing this!

Issue #404, opened 2019-10-16
duplicate log normal draw methods in simulation.py


sbenthall:
In `simulation.py`, there appears to be two methods with the functionality of drawing N lognormal values.

They have slightly different inputs (one takes a `mu` argument, the other doesn't) but identical descriptions in the documentation.

`drawMeanOneLognormal`
https://github.com/econ-ark/HARK/blob/master/HARK/simulation.py#L9

`drawLognormal`
https://github.com/econ-ark/HARK/blob/master/HARK/simulation.py#L46

It looks like this was code duplication 3 years ago, maybe because of a changing model need in a downstream use.

My guess is that it's possible to combine these two functions by giving a sensible default `mu` value. This would clean up the code.

sbenthall:
That said: It looks like most of the functions in `simulation.py` are sampling from various known probability distributions.

This functionality is already provided by `numpy`, which this package imports.
So this code could be replaced by much simpler calls to lower level library functions, e.g.:

https://docs.scipy.org/doc/numpy-1.15.0/reference/generated/numpy.random.uniform.html
https://docs.scipy.org/doc/numpy-1.14.0/reference/generated/numpy.random.weibull.html

mnwhite:
One of those is explicitly for *mean one* lognormal distributions; the
other is more general.  The first calls the second, as a convenience
function.

Yes, essentially all of the functions here are just wrappers (as
acknowledged at the top, I think).

sbenthall:
I see what you mean about drawing with mean one, by setting `mu` relative to `sigma`:
https://github.com/econ-ark/HARK/blob/master/HARK/simulation.py#L37

In the version I'm looking at, I don't see `drawLognormal` called in the first function:
https://github.com/econ-ark/HARK/blob/master/HARK/simulation.py#L34-L44

I see that the wrappers:
 - allow random seed as input
 - allow multiple distribution parameters to be used in one function call.

It may be possible to write the wrapper generically, so the same wrapper code is used for all functions.

This is all quite minor. But I believe I could refactor much of this code to make it tighter.

mnwhite:
Sorry, I was confusing the discretizations functions in HARK.utilities with
these.  Yeah, basically everything in HARK.simulation is very, very weak.
This file and HARK.estimation are mostly placeholders.

sbenthall:
Ok, I'll work on this one as a first contribution.

I don't seem to have the ability to assign myself the ticket.

PR #408, opened 2019-10-20
Issue 404 - refactoring simulation.py

sbenthall:
This PR is for refactoring of simulation.py as per #404

Currently it:
 - adds a unit test for each method in simulation.py
 - refactors drawMeanOneLognormal to use drawLognormal

I would like to go farther and refactor more of the distribution sampling code to be more concise.
But I'm making this PR mainly to test project workflow.


HARK/simulation.py
```diff
@@ -33,19 +33,13 @@ def drawMeanOneLognormal(N, sigma=1.0, seed=0):
     # Set up the RNG
     RNG = np.random.RandomState(seed)
```
MridulS: This isn't needed in this function as the seed is passed on to `drawLognormal`.

HARK/tests/test_simulation.py
```diff
@@ -0,0 +1,45 @@
+import unittest, sys
+
+import HARK.simulation as simulation
+
+class SimulationTests(unittest.TestCase):
+    '''
+    Tests for ....
```
MridulS: sampling distribution with default seed?

PS: looks like `import sys` isn't used

HARK/simulation.py
sbenthall: good call

HARK/tests/test_simulation.py
sbenthall: yep

sbenthall:
I corrected based on @MridulS 's feedback.

Based on today's meeting, my understanding now is that `simulation.py` code is deprecated and slated to be replaced with `dolark` in the near future.

So I won't do any more refactorings in this PR. I submit this PR to be reviewed and pulled.

PR #411, opened 2019-10-22
fixes to requirements.txt to fix sphinx dependency errors. Fixes #406

### LINKED ISSUE #406: Current Sphinx version incompatible with numpydoc
I'm trying to build the documentation using Sphinx.

With sphinx-build properly running with Python 3, I've tried `make html` in the `Documentation` directory as per [the instructions](https://github.com/econ-ark/HARK/blob/master/Documentation/instructions.txt#L168)

I get the following error:

```
~/projects/econ-ark/HARK/Documentation$ make html
sphinx-build -b html -d ../../HARK-docs/doctrees   . ../../HARK-docs/html
Running Sphinx v2.2.0
Inserting /home/sb/projects/econ-ark/HARK

Extension error:
Could not import extension numpydoc (exception: cannot import name 'Directive')
Makefile:63: recipe for target 'html' failed
make: *** [html] Error 2
```

Looking up this error, it looks like it's the result of incompatibility between the current version of Sphinx (v2.2.0) and the `numpydoc` extension:
https://github.com/readthedocs/readthedocs.org/issues/4057

The recommended fix from that issue, which is quite old (from May, 2018) is the fix the Sphinx version in `requirements.txt` to an early version, `sphinx==1.6.7`

sbenthall:
Addresses #406 

- fixes `sphinx` version to one compatible with extension `numpydoc`
- installs `recommonmark` dependency

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 numpy
 matplotlib
-Sphinx
+sphinx==1.6.7
 numpydoc
 future
 joblib
@@ -12,3 +12,4 @@ funcsigs
 jupyter
 sphinx-rtd-theme
 nbsphinx
+recommonmark


sbenthall:
This PR makes changes to the sphinx configuration and other files to get the documentation to autogenerate without error. See also issue #412 

Because of the `recommonmark` issue, I've folded in the work on #412 into this PR, originally for #406.

New changes:
 - removing references to files that are not modules with APIs form the `autosummary` list in the Sphinx configuration
 - removing the `recommonmark` extension

Issue #420, opened 2019-10-29
Markdown errors in HARKmanual causing Sphinx errors


sbenthall:
In issue #412, it's noted that with `recommonmark`, the `make html` Sphinx command errored.

https://github.com/econ-ark/HARK/issues/412#issuecomment-545105238

PR #411 "solved" this problem by removing `recommonmark`. Unfortunately, this appears to prevent Sphinx from parsing any markdown at all. The baby was thrown out with the bathwater.

I've isolated the broken markdown to these files:
https://github.com/econ-ark/HARK/blob/master/Documentation/HARKmanual/HARKmanual.md
https://github.com/econ-ark/HARK/blob/master/Documentation/HARKmanual/ContributingToHARK/ContributingToHARK.md

The former appears to be autogenerated from the original LaTeX version of the manual and contains many irregular looking image references, as they are are used for rendered mathematical formulae.

These documents have been generated for uses that are _not_ the sphinx-generated ReadTheDocs:
https://github.com/econ-ark/HARK/issues/409#issuecomment-546629439

Therefore, the errors they are caused are due to the confounded directory structure.
I'll remove these files as part of the documentation cleanup effort. See #409

MridulS:
I was able to setup a readthedocs copy running at https://hark-mriduls.readthedocs.io <https://hark-mriduls.readthedocs.io/> with recommonmark and the HARKManual.
But I get the same recommonmark  error locally, this could be also about our local installations of the dependences.

Issue #417, opened 2019-10-29
Examples should be removed


llorracc:
@sbenthall, I just came across the "Examples" directory which (as you point out) has a copy of the "Gentle Intro" in it, and some tests.  

My recollection is that this was a first "test case" of how to go about including notebooks in testing.  

As part of your general reorganization of the intro/testing infrastructure, you should remove this.  

(I almost just did it myself, but decided that this is now your domain and I should turf it to you).

PR #421, opened 2019-10-29
ReadTheDocs fixes

sbenthall:
This PR:
 - removes the generated HARKmanual related markdown, which was ill-formed and breaking the `recommonmark` parser, and brings back the parser. See #420 
 - points to the Gentle notebook in DemARK (#409), rather than the one in Examples/ (which is now deleted #417 #419 ) 
 - adds some introductory text to the RTD index, explaining purpose of docs and linking students to DemARK.
 - refactoring the index.rst logic so that it can include the autosummarized API docs without an awkard 'placeholder' element with a dead link.


shaunagm:
I'm still seeing the placeholder element with the dead link:

Screenshot from 2019-11-06 10-48-56

PR #425, opened 2019-10-31
Refactor 'readme' in docs into separate files to better use sidebar nav

sbenthall:
#409


sbenthall:
- fixed README link so that it works with the first change
- adding (missing) link to CONTRIBUTING file into ReadTheDocs. Technically for #410

PR #432, opened 2019-11-07
removing Model directories from the API documentation

sbenthall:
The Python API readthedocs have too many entries for the specific models implemented in HARK (such as BayerLuetticke).

This PR removes all of those entries from the API documentation.

In the project meeting, we discussed remove everything below two layers in the module hierarchy (i.e., only entries of the form `HARK.*`). But since `HARK.BayerLuetticke` is a directory, not a module, the corresponding "API Doc" for this entry was empty and so is nonsensical.

The current implementation cleans up the ReadTheDocs to make them more usable.
It punts on the question of how to document the Models.

### DIFF
--- a/Documentation/api.rst
+++ b/Documentation/api.rst
@@ -4,41 +4,7 @@ API Documentation
 .. autosummary::
    :toctree: generated
 
-   HARK.BayerLuetticke.Assets.One.FluctuationsOneAssetIOUs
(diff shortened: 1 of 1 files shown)


sbenthall:
See: https://github.com/econ-ark/HARK/issues/409#issuecomment-551115644

PR #433, opened 2019-11-09
Pass the RNG stored in AgentType to portfolio agents' methods for dra…

### LINKED ISSUE #430: Risky return rate in ConsPortfolioModel is always the same.
The current iteration of the portfolio problem implementation always draws the same return factor for the risky asset. This is a problem for simulations, but does not seem to create issues for the solution process (the discretization is properly created).

The simplest way to reproduce this issue is:
- In [ConsPortfolioModelDoc.py](https://github.com/econ-ark/DemARK/blob/master/notebooks/ConsPortfolioModelDoc.py), after line 64 (once RiskyDrawFunc is defined) do a print([RiskyDrawFunc(), RiskyDrawFunc()])

This seems to stem from:
- In line 549 of [ConsPortfolioModel.py](https://github.com/econ-ark/HARK/blob/master/HARK/ConsumptionSaving/ConsPortfolioModel.py) self.getRisky() calls self.drawRiskyFunc(). In the default example this is created using LogNormalRiskyDstnDraw() in line 266.
- LogNormalRiskyDstnDraw() creates a function that calls drawLognormal() from [simulation.py](https://github.com/econ-ark/HARK/blob/master/HARK/simulation.py), which by default resets the RNG seed to 0.

Therefore ever time self.getRisky() gets called (since a single draw is requested) the same number is returned.

A possible workaround would be to draw the whole path of risky return factors at the start of the simulation instead of drawing them every period.

pkofod:
…wing risky returns rates.

Fixes #430 

We should of course verify that a) this works with the different examples I provided in the original PR and make it clear that b) you now have to define your function to accept 1 positional argument that represents a RNG state.

@llorracc @mnwhite


mnwhite:
I will prioritize looking at this.  Ok, I just did look at this.  You
should have it grab a seed from self.RNG, as with the other shock
generators in AgentType.  I think the method is to do something like
int(np.floor(2**31*self.RNG.rand())) as the seed.  It's clunky, but it
makes sure that the same seed is passed to all of the RNGs on each run of
the code, as long as self.reset() is called (that's why it appears in
AgentType.initializeSim()).

Man, this sure would be a lot easier on Patrick if I had written a
documentation notebook about our simulation structure.  I mean, what's up
with that??

pkofod:
Oh, of course. So RNG is the actual generator, and the idea is to use that to generate a seed for each simulation. So if the RNG is reset, you get the same sequence next time around. Can you have a look again? Thanks!

llorracc:
Is this ready for merging now?

mnwhite:
Yes.  I haven't run this branch, but if it works for Patrick after making these changes, it's ready for merge.  His RNG is now formatted the same as our other shock-drawers.

shaunagm:
Can someone write a release note for this?

llorracc:
Release note: Fixed bug in ConsPortfolioModel in which the same risky rate of return would be drawn over and over

PR #438, opened 2019-11-26
Tiny fix to getPercentiles

mnwhite:
Added two lines to utilities.getPercentiles to correctly handle case when data has size 0 or 1: return NaN vector of same size as requested percentiles of data.  That is, there is no such thing as "percentiles" of an empty dataset or a singleton.  This can come up when simulated or empirical moments are calculated, but no observations qualify.


mnwhite:
This is a very tiny PR that can be merged immediately.  Adds two lines that prevent a certain error from arising.

PR #443, opened 2019-12-02
Add which_vars to minimizeNelderMead

mnwhite:
Added option to HARK.estimate.minimizeNelderMead to select which parameters are estimated by passing a vector of booleans of the same size as parameter_guess.  When not passed, it defaults to estimating all of the parameters (previous behavior).  Not a breaking change.

This code has been sitting on my computer for many months.  I only realized it wasn't in public HARK when I switched one of my projects from a local, private version of HARK to public package HARK.


llorracc:
This has been reviewed and is ready to merge.  Anybody seeing this who has permission should do it.

Issue #199, opened 2018-10-15
Using HARK in graphical environments or working around it


llorracc:
Just tried to install on an AWS EC2 instance, and on trying to use it got the error "HARK must be used in a graphical environment."  

In fact, there are a number of ways in which someone might want to use Econ-ARK/HARK tools without a graphical front end.  Most notably, someone might want to offload some compute-intensive stuff to AWS or another platform, let it run and generate results, and then collect those results.  This could even include figures; matplotlib can be used without a graphical environment (the figures do not render but can be saved).  

A test case would be to run any of the `do_min.py` files in the [REMARK](https://github.com/econ-ark/REMARK) repo.  These should be executable from within the appropriate directory just by the command-line command `ipython do_min.py` -- which has been tested and found to work on full-blown systems.  (A good choice might be [BufferStockTheory/do_min.py](https://github.com/econ-ark/REMARK/blob/master/REMARKs/BufferStockTheory/BufferStockTheory.py)).

However, it would also be useful to know how someone who DOES want to use interactive graphical tools in HARK could set up a VM that would satisfy whatever are the 'graphical environment' requirements.  (I found that installing the full-blown Ubuntu/Xubuntu desktop was sufficient, but unclear if necessary -- and it took a while to figure out how to do that).

MridulS:
I can look into this.

llorracc:
Great -- the goal is basically to make it easy for people not very familiar with AWS or Google Compute Platform to spin up a VM that can be used with all the graphical tools of HARK.  A script or how-to is what is needed.

sbenthall:
Can this issue be closed?

MridulS:
Yeah, all good now.

PR #444, opened 2019-12-05
Fix HARK GUI issue 

MridulS:
https://github.com/econ-ark/HARK/issues/199

User will no longer get the import Matplotlib error as I have moved it to inside the function instead of a global level import.

For example this will work without any issue

``` python 
In [21]: import HARK 

In [22]: from HARK.ConsumptionSaving.ConsIndShockModel import PerfForesightConsumerType

In [23]: PF_params = {
    ...:     'CRRA' : 2.5,           # Relative risk aversion
    ...:     'DiscFac' : 0.96,       # Discount factor
    ...:     'Rfree' : 1.03,         # Risk free interest factor
    ...:     'LivPrb' : [0.98],      # Survival probability
    ...:     'PermGroFac' : [1.01],  # Income growth factor
    ...:     'T_cycle' : 1,
    ...:     'cycles' : 0,
    ...:     'AgentCount' : 10000
    ...: }

In [24]: PFexample = PerfForesightConsumerType(**PF_params) 

In [25]: PFexample.solve()

In [26]: PFexample.solution[0].cFunc
Out[26]: <HARK.interpolation.LinearInterp at 0x7f102555aa50>

In [27]: PFexample.solution[0].hNrm
Out[27]: 50.49994992551661

In [28]: PFexample.unpackcFunc()

In [29]: PFexample.timeFwd()

In [30]: mMin = PFexample.solution[0].mNrmMin
```

This fix will give an error if the user tries to plot something on the server without having Matplotlib with the proper backend installed. 

the plotting wouldn't
``` python
In [31]: plotFuncs(PFexample.cFunc[0],mMin,mMin+10)
    ..........
    ..........
ImportError: No module named _tkinter, please install the python-tk package
```

The plotting one can be fixed by changing the backend to 'agg' (this has to be done before anything Matplotlib related is imported)

``` python
In [1]: import matplotlib

In [2]: matplotlib.use('agg')

In [..]: plotFuncs(PFexample.cFunc[0],mMin,mMin+10)

```

`plotFuncs` wouldn't throw any error now and it wouldn't do anything as it is implemented to just do `plt.show()` (which doesn't work for a server without any display) . Should we add the capability to save the figure too with plotting functions by passing an argument like `plotFuncs(save_fig=True, show_fig=False)`?


MridulS:
This is good to go, will update other instances of Matplotlib in ConsumptionSaving refactor.

PR #452, opened 2019-12-10
solution_distance and completed_cyles as agent attributes of agent - …

llorracc:
…useful for debugging


shaunagm:
@llorracc can you write a release note?

shaunagm:
@llorracc please don't change the release note labels, as a rule of thumb - it only ever creates more work for me.  in this case, I was planning to use the "needs release notes" tag to identify which PRs still needed to be added to the release notes changelog, but now I have to find them manually.  feel free to change other labels, but not release labels.

Issue #440, opened 2019-11-26
Split out model execution into examples/ from HARK package code


sbenthall:
Related to #415 

HARK is a Python package that is imported by other packages. It is in most respects structured like [a typical Python package](https://docs.python-guide.org/writing/structure/#sample-repository).

One way it departs from the normal way Python packages are structured is how it uses subpackages. Its subpackages, like `cstwMPC`, `BayerLuettecke`, and `FashionVictim`, `SolvingMicroDSOPs`, are currently crafted as reproductions of particular research results. Several have directories for `Figures` and `do_all.py` scripts as if they are REMARKs. Most have long `__main__` methods that do display of results (in print lines or plots) in the same file as the model is built and configured.

This is not standard software architecture. There is a missing distinction here between:
 - a **library**, which is robustly tested reproducible code for wide use, and
 - an **application**, a particular piece of functionality not intended for reuse.
 - a **demonstrations** of the software functionality designed to guide new users.

I gather that Econ-Ark has some idiosyncratic needs. But from what I understand so far, these categories fit Econ-Ark's use cases. Roughly speaking, research publication uses of HARK are like applications. Pedagogical uses of HARK are like demonstrations. The core library code, HARK, is neither of these.

If this tracks, then it would make sense to following the conventions of other scientific software packages in separating core 'library' package code and other code that demonstrates it. A good example of this is `matplotlib`. That project has [a top level ](https://github.com/matplotlib/matplotlib/tree/master/examples)`examples` directory which contains its tutorials. This is adjacent to the directory that contains its library code. If implemented and seen this pattern in other scientific software repositories I've worked with. (e.g. [networkx](https://github.com/networkx/networkx)). [Dolo](https://github.com/EconForge/dolo) does this as well.

There are many benefits to this architecture. Perhaps the greatest is that it makes software developers make decisions about what functionality is supported as reusable code. Code that's in a library can be held to a rigorous testing standard and held to be independent of configuration parameters set in the examples.

sbenthall:
Application examples that need to be fixed up:
- [ ] BayerLuetticke
- [ ] cAndCwithStickyE -- to be removed. See #335 
- [ ] cstwMPC
- [ ] FashionVictim
- [ ] SolvingMicroDSOPs

Based on [this note](https://github.com/econ-ark/HARK/pull/335#issuecomment-513493936), I will check for each example to make sure there isn't a dependency on them.

[...]

Issue #370, opened 2019-07-23
Break scripts to code and tests


MridulS:
A lot of files in HARK are currently python scripts (with a main() function to test the classes and functions in the file) rather than a "module".
For example if you look at the documentation of https://hark.readthedocs.io/en/latest/generated/HARK.interpolation.html#module-HARK.interpolation, https://hark.readthedocs.io/en/latest/generated/HARK.estimation.html,  you would see that dangling `main()` function.

llorracc:
So, I guess you think they should all be modules instead of scripts?  Sounds right to me.  I think their scriptyness dates from early days when we were trying to build in examples and testing in the files themselves rather than having standalone tests and examples.  

In cases (if any) where the main() actually includes some content, we should make sure to port that content to an appropriate place (a DemARK or a test, probably) before deleting it.

JackShiqiLi:
I had a look at this, it seems that in almost all files in HARK, the main() only contains a warning like this,
![Core](https://user-images.githubusercontent.com/41880047/65500712-0d25fc80-df03-11e9-857d-fd61d76c7e46.jpg)
except for parallel.py that has a **test** in there and interpolation.py that I think is designed to have some **examples** for each type of interpolation, but has not finished yet.
![Interpolation](https://user-images.githubusercontent.com/41880047/65500895-5aa26980-df03-11e9-960f-b59f645c35cb.jpg)
If @MridulS is concerned about the main() under functions in the readthedocs webpage, I think all functions, class, etc are recognized by Sphnix and generated in the webpage automatically, to avoid that, I suggest that we can move the warnings directly into the `if __name__ == "__main__":` block (instead of having a function),  and move the test into a new file if we still need it for further development of parallel.py. For the examples in interpolation.py, if they are there on purpose (as suggested by the warning message), I think we should also move it into the `if __name__ == "__main__":` block instead of a new file.

MridulS:
@JackShiqiLi my main intent is  to remove *all* the main() as well if `if __name__ == "__main__"` block of code which should never be inside a library(individual files which implement models/helper code), otherwise it's a collection of scripts.
If you look at various models, majority of them has `if __name__ == "__main__"` block which tests the model by initiating it, playing with parameters and plotting. A grep gives me 
```
./HARK/simulation.py:if __name__ == '__main__':
./HARK/interpolation.py:if __name__ == '__main__':
./HARK/FashionVictim/FashionVictimParams.py:if __name__ == '__main__':
[...]

PR #455, opened 2019-12-12
Consumption Saving Refactor - move examples to notebooks - tests boilerplate

MridulS:
I have removed the def main() code from ConsIndShockModel.py file and moved it to a example file.
This file will be present in the examples/ directory and parts of that file will be in the tests/ directory to create base level testing coverage of the module.

We also need to discuss what is exposed to the final user, for example `ValueFunc`, `MargValueFunc` and `MargMargValueFunc` are use by other classes to solve the model but I don't see users using this directly. Should this be exposed in the final documentation on readthedocs?

In this PR I'll repeat the same with all the other models and files in the ConsumptionSaving folder.

In this refactor I am also removing py2 specific dependencies in the ConsumptionSaving module as the py2 drop date is 15th Jan according to the roadmap.


MridulS:
related: https://github.com/econ-ark/HARK/issues/370
https://github.com/econ-ark/HARK/issues/440

MridulS:
[Build fails because it is running for py2.7 too]

MridulS:
Move def main to example_model.py

- [x] ConsAggShockModel.py
- [x] ConsPortfolioModel.py 
- [x] ConsGenIncProcessModel.py
- [x] ConsPrefShockModel.py
- [x] ConsIndShockModel.py
- [x] ConsRepAgentModel.py
- [x] ConsMarkovModel.py
- [x] ConsMedModel.py
- [x] TractableBufferStockModel.py

sbenthall:
Would you like me to review this some time?

MridulS:
@sbenthall that would be great! Almost done with this one.

MridulS:
@sbenthall Can you go through this once, I don't want to make more changes to this as it will create a big PR. I'll start a new PR with further changes.

sbenthall:
@MridulS Sorry for the slow action on this one.
I merged updates from master.
It looks like there are some tests failing.

MridulS:
Ah the consumption parameters change is failing the tests, I'll update this PR.

MridulS:
okay this is weirder than I thought, something is breaking between import structure of python3.6 and python3.7

This works for py3.7 and py3.8

sbenthall:
Because the 3.6 build was broken on an issue involving import of the `ConsumptionSaving` files, and because there was a merge conflict with `master` over `HARK/__init__.py`, I took the liberty of trying to resolve the merge.

I removed this line:
`from HARK.ConsumptionSaving import *`

which seemed to be opening up a new way to import ConsumptionSaving files.
If I understand correctly, that line lets you do:
`from HARK import MarkovConsumerType`

instead of
`from HARK.ConsumptionSaving import MarkovConsumerType`

I see now that you used the former style (`from HARK import MarkovConsumerType`) elsewhere in the PR.
So I have to apologize for breaking something new.

A change I would test myself if I could use the different build environments in GitHub actions myself is whether if all these references are changed back to `from HARK.ConsumptionSaving import XXXXXX`, does that then pass the 3.6 build test.

MridulS:
reverted changes to make this py2.7 compatible and make it v0.10.4 friendly

PR #464, opened 2019-12-28
clock is removed in py3.8

MridulS:


MridulS:
travis fails for py2.7 as perf_counter is py3 only.

mnwhite:
We can use time.time() or process_time(), right?

We can also drop the py27 Travis tests next week. But we really ought to
announce this.

MridulS:
True. Dropping py2.7 is on the timeline wiki for 15/01/2020 but we should
make a more public announcement.

MridulS:
https://github.com/econ-ark/HARK/pull/471

PR #465, opened 2020-01-01
Linting parallel.py

MridulS:
Supersedes https://github.com/econ-ark/HARK/pull/351


MridulS:
Linting through black, merging in.

PR #466, opened 2020-01-03
Loading default parameters into class during initialization

sbenthall:
This PR is for progress on #446 -- parameter management.

One attractive possibility is that each Model class has default parameters that can be overridden during initialization.

Commit https://github.com/econ-ark/HARK/commit/5c6afef324025141a8bad0023e658ffc4bc663bf models a way to do this with minimal change to the library for `PerfForesightConsumerType`:
 - The same dictionary of parameters from is used as before (this dictionary can be moved later)
 - It is accessed and copied during `PerfForesightConsumerType` initialization
 - It is `update()`ed with the keyword arguments to the initialize
 - This is then passed through to the superclass constructor as initialization proceeds.

In other work on this issue, I'll set this up for the other Model/Type classes. I wanted to put this early PR version out there because I wonder if others think this is a good change.


sbenthall:
This pull request is now ready for review. Requested from @MridulS

HARK/ConsumptionSaving/ConsGenIncProcessModel.py
```diff
@@ -965,7 +966,7 @@ class GenIncProcessConsumerType(IndShockConsumerType):
     solution_terminal_ = ConsumerSolution(cFunc=cFunc_terminal_, mNrmMin=0.0, hNrm=0.0, MPCmin=1.0, MPCmax=1.0)
     poststate_vars_ = ['aLvlNow', 'pLvlNow']
 
-    def __init__(self, cycles=1, time_flow=True, **kwds):
+    def __init__(self, cycles=0, time_flow=True, **kwds):
```
MridulS: is there a reason for this switch?

HARK/ConsumptionSaving/ConsMedModel.py
```diff
@@ -509,7 +510,7 @@ class MedShockConsumerType(PersistentShockConsumerType):
     '''
     shock_vars_ = PersistentShockConsumerType.shock_vars_ + ['MedShkNow']
 
-    def __init__(self,cycles=1,time_flow=True,**kwds):
+    def __init__(self,cycles=0,time_flow=True,**kwds):
```
MridulS: same thing

HARK/ConsumptionSaving/ConsumerParameters.py
```diff
@@ -250,13 +249,11 @@
 
 pLvlPctiles = np.concatenate(([0.001, 0.005, 0.01, 0.03], np.linspace(0.05, 0.95, num=19),[0.97, 0.99, 0.995, 0.999]))
 PrstIncCorr = 0.98       # Serial correlation coefficient for permanent income
-cycles = 0
 
 # Make a dictionary for the "explicit permanent income" idiosyncratic shocks model
 init_explicit_perm_inc = copy(init_idiosyncratic_shocks)
 init_explicit_perm_inc['pLvlPctiles'] = pLvlPctiles
 init_explicit_perm_inc['PermGroFac'] = [1.0] # long run permanent income growth doesn't work yet
-init_explicit_perm_inc['cycles'] = cycles
```
MridulS: probably I am missing something here but why these changes?

llorracc:
Can you trace back to the PR where the changes were made?  

Looks to me as though the default is being changed from finite to infinite horizon models.  It should have been infinite horizon all along, so if that is what is being done it make sense -- puzzle is why it wasn't infinite horizon all along.

sbenthall:
@MridulS Those changes were made because:
 - the default parameters in `ConsumerParameters.py` set `cycles=0`
 - The classes had `__init__(cycles=1)`, but this was overwritten by the loading of the parameter dictionary at runtime
 - When the superclass `__init__` is being called, it is both passing through `cycles=cylces` as a named argument and also passing `**kwds` through
 - Without these changes, the Python interpreter complains that the superclass `__init__` is getting 2 different `cycles` arguments on the same call.

As @llorracc says, the infinite horizon was being passed through to override the finite horizon anyway. So I made the call that this is set properly in the `__init__` and not at all in the parameter dictionary.

@llorracc : I think you may be responded to comments on PR's in email and the lack of context is confusing. The subject line of the email notification names the PR where this discussion is happening. It should say "#466"

sbenthall:
Ok to merge #466?

PR #471, opened 2020-01-16
time.clock is removed from py3.8

MridulS:
Instances of time.clock in consumptionsaving is removed in https://github.com/econ-ark/HARK/pull/455

### DIFF
--- a/HARK/core.py
+++ b/HARK/core.py
@@ -18,7 +18,7 @@ from distutils.dir_util import copy_tree
 from .utilities import getArgNames, NullFunc
 from copy import copy, deepcopy
 import numpy as np
-from time import clock
+from time import process_time
 from .parallel import multiThreadCommands, multiThreadCommandsFake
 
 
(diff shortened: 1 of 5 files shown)


MridulS:
the only option to make it backward compatible with py2.7 is to use time.time() but it is not the recommended way going forward as there is no guarantee that the precision will be better than 1 second, from the official docs

> Note that even though the time is always returned as a floating point number, not all systems
> provide time with a better precision than 1 second. While this function normally returns non-
> decreasing values, it can return a lower value than a previous call if the system clock has been set back between the two calls.

MridulS:
Do we make this backward incompatible change? @mnwhite @llorracc

llorracc:
@MridulS, what do you think we should do about this?

I was going to say "could you do a global search on HARK, REMARK, and DemARK repos and replace any instances of the use of time.clock() with whatever" but I'm not sure there's a "whatever" that is future-proof.  Is there, for py3, an equivalent to "from future import"?

MridulS:
for time.clock() there is time.time() which works for both py2 and py3. I have replaced it in HARK, will do it for REMARKs and DemARKs too.

adsk2050:
I think these changes are not yet published to the conda-forge.

MridulS:
@adsk2050 what version of HARK are you using from Conda-forge?

adsk2050:
I installed it by using the command `conda install -c conda-forge econ-ark`. The version is 0.10.4, but this is the same version as what is showing when I installed from source.

PR #477, opened 2020-01-23
add 3.8 and trigger the action on pull_request

MridulS:


MridulS:
GitHub Actions will test PRs against the matrix of [Windows, MacOS, Ubuntu] on py3.6, py3.7, py3.8.

Issue #484, opened 2020-02-03
HARK Docs version has been stuck at 0.97 for years


llorracc:
Now that we have Sphinx working we should update that.

sbenthall:
See #474

PR #489, opened 2020-02-04
changing version and author in sphinx

sbenthall:
Addresses #474 and #484


Documentation/conf.py
```diff
@@ -69,17 +69,17 @@
 
 # General information about the project.
 project = u'HARK'
-copyright = u'2016, TEMP'
-author = u'TEMP'
+copyright = u'2016, Econ-ARK team'
```
MridulS: can you make this 2020?

sbenthall: ok i made this change

MridulS:
@sbenthall if you have access to readthedocs can you go in and force start the build? Not sure if changes are visible there.

PR #498, opened 2020-02-09
simplify checks for time varying solveOnePeriod, fixes #496

### LINKED ISSUE #496: simplify code around time-varying solveOnePeriod methods
Related to #492. There is some complexity introduced to handle the case where a solution algorithm depends on a `solveOnePeriod` function that varies over time. This involves putting it in the `time_vary` list of strings.

https://github.com/econ-ark/HARK/blob/master/HARK/core.py#L174-L176
https://github.com/econ-ark/HARK/blob/master/HARK/core.py#L896-L898

I have not been able to find a downstream case where this feature is used.

It is also a needlessly cumbersome way to signal that there are varying solution algorithms: since the code expects that `agent.solveOnePeriod` is indexical by `t`, a type check would suffice here. A boolean argument would also do, etc.

Other uses of the `time_vary` list seem to indicate things like state variables within the underlying MDP model. It is a rather confusing extension of scope to have this list contain both model variables as well as referenes to solution functions. Ideally, a well-defined model could be solved using more than one algorith.

I would like to simplify this aspect of the code.

Is this functionality needed? I.e. can it be removed entirely?
Otherwise, I can simplify it while leaving the functionality intact.

sbenthall:
This PR resolves #496 by simplifying the code around multiple solveOnePeriod methods:

- The check is performed in only one place (less overhead on user reading and understanding the code)
- It does not require special use of the time_vary list (less overhead on user building on the code)


sbenthall:
Regarding automated tests for the functionality touched by this PR:

There are two cases:
- The first case, where `solveOnePeriod` is a function, is tested by this test: https://github.com/econ-ark/HARK/blob/master/HARK/tests/test_core.py#L100
- The second case, where `solveOnePeriod` is subscriptable:
   - has never been used
   - there appears to be no example in the literature of how it would be used
   - theoretically, if there were a case where it could be used, it would be possible to implement it otherwise by having `solveOnePeriod` be a function that had several cases.

I'm hesitant to write a test for this functionality that simply may not be necessary or desirable in the first place.

PR #509, opened 2020-02-10
New model: Labor supply on intensive margin

mnwhite:
This PR provides a new model for HARK, in which agents make both a consumption-saving decisions and a (continuous) labor-leisure decision.  Agents observe their permanent and transitory productivity shocks before deciding how much labor to supply each period, so TranShk is a second state variable.  Utility function has a Cobb-Douglas-like aggregator for leisure and consumption under a single CRRA term.

Most model code was written by Tiphanie Magne in 2017, with some cleanup by me in 2019 and then a bit more by Jack Shiqi Li.

I could swear this was merged into master months ago, but apparently not.


sbenthall:
What if, for new Models, rather than putting all the test and demo code into the `__main__` method in the model class file, rather:
 - the demo code was put into `examples/` as a notebook as per #455 
 - the test code had automated tests that would be run with Travis: https://github.com/econ-ark/HARK/tree/master/HARK/tests

This has two benefits:
 - the demonstration code would then be available to be ingested into ReadTheDocs
 - the test code would serve as further confirmation of the robustness of the more basic functionality, which would make the library more robust to refactoring, etc.

mnwhite:
Yes, that's a better structure.  I'll move the code.

mnwhite:
Ok, the main block code has been moved to a new examples file, and tested in place.  This revealed a bunch of problems with the code (now fixed), so it was a very good idea.

HARK/ConsumptionSaving/ConsLaborModel.py
```diff
@@ -0,0 +1,567 @@
...
+        Returns
+        -------
+        None
+        '''
+        raise NotImplementedError()
+
+        
+    def getControls(self):
+        '''
+        Calculates consumption for each consumer of this type using the consumption functions.
```
sbenthall: I think this docstring is incorrect. There are more controls than just consumption in this model

MridulS:
for the unit tests in https://github.com/econ-ark/HARK/pull/455 I use ideas from the example and then create incremental unit test. Example: testing one function `checkMarkovInputs` of `MarkovConsumerType` https://github.com/econ-ark/HARK/blob/2b99308daabea1dc4c6109509793ddfbbb269068/HARK/ConsumptionSaving/tests/test_ConsMarkovModel.py

mnwhite:
I fixed that docstring error just now (thanks).  I've never made automated tests before; I'll try to get around to figuring out how to do that and put them in here.

sbenthall:
@mnwhite I'd be happy to do a quick call with you some time about automated tests!
we could even pair code it (remote) if you had the time for it.

sbenthall:
Shall I merge this in? It's better if there are tests, but there is no such general policy requiring tests on PRs at this point, as far as I know.
If this feature is needed downstream, might as well merge it.
My offer to walk you through automated testing stands, any time you've got time.

mnwhite:
Thank you!  We will make tests, I swear.

Issue #427, opened 2019-10-31
Small (easy) improvements in LaborIntMarg model


llorracc:
Two (tiny, not that important) tasks left to do after #392 was merged:

(1) To adjust the xlim and ylim for the infinite horizon figures so that they're "nice" like the lifecycle ones below.

(2) To add a little 6 line method that looks at some attribute of a LaborIntMargType, reading it as nth degree polynomial coefficients, and constructs self.LbrCost using that polynomial.  This is just so that time-varying LbrCost "comes from" somewhere as a constructed attribute, rather than being manually written in a parameter dictionary.

_Originally posted by @mnwhite in https://github.com/econ-ark/HARK/pull/392#issuecomment-548514666_

mnwhite:
In fact, here's the code for that method, from a (private) branch:

`
    def updateLbrCost(self):
        '''
        Construct the age-varying cost of working LbrCost using the attribute LbrCostCoeffs.
        This attribute should be a 1D array of arbitrary length, representing polynomial
        coefficients (over t_cycle) for (log) LbrCost.
        
        Parameters
        ----------
        None
        
        Returns
        -------
        None
        '''
        Coeffs = self.LbrCostCoeffs
        N = len(Coeffs)
        age_vec = np.arange(self.T_cycle)
        LbrCostBase = np.zeros(self.T_cycle)
        for n in range(N):
            LbrCostBase += Coeffs[n]*age_vec**n
        LbrCost = np.exp(LbrCostBase)
        time_orig = self.time_flow
        self.timeFwd()
        self.LbrCost = LbrCost
        self.addToTimeVary('LbrCost')
        if not time_orig:
            self.timeRev()
`

So no PhD in economics required, just the ability to copy-paste that code, remove 'LbrCost' from the class definition of time_vary, and call self.updateLbrCost() in from the update() method.

llorracc:
Thx.  PhD-level econ is required for understanding what you are doing and why, but not for what to do.

rkcah:
I will have a go at this. This is my first attempt at open source!

mnwhite:
I already did #1, I think.

rkcah:
I have successfully implemented the second feature.

I will send a PR then.

MridulS:
fixed in https://github.com/econ-ark/HARK/pull/524

PR #524, opened 2020-02-15
Added updateLbrCost Function

rkcah:
Added the second feature as mentioned in #427 

Included the following three changes:

1. Removed `LbrCost` from `time_vary` class definition
2. Added `self.updateLbrCost()` to the `update()` function
3. Added the function `updateLbrCost`


mnwhite:
Great!  Can you edit ConsumerParameters to have the example dictionaries use this method?  Note that for the infinite horizon example, LbrCostCoeffs will only need one element-- there's only one (infinitely repeated) period, so any function beyond f(t) = a would be very silly!  The short lifecycle example can have two coefficients, just to demonstrate this function.

rkcah:
I have added the required changes in `ConsumerParameters.py`. Please have a look and let me know if something else is required to be done.

mnwhite:
Not quite.  The old LbrCost was [0.36], and your LbrCost-generating method takes the exp of the polynomial function (as it should), so the new LbrCostCoeffs should have log(0.36) ~= -1.0 for its constant term.

In the lifecycle example, my point was that you should choose a constant and linear term and explicitly write them.  Just use [-2.0, 0.4]. It doesn't make sense to arbitrarily use the same coefficient for the constant and linear coefficient.

rkcah:
Changed the values as you suggested. I was a bit confused about which values to give so I just kept the existing values. Thanks for pointing out the error.

mnwhite:
Great!

PR #526, opened 2020-02-15
adding tests for IndShockConsumerType class

### LINKED ISSUE #510: Tests for IndShockConsumerType based on BufferStock
Related to #504 

BufferStock uses a lot of IndShockConsumerType functionality.
https://github.com/llorracc/BufferStockTheory/blob/master/Code/Python/BufferStockTheory.py

Write tests for this functionality, using the parameters from that REMARK case.

sbenthall:
... based on BufferStock REMARK. Fixes #510 

Note limitations due to #525


sbenthall:
Looks like there is a very, very small difference in a computed value when testing on Windows.
I trust this difference is not significant.
I'll replace the `assertEqual` with `assertAlmostEqual`, which rounds to 7 digits.

sbenthall:
requesting review from @MridulS

MridulS:
@sbenthall can you move this tests to the directory `HARK/ConsumptionSaving/tests` instead of `HARK/tests/`. As we start building up the test suite this will severely clutter the main `tests` folder.

sbenthall:
Ok, done.

sbenthall:
I'm pretty confused by why the different 3.6/3.7 etc. path handling keeps breaking the test suite.

Is this related to what was happening in #455 ?

sbenthall:
Adding an __init__.py to the tests directory got the tests passing.
Hows it now, @MridulS ?

MridulS:
All looks good now.

PR #541, opened 2020-02-24
fix checkConditions refactor in ConsIndShockModel

MridulS:


MridulS:
@sbenthall I merged this in to make DemARK travis happy but can you have a look here?

sbenthall:
I see. Thank you

MridulS:
@sbenthall all good here?

llorracc:
I think so -- the revised BST notebook works with the new candidate release.

If we don't find any problems in the next day or two, I think the candidate release should  become a "real" release.

sbenthall:
I worked out one more bug related to this in #546.
In my view it's good to go.

PR #547, opened 2020-02-27
Tests from DemARK

sbenthall:
Tests from DemARKS, see #543 

Currently in PR:
 - ChangeLiqConstr
 - ConsPortfolioModel Documentation
 - DCEGM Upper Envelope
 - HoweWeSolveIndShockConsumerType


sbenthall:
This PR fails tests only on 2.7, which is no longer supported.
We no longer support 2.7, so we should remove the 2.7 tests from Travis.

Otherwise... I request a review of this PR from @MridulS

Issue #558, opened 2020-03-06
move Documentation/notebooks to /examples as appropriate


sbenthall:
The notebooks in Documentation/notebooks should be moved into examples/ as appropriate.
Can do some cleanup in the process.

https://github.com/econ-ark/HARK/tree/master/Documentation/notebooks

llorracc:
(continuing an email thread between @MridulS and @sbenthall and @llorracc, so there will be a record on GitHub):

The discussion on problems compiling from jupyter to sphinx made me realize a complication in the scheme to make DemARKs entirely self-contained so they can be emailed etc:  We rely extensively on the `latex_envs` notebook extension so that we can use standardized LaTeX commands for objects we use all the time (with the considerable advantage that these objects have exactly the same names in the code).  Like `DiscFac` is defined as `\beta'.

But if someone emails the document without the latexdefs.tex file, it won't compile properly.

The only way I can see around this problem is to add a tool to HARK.utilities that gets invoked at the beginning of every notebook and does what latexdefs.tex does.

sbenthall:
In my opinion:
1. Executed jupyter notebooks can be emailed easily, like a document. These shouldn't be executed, they should be 'read-only'
2. If you want to execute a notebook, you'll need to do it in an environment equipped to do so.
3. The functionality where:
    - A emails notebook to B
    - B executes notebook locally with no environment setup or errors

is well beyond the scope of what can be done with a few lines of support code in a library whose purpose is something else.
If this functionality is important, it would be best to escalate it to the Jupyter notebook community.

llorracc:
The functionality where:
A emails notebook to B
B executes notebook locally with no environment setup or errors

That's not what I had in mind.  What I had in mind was that the person has a proper installation of the econ-ark toolkit on their computer.  (Or they are running in an approved Docker container or VM).  

I could be persuaded that the goal of having DemARK notebooks be standalone has costs that are too great.  But so far, aside perhaps from the whole set of LaTeX nightmares, the costs to get from where we are to that goal do not seem large.  And, for LaTeX, maybe we should just insist that DemARK's should use the "least common denominator" LaTeX (defined by "does it work with reproduce.py").

llorracc:
Hmmm, I didn't mean to close this.  Finger must have slipped.

sbenthall:
I think this issue can be closed, as these Documentation/notebooks are now in examples/

Issue #488, opened 2020-02-04
Editing things in the examples/ directory for consistency, articulating policy in documentation


sbenthall:
@llorracc writes here: https://github.com/econ-ark/HARK/pull/472#issuecomment-582010196

> I think you should remove the "Doc" from their names if
they are being explicitly moved into an area that is already designated for
documentation.

> LifeCycleModelExample ... though if it does go into the "examples" area it should lose the "Example"
from its filename.

It would be a good idea to go through the `examples/` drrectory, which will be used for documentation with #459 / #472 , and edit those notebooks for consistency. This would include rationalizing the naming. This process can result in a "Style Guide", which should be part of the "Contribution Documentation". #410

MridulS:
Cleanup the structure of examples directory. Move example notebooks to ConsumptionSavings from individual folders.

sbenthall:
I think it would be better to use Markdown's code format `this like` rather than the LaTeX ttype formatting

sbenthall:
I see that making this change across all the examples is a bit complicated, as this is used widely, sometimes the code variable names are included in larger mathematical expressions, and the LaTeX formatting is used in the long parameter table lists. I see this as connected to issue #889 in that it reflects the ways HARK code has been written to crossover between LaTeX and Python. Maybe this aspect of the docs should be revisited when this is better sorted out conceptually.

sbenthall:
Possibly something more nuanced is needed. `code format` for class and method names, keeping latex for the mathematical objects --- i"ll try to stick with that for 1.0

sbenthall:
just getting the examples and demarks working is taking so much time, I can't do this by 1.0, especially because of notebook overwrites from other PRs.

sbenthall:
@nicksawhney

PR #560, opened 2020-03-09
Examples cleanup

sbenthall:
This moves and cleans up the 'Journeys' notebooks. See #558 
It also addresses redundant 'Docs' and 'Example' in file names, see #488


llorracc:
@sbenthall,

Have spent a little while looking through this and it looks great.  Merging it now, because I'm hoping to move Bayer-Luetticke from "examples" to REMARK soon, and don't want to mix that with your revamp of the examples.

PR #561, opened 2020-03-10
(maybe) fix some sphinx problems for Gentle

llorracc:
Standard definitions of many objects are contained in the latexdefs.tex file, rather than in each notebook.  This removes a stray \newcommand that was accidentally left in the 'Gentle Intro' notebook.


sbenthall:
This is what I see in my local rendering of the docs. It looks like change causes one of the `rho`'s to not render.

image

PR #564, opened 2020-03-11
removing 2.7 from travis

sbenthall:
Removes 2.7 from travis


MridulS:
v0.10.4 is released which is the last version to support py2.7

Issue #542, opened 2020-02-24
Refactor checkConditions() code


sbenthall:
The checkConditions() code is a bit messy.

Here's one reason:
https://github.com/econ-ark/HARK/commit/525cb7e25a1252bc43875eec8acd62d5e0fa0a9b#commitcomment-37457605

There is a lot more that could be done to streamline this code.

sbenthall:
Other checkConditions() related issues to take into account in this refactor:
- #534
- #525 
- #416

sbenthall:
This issue is also related: #283

sbenthall:
Based on a chat with @llorracc last week, my understanding is that there are three classes of conditions:
 - Conditions that must be met for a model to have non-degenerate solutions
 - Conditions that must be met for the model to have a steady state distribution when solved and simulated for an indefinite amount of time
 - Conditions that are worth noting but have no direct implications for the functionality of software

In the current code, there is only one method for testing for the violation of these conditions.

But since one set of tests matters for solving the model, and another set of tests matters for simulating it, it follows that these tests should be handled separately and executed at different times.

llorracc:
> Based on a chat with @llorracc last week, my understanding is that there are three classes of conditions:
> 
> * Conditions that must be met for a model to have non-degenerate solutions
> * Conditions that must be met for the model to have a steady state distribution when solved and simulated for an indefinite amount of time
> * Conditions that are worth noting but have no direct implications for the functionality of software
> 
> In the current code, there is only one method for testing for the violation of these conditions.
> 

Right; the original code, though messy, did give varying responses attuned to these varying conditions.  

> But since one set of tests matters for solving the model, and another set of tests matters for simulating it, it follows that these tests should be handled separately and executed at different times.

I don't agree with that.  People using these kinds of models are almost always going to want to simulate them, and it would be best for them to be forewarned at the time when they are solving the model if there will be problems simulating it.  No point in working hard on calibrating or solving the solution part only to find that it is useless for your purposes when you get to the sim part.

sbenthall:
> did give varying responses attuned to these varying conditions.

Do you mean varying responses _in the print statements_?
If so, I believe these statements are still there in #568 in its current state.

> People using these kinds of models are almost always going to want to simulate them, and it would be best for them to be forewarned at the time when they are solving the model if there will be problems simulating it.

I don't disagree with this.

But is it always a problem if a simulation does not have a steady-state distribution?

I would propose that:
[...]

PR #568, opened 2020-03-13
Refactoring checkConditions()

sbenthall:
This PR addresses issue #542 

First commit:
 - remove 'public_call' argument from checkConditions(); redundant wit verbose


llorracc:
The `public_call` indicator was supposed to be set to `True` whenever checkConditions was invoked by someone directly using it interactively from the command line.  The point was that if you are running it by hand at the command line you should get the full set of `verbose` results.  Does this modification preserve that functionality?

sbenthall:
No, it does not.
I don't think the code before this PR did that either though. `public_call` was False by default.

It is possible that this functionality changed in my earlier refactoring of this in a prior PR.
These kinds of side-effects are difficult to test. See #525 

It remains the case that `public_call` was redundant with `verbose` no?
I'd imagine the right way to do this would be to have the functionality you want work with the `verbose` argument. Why is an additional one necessary?

Separately, @llorracc , I hope you'll take a look at this commit, which is a proposed refactoring of the condition logic:
https://github.com/econ-ark/HARK/pull/568/commits/35808cebc52af53e1355b7afd978cdc18ce48b00

sbenthall:
@llorracc I recommend we continue discussion of the `public_call` option on issue #283, as it is tied up with other design decisions brought up there.

sbenthall:
Likewise with https://github.com/econ-ark/HARK/pull/568/commits/413bef1a11ba08dce22f997e1dc3e9c48d30b73f, @llorracc 

The motivation here is:
 - more code reuse for checking Conditions by functionalizing out the key logic. This makes the code more succinct (you can see this PR reduces the lines of code by 30)
 - using `conditions` as a dictionary namespace for tested conditions, which puts these important variables in a place where (a) they are easy to find (b) they won't be accidentally overwritten
 - disentangles the specific print statements from the computational logic. This is a first step towards wrapping these in a more flexible logging/UI logic.

I know that this code addresses/expresses some of your substantive research contributions, and so I expect you are sensitive to changes here.

What I'm working towards here is trying to figure out the general feature that this is a special case of.
How would "checkConditions" be used in models that a new user might develop?

sbenthall:
The merge conflicts have no been fixed.
This PR now addresses one of four issues raised in #596

sbenthall:
@llorracc you mentioned on the call today an error with this PR concerning GICPF or something.
I haven't been able to find where it's mentioned in an issue.

llorracc:
@sbenthall, on my screen I see my own question/comment immediately above your query about where my question/comment might be.  Maybe you are seeing it differently?  Anyway, the point was that I'd changed the 'GIC' to 'GICPF' because this was the version of the GIC that applied to the Perfect Foresight model; but your edit reverted to the old 'GIC' terminology, and I wanted you to change it to 'GICPF'

sbenthall:
That's very weird, us seeing the issue differently.
In any case, I'll fix this.

sbenthall:
Ok, I have looked over the code in this PR. I think you have mischaracterized what's happening in it.
 - 'GICPF', via `checkGICPF()` is still used in the PerfForeseightConsumerType
 - 'GIC' is used via `checkGICInd()` in `checkGICInd`

HARK/ConsumptionSaving/tests/test_IndShockConsumerType.py
```diff
@@ -149,7 +149,7 @@ def test_GICFails(self):
         self.assertAlmostEqual(c_m[500], 0.7772637042393458)
         self.assertAlmostEqual(c_m[700], 0.8392649061916746)
 
-        self.assertFalse(GICFailExample.GICPF)
+        self.assertFalse(GICFailExample.conditions['GIC'])
```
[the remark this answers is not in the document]
sbenthall: This test is for the IndShockConsumerType.
I changed it to a test for GIC because I believe that's the appropriate condition to test for this class.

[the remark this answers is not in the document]
llorracc: OK, now I see you're right.  My problem was that I assumed that you were only updating the code to reflect the new way of doing things.  But in this case, you were at the same time also fixing a bug.  I'd been assuming that the original code was correctly using GICPF, but you're right that it should have been looking at GIC all along.

llorracc:
I'll look this over and merge if it seems to work.

sbenthall:
The failing builds look to be due to something in GitHub's internals/backend.
GitHub seems to be undergoing some kind of maintenance--GitHub Pages was down this morning.
I'll try rerunning the jobs again tomorrow.

HARK/ConsumptionSaving/ConsIndShockModel.py
```diff
@@ -1872,113 +1872,120 @@ def getPostStates(self):
...
+
+        Parameters
+        ----------
+        name : string
+             Name for the condition.
+
+        test : function(self -> boolean)
+             A function (of self) which tests the condition
+
+        messges : tuple[string]
```
llorracc, line 1891: messages

llorracc: Seb: Check type signature

HARK/ConsumptionSaving/ConsIndShockModel.py
```diff
@@ -1872,113 +1872,120 @@ def getPostStates(self):
...
+            A length four tuple of string to be printed to the 
+            user, in order: True, always; True, verbose; 
+            False, always; False, verbose.
+
+        verbose : boolean
+            (Optional) A boolean to set verbose printing
+        """
+        self.conditions[name] = test(self)
+
+        print(messages[0].format(self)
```
llorracc, line 1901: Give names to messages 0, 1, 2, like 

condTru=0
condFalse=1

so then you could have print(messages[condTru]) which would be more transparent

sbenthall:
See https://github.com/econ-ark/HARK/issues/542#issuecomment-615874441

sbenthall:
@llorracc the most recent commit resolves the issues that came up in your review.

The issue of the different conditions having different meaning is not yet resolved in this PR, but seems to be an extension of existing functionality:
#542 (comment)

If you need this refactor in before your BufferStock publication, this PR can be merged in as is.

That may be wisest, since checking the conditions for a steady-state ergodic distribution seems to naturally connect to the upcoming work on seeding the simulation with the ergodic distribution. It would be best not to block the Bufferstock publication on this new work.

llorracc:
Let's discuss briefly on Thu?

sbenthall:
messages are not rendering properly, showing `{0.url}` etc. for some reason

sbenthall:
Ok, string formatting is back in.

You can see the printed messages from the automated test with the `-s` flag on `pytest` i.e.

`pytest -s tests/test_IndShockConsumerType.py`

Issue #569, opened 2020-03-15
forward/backward through data access, not data storage


sbenthall:
In light of the goal of disentangling solution and simulation #495 ...
by pulling out a separate Policy class #481 ...
which would be documented #482 ...

A core part of the model logic is going to need to change.

Currently, variation over time is represented as Python lists.
Depending on whether the model is being used to 'solve' (backwards) or 'simulate' (forwards), these lists are `reverse()`d.

https://github.com/econ-ark/HARK/blob/master/HARK/core.py#L248

In the current design, the way the data is being stored in memory is being used to determine which functionality of the model is "active".

This design makes it very difficult to make the solution/policy portable between model instances.
It also introduces an unnecessary O(n) reverse operation.

Instead of doing it this way, the representation of the time-varying attributes should stay stable in memory.
Backwards operations should traverse this data backwards, rather than reversing the data then traversing it forwards.

mnwhite:
The good news is that the O(n) reversing operation is done only rarely, and
never actually moves that much stuff in memory.

But I strongly agree with your point here.  This was basically the very
first design disagreement we had on HARK: Chris wanted time-varying lists
to always run backward in time, I wanted them to always run forward in
time.  I put in the timeFlip stuff as a compromise: they would run backward
when solving but forward when simulating (or basically any other time).

The reason I'd like time-varying parameters to run from beginning of time
to end of time is that *that's how we refer to them on paper*, e.g. we
write c_t(m_t) to refer to the t-th period's consumption function.
Somewhere between 99 and 100% of the time that any user will interact with
an AgentType in HARK, they want model results: something from the solution
or simulation output.  The *only* time that anyone ever wants these
variables phrased backward is during solution... and the time loop stuff is
handled internally by HARK!  We are very capable of programming solve and
solveOneCycle to get the correct inputs even if the lists run forward in
time.

sbenthall:
@mnwhite we are in total agreement here.
I'm working on a PR that removes the backwards time representation issue.

@llorracc I'm curious what you were thinking here and if you would reconsider your earlier position in light of what @mnwhite have argued here.

llorracc:
As Matt mentions, he and I discussed this at some length early on.  

[...]

PR #570, opened 2020-03-15
Removing time flipping and time flow state

sbenthall:
Addresses #569 

Refactoring so that models always store time varying data in time-forward format.

'Backwards time' is to be a matter internal to dynamic programming solvers.

It would also be possible to have a `report` method that reports this data in a `time backwards` way if that's important for presentation.

There are a few failing tests in this PR at the moment--honestly fewer than I expected.
They seem mostly due to lifecycle and cyclical models.
I wonder if @mnwhite has any ideas about how time flipping interacts with those cases so that I can go make the necessary corrections.


llorracc:
As long as you are tackling this, I'd propose that you investigate whether there might be some standardized  data structures for representing time-varying data.  For example, for time series econometrics tools, there is often a way to represent data that "understands" that an interval (of, say, a year" can be divided into subintervals (like, say, quarters, months, weeks, days) which should then aggregate back up to the full interval.  It would not be hard to persuade me that using such structures would be overkill from our perspective.  But maybe there are lightweight standards that would work well.  

This is actually something of a substantive issue, because one of the challenges facing the literature at the moment is the choice between the use of continuous time methods and discrete time methods.  Any choices that we can make that would ease the comparison between continuous time and discrete time solutions might prove useful down the road.

sbenthall:
@llorracc There's an easy answer to your question. It's `pandas`

https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html

It's very widely used. I've used it in both finance and communications analytics.

I'm glad you brought this up. Pandas could be a useful backend for a lot of things. It has some nice integration with Jupyter notebooks and is considered part of the same software ecosystem.

I'm trying to make incremental steps so that we can do thorough testing and review before changing core functionality. So I don't think I'll swap in a Pandas backend in _this_ PR or issue. But I'll keep it in mind if it looks viable in future work.

llorracc:
Great, I've used pandas a good bit already, but did not realize that it has special facilities for handling time series data (which is I think what you are saying?)

sbenthall:
Yes. I've linked to its documentation on that topic.

sbenthall:
Ok, I have traced the discrepancies in one of the tests:

The test in `test_ConsIndShockSolverBasic` is failing because there is a difference in `solver.vPfuncNext` when the time reversal logic is removed (as per this PR).

sbenthall:
This backs up to a difference in `LifecycleExample.solution[1].vPfunc`

sbenthall:
Ah, ok, now I see...

ConsIndShockSolver, and I assume most of the other solvers, have been written with the assumption that the time varying parameters will be flipped backwards at the time when solve() is run.

Wow.

mnwhite:
None of the solvers themselves work that way.  It's
HARK.core.AgentType.solve and HARK.core.solveOneCycle that work this way.
That's what needs to change, nothing in the model files.

sbenthall:
Ah, that's a relief. I must have been misreading earlier.

sbenthall:
I'm requesting a review from @mnwhite .

The tricky part was the loop in `solveOneCycle`.
I've added some more tests for this code to make sure it's tight.

llorracc:
@mnwhite: It seems like this might break a bunch of things?

sbenthall:
@llorracc This PR does not break anything currently covered by the automated tests in HARK.

It is possible that some custom code built on top of HARK might break.
This is always a risk when changing HARK, and why it is a good idea to pin downstream code that is not being actively maintained to a stable release version.

sbenthall:
Any objection to my merging this PR? It has been sitting for 12 days.

mnwhite:
Sorry, pandemic has drawn my attention onto a time-sensitive project.  It
will be days before I get a chance to look at this meaningfully.

If current solver tests run (on a model with cycles=1), then the only thing
this could break is other projects that use HARK but specified their
lifecycle parameters from end to start (no one does this).  It would also
break any AgentType subclasses that people wrote privately, if they
remembered to do time_flow checking in their updateXXX() methods.

PR #577, opened 2020-03-17
Improve ConsPortfolioModel

mnwhite:
This PR includes improvements and cleanup to ConsPortfolioModel.py to bring it a little more in line with other model files, and adds the ability to specify an age-varying list of RiskyAvg and RiskyStd.  These represent agent *beliefs* about risky returns, which might differ from the true (time invariant) risky return distribution.

It also includes a new alternate version of this file that (I think) will replace ConsPortfolioModel.py.  It takes bits and pieces from the old version, but the solver is a wholesale rewrite and simplification of the current version.

This PR is not ready to merge, as I need to rectify it with other changes that have been made to master in the meantime.


mnwhite:
Okay, I think this is ready for merge.  The old version still works, and my new version has its own example file.

In a separate PR, after this is merged, we can remove the old files and remove the ALTs from the new ones.

llorracc:
@mnwhite I've posted my tweaks to add the numerical Merton-Samuelson limit to the ConsPortfolioModelALT.py notebook.  

@MridulS, I think this is ready to merge (assuming Travis agrees)

Issue #580, opened 2020-03-17
Update examples to not use ConsumerParameters


sbenthall:
Following up #527 ...

Fix the `examples/` to not use ConsumerParameters

PR #585, opened 2020-03-18
Remove ConsumerParameters from examples/

sbenthall:
Addresses #580


mnwhite:
example_ConsIndShock.py (and its notebook) still has references to Params.

examples/ConsumptionSaving/example_ConsGenIncProcessModel.py
```diff
@@ -7,13 +7,16 @@
     IndShockExplicitPermIncConsumerType,
     IndShockConsumerType,
     PersistentShockConsumerType,
+    init_explicit_perm_inc
```
MridulS: init_persistent_shocks

also needs to be imported here.

MridulS:
example_ConsRepAgentModel.ipynb notebook needs to be updated too.

mnwhite:
All of the notebooks need to be updated.  And that dictionary does not need
to be imported there.  The example file runs correctly.

MridulS:
Oh yeah, you are right. With the class initiations there is no need to import the dictionaries.

@sbenthall  I was running tests locally on examples notebooks.

```
pytest --nbval-lax examples/
```
this will take a *lot* of time so batch them.

mnwhite:
Someone else should play with this branch before we merge it.  But it's ready from my perspective.

llorracc:
@MridulS can you do this?

PR #587, opened 2020-03-18
syntax and param dictionary fixes for ConsPrefShockModel

sbenthall:
There are some breaking errors in ConsPrefShockModel that this PR fixes.


mnwhite:
Will merge as soon as tests finish.

PR #598, opened 2020-03-25
latexdefs.tex added to examples/ConsIndShockModel

llorracc:
so LaTeX compiles correctly.

This is an illustration of why I want DemARKs to be standalone documents.  If there are files (like latexdefs.tex) that need to be moved along with the notebook in order for the notebook to work properly, people are likely not to realize that those files need to move. 

In this case, what is necessary is the latexdefs.tex file, which has some standard definitions of LaTeX objects that we use all the time.  Though I am the culprit in setting this up, I'd propose that we deprecate this way of doing things in favor of finding a way to add the latexdefs in HARK.utilities.


llorracc:
This is related to https://github.com/econ-ark/HARK/issues/599#issue-587872739

sbenthall:
How does a Jupyter notebook load this .tex file?
I'm unfamiliar with the mechanics at work here.

sbenthall:
Oh I see this is discussed in #599 
Should this PR also include what's necessary to install `latex-envs` ?

llorracc:
I don't know how it works.  It's a "notebook extension" whose specs say that when "enabled" on your jupyter setup, it looks for a file named `latexdefs.tex` in the same directory as the notebook.  Works for both jupyter lab and for notebooks.

llorracc:
My impression is that somewhere in our setup instructions we explain to people that they should make sure to install the `latex_envs` notebook extension.  And if we don't, we should.  It is definitely included in the binder postbuild scripts so it always works when people launch the notebooks that way.  Probably you installed it on your machine a long time ago and that's why you've not had any trouble viewing notebooks that assume it is there.

sbenthall:
Ok, it does not appear that we discuss jupyter notebook extensions in the installation documentation.

This is the relevant documentation on the `latex_envs` extension:

https://jupyter-contrib-nbextensions.readthedocs.io/en/latest/install.html

https://jupyter-contrib-nbextensions.readthedocs.io/en/latest/nbextensions/latex_envs/README.html

It looks like:
 - installation of the extension can be done by adding it to the requirements
 - enabling the extension will require an additional command

llorracc:
> It looks like:
> 
> * installation of the extension can be done by adding it to the requirements
> * enabling the extension will require an additional command

OK.  Can you add it to the requirements.txt file and make a PR?  And, by "will require a new command", do you mean that we need to edit the installation instructions?  If so, could you make such an edit and issue a PR?

PR #610, opened 2020-04-05
Use a python class for discrete probability distributions

sbenthall:
This PR addressed issue #519 

This is _only_ a refactor of the current HARK code. It does not yet pull in any new dolo dependencies.

Rather, it refactors the discrete distribution-related code into a new module, `distribution`, containing a class, `DiscreteDistribution`, with attributes `pmf` and `X`.

This should be familiar enough to the current users of HARK.
The principle benefit of this is that it:
 - begins to compartmentalize this functionality
 - makes the code a little easier to understand because one type is used unambiguously for distributions, instead of a generic type (list of arrays).

Tests are passing locally for me.

I'd like to ask to expedite this PR because my understanding is that @llorracc is doing high-priority work on seeding models with ergodic distributions.
This change, though it touches a lot of code, is meant to be a minimal restructure to make it easier to:
 - build a mixture of Guassian representation
 - eventually swap in Dolo, which would allow smoother generation of discrete distributions from continuous ones.


HARK/distribution.py
```diff
@@ -0,0 +1,538 @@
...
+                    X[i] = -0.5*np.exp(mu+(sigma**2)*0.5)*(erf(tempTop) - erf(tempBot))/pmf[i]
+                else:
+                    X[i] = -0.5*np.exp(mu+(sigma**2)*0.5)*(erfc(tempBot) - erfc(tempTop))/pmf[i]
+
+    else:
+        pmf = np.ones(N)/N
+        X   = np.exp(mu)*np.ones(N)
+    return DiscreteDistribution(pmf, X)
+
+@memoize
```
MridulS, line 129: Any ideas why is this here?
(I know this isn't this PR change)

sbenthall: Are you referring to the @memoize decorator?
I assumed it was for performance reasons.

mnwhite:
Yeah. Memoize is there so that in a problem with a large T_cycle but
identical permanent or transitory shock distributions, the function doesn't
get re-computed dozens of hundreds of times.

     -- mnw

sbenthall:
Please merge #616 before this PR
It is expected that the new test there will fail, since ConsRepAgentModel has not yet been included in this refactor

sbenthall:
I believe this is now ready to merge

HARK/ConsumptionSaving/#ConsumerParameters.py#
```diff
@@ -0,0 +1,81 @@
+'''
```
llorracc, line 1: This is a junk file that should be deleted before merge

sbenthall:
junk is removed in latest commit.

The issue you refer to is #227 I think

llorracc:
Yes, #227 is right

PR #613, opened 2020-04-07
read_shocks now reads mortality as well

mnwhite:
For a long time we've had the ability to store a history of shocks in an AgentType instance and then read them on a simulation pass (rather than draw them anew each time).  This used the makeShockHistory() method of AgentType.  However, this didn't fully do what it was supposed to, as *mortality* shocks need to be recorded as well.

This commit makes the history of who_dies be recorded when makeShockHistory is run, and for who_dies to be read from that history when read_shocks is True.  It also breaks up one method of MrkvConsumerType into two; just a copy-paste of code with a new docstring.


mnwhite:
This functionality is required for code for a soon-to-be released research paper to run.  We'll need to merge and do a release.

MridulS:
@mnwhite anything else that needs to be done in HARK before a release?

mnwhite:
Not for this project to work, no.

Issue #611, opened 2020-04-06
Replace continuous distribution 'draw_' methods with dolo IID processes


sbenthall:
HARK currently has several functions in `simulation` for drawing from continuous distributions:
 - `drawMeanOneLognormal`
 - `drawLognormal`
 - `drawNormal`
 - `drawWeibull`
 - `drawUniform`
 - `drawBernoulli`

This is custom code that has been implemented in many other libraries.
In particular, we can move towards integration with Dolo by replacing this code with calls to the 'simulate()` method in the Dolo IIDProcess class, e.g.:
https://github.com/EconForge/dolo/blob/master/dolo/numeric/processes_iid.py#L114

There are some complications here:
 
- It's possible that the HARK functions and Dolo do not have feature parity. A few observations:
  - HARK `draw` functions have:
     - an `exact_match` option -- I frankly don't understand what this does yet
     - the ability to set a `seed` for the random generator
  - Dolo `simulate` has:
     - a `T` argument. What is this?
     - a `stochastic` argument. What is this?
  - It looks like Dolo does not yet have `simulate` implemented for the LogNormal distribution, and has no class for the Weibull distribution

- The automated tests for HARK depend on the current `draw_` code, including the random seed setting. Swapping in Dolo will require recalibration of all the tests. This will make it difficult to guarantee that the new code does not introduce any errors.

- This would introduce a direct dependency between HARK and Dolo. This is the direction it sounds like @llorracc wants to go. However, from an open-source development perspective, Dolo does not currently have any documented contribution guidelines. Does it have [a release cycle or stable releases](https://github.com/EconForge/dolo/releases)? It's not clear how this dependency will effect HARK's development process.

 - The Dolo IIDProcess classes are for the most part not documented. Unless new documentation is written for Dolo, this change will make HARK less easy to understand. Dolo documentation for IID processes is ticketed here: https://github.com/EconForge/dolo/issues/183

[This is related to #519, which is about discrete distributions. Currently in HARK, continuous distributions are handled entirely separately, in `simulation.py`, from discrete distributions, which at the time of this writing are in `utilities.py` (changing with #610). Dolo IIDProcesses support discretization, so this change is instrumental to syncing up these functionalities.]

sbenthall:
While doing a full replacement with dolo IIDProcesses would be major change--maybe one more appropriate for 2.0 Dolark integration, a more moderate change would be to the necessary refactoring to start mirroring Dolo.

[...]

PR #624, opened 2020-04-14
classes to represent continuous distributions for simulation draws

sbenthall:
As an incremental step towards #611 , this PR refactors the `draw_` functions in `simulation.py` so that each one is encapsulated in a class (`Normal`, `Bernoulli`, etc.)

Draws are now accomplished by using the `draw` method on a class initialized with the appropriate parameters.

This brings the HARK programming idioms closer to mathematical invocations of random variables.
It also brings HARK closer in line with best practices for representing distributions in other libraries, such as Dolo.


sbenthall:
The HARK builds are failing here due to some kind of GitHub checkout error?
It doesn't look related to the code.

mnwhite:
Wooooo

Issue #632, opened 2020-04-18
example_ConsMarkovModel failing


sbenthall:
The `example/example_ConsMarkovModel` is failing.

It's known that the new distribution stuff broke the ConsMarkovModel.

Issue #627, opened 2020-04-16
ConsMarkovModel tests for solution and simulation


sbenthall:
The ConsMarkovModel tests do not yet test solution and simulation

PR #633, opened 2020-04-18
fixes for ConsMarkovModel, example, tests

sbenthall:
Introducing DiscreteDistribution broke the ConsMarkovModel

This PR addresses #632 and (currently, partially) #627


mnwhite:
PR pileup strikes again.  We need to remove the reference to timeFwd() in the test.

PR #636, opened 2020-04-21
[WIP] set default PortfolioSolution values to NullFuncs, fixes #635

### LINKED ISSUE #635: default PortfolioSolution values should be NullFunc
PortfolioSolution's initializer defaults all inputs to None then turns each None into NullFunc()
Simpler to set default values to NullFunc

https://github.com/econ-ark/HARK/blob/master/HARK/ConsumptionSaving/ConsPortfolioModel.py#L78-L108

sbenthall:


sbenthall:
Setting to Work in Progress because it would be best to standardize the code throughout HARK than work on it bit by bit.

PR #646, opened 2020-04-23
Fix DiscreteDstn

mnwhite:
DiscreteDstn wasn't working properly with multivariate distributions.  This PR makes some small fixes to it.


HARK/distribution.py
```diff
@@ -405,15 +405,18 @@ def drawDiscrete(self, N,X=None,exact_match=False,seed=0):
...
             X = X
-        
-        # Set up the RNG
-        RNG = np.random.RandomState(seed)
+            J = 1
 
         if exact_match:
+            # Set up the RNG
+            RNG = np.random.RandomState(seed)
+
```
sbenthall, line 419: I'm confused-- why did you move the RNG setting into exact_match?

mnwhite:
I didn't get around to writing this test yet.

sbenthall:
I saw that. I figured it was better to have the fix in than not.
But a test case is always welcome--you can add it to the same branch and file it as a PR whenever you get to it.

PR #662, opened 2020-04-28
move file manipulation methods into helpers.py

sbenthall:
`core.py` has contained some methods for copying modules on the file system and somesuch.

Elsewhere, @llorracc has expressed the idea that methods that should help with the configuration of the Python environment or Jupyter notebooks should go in the HARK library somewhere.

These sorts of functions are of a very different character from the model code and it doesn't make much sense for them to be sitting side by side.

This PR moves this code from `core` to a new module `helpers` where it will be out of the way.

(it also removes the dummy main() method from core.py, which is unnecessary)


mnwhite:
All of our HARK.blah modules (used to?) have dummy main blocks that printed a message indicating they don't do anything, as in its old form all of the files would be in the user's working directory, and Chris wanted to be sure that *something* happened if a user hit run on *any* file.

PR #663, opened 2020-04-28
remove other main() methods from HARK library files

sbenthall:
removes the other apologetic main methods from the HARK library files


mnwhite:
Will merge once I get to a computer; signing in to GitHub w/ 2FA is hard 1
handed on phone.

Issue #664, opened 2020-04-28
Parameter typing (more support for time varying elements)


sbenthall:
In 2014, Python 3.5 got stronger support for static types, in the form of [type hints and other functionality](https://realpython.com/python-type-checking/#pros-and-cons)`. See [PEP 484](https://www.python.org/dev/peps/pep-0484/) for more information.

Previously, Python hadn't had the ability to do static typing, and so there were widely varying conventions around doing it. But since 3.5, the expectations of the Python community have converged and now Python code is encouraged to be in a more statically typed style.

[I'll say: I personally haven't had the opportunity to work in this way yet! But I think it's important to keep up with the times.]

Static typing makes programming _more rigorous_, much more like mathematics. I recommend we adopt it for HARK. Dolo has done this.

A first area where it would be impactful to introduce static typing would be in the model parameters setting. This is a very user-facing aspect of HARK, and also one connected to HARK's compatibility with Dolo. Some related issues: #640 #659 #660 

The biggest type ambiguity in the model parameters right now is the _optional_ use of a list for a time varying parameter. This design choice has implications throughout HARK: often, there will need to be two nearly duplicate code blocks, one for handling the time-varying case of a variable, one for handling the static case. If there is not support for both kinds of input, then a non-varying parameter needs to be entered as a list of identical values--which is suboptimal. For example:

https://github.com/econ-ark/HARK/blob/master/HARK/ConsumptionSaving/ConsIndShockModel.py#L2912-L2913

The time-varying option could be better supported. I think Dolo supports this through the definition of Processes:

https://github.com/EconForge/dolo/blob/7a12eb7117f80d95bd36955328329d6573b0c23c/dolo/numeric/processes.py

So, what we've called a Distribution would, in Dolo, be a UnivariateIIDProcess. But a parameter might also be defined as an AR1 process, which would capture not only that it is time varying, but how it varies in time.

What I'm proposing is that we develop the Distribution/Process classes in HARK so that there are time-varying subclasses of them (or something like that). Then, when the Model needs to simulate or solve them, it can call Process class code to get at the time-appropriate value. This should reduce the amount of code in the solvers and simulators, making it more readable.

mnwhite:
To clarify, all time-varying parameters must be lists.  They are always
lists, not optionally.  In some instances, they are *trivial* lists with
one element, because this instance is infinite horizon with a single
repeated period.  So there shouldn't be any type ambiguity here.

I need to think more about your Process idea and how it would interact with
[...]

Issue #658, opened 2020-04-24
solveOnePeriod is not a method, but sure looks like one


sbenthall:
There is a strange coding pattern in the Model classes. I am tempted to try to turn it into something more standard, but want to check if there is a hidden purpose to it.

Typically, a ConsumerType will have an attribute, `solveOnePeriod`, that is set upon initialization to some stand-alone function in the module. E.g.:

https://github.com/econ-ark/HARK/blob/master/HARK/ConsumptionSaving/ConsIndShockModel.py#L1658

This stand alone method then typically initalizes a Solver object, runs `solve()` on it, then returns the result (a solution). E.g.:

https://github.com/econ-ark/HARK/blob/master/HARK/ConsumptionSaving/ConsIndShockModel.py#L590-L625

These methods are for the most part, except for the choice of Solver class, identical across the library.

This is not ideal because it makes it difficult to read the code: to follow the solution logic, you have to jump out of one class into the global scope, then into another.

It seems to me that either:
 - (A) solveOnePeriod could be a method, which would just look like exactly the function that is currently assigned to it, or
 - (B) solveOnePeriod could be a standardized method, which points to a particular Solver class, which is then parameterized in the model code.

The only motivation I can think of for the current implementation is that it's suggested by this code block that the idea was to make it possible to have time-varying solution functions:

https://github.com/econ-ark/HARK/blob/78d1981233a6cda398d4e3007be03f558a7bc910/HARK/core.py#L807-L810

However, this flexibility does not seem to be used anywhere in the HARK repository (I checked), nor does it have any tests associated with it. If this was still a desirable feature, it would be possible to build 'solveOnePeriod', as a standardized method, to optionally take a list of class names that could be indexed by time periods.

[Separately, and maybe to be addressed in a different issue: I'm skeptical about the performance of this method of firing up a new Solver object, complete with a lot of initialization, at every step of the backwards inductions. It seems to me there may be a lot of redundant code calls. I'll revisit this once I've got a profiler set up, but I wanted to mention it now to start circulating the thought.]

mnwhite:
Yes, I coded it this way to allow for time-varying solveOnePeriod
functions. This was intentional.

Not *all* of our solution functions create a solver class. Some of them are
just coded as a single function, which is a lot easier to read and write.

I don't particularly like our "object oriented solvers". The idea was to
not rewrite code across solvers, but they made the code an absolute maze to
navigate. Moreover, we never actually had a meeting to plan out what the
[...]

Issue #660, opened 2020-04-27
separate out model definition, solution method parameters, and simulation parameters


sbenthall:
The parameters fed to HARK models may be:
 - part of the mathematical model definition
 - parameters for the solution algorithm (e.g. number of discretization points)
 - relevant to the simulation, but not the solution

These are currently all tossed to the AgentType class on initialization with no namespacing.

This makes it hard to, for example, export model definitions to HARK (#659), compare different solutions to the same model but with different solution algorithms (https://github.com/econ-ark/HARK/issues/659#issuecomment-620193576 ), and disentangle the solution and simulation functionality (#495 )

llorracc:
This is a good broad-stroke characterization.  But in some cases the distinctions may be hard to make.  For example, the "true" results for the model might depend on how many periods are simulated, or the process by which the aggregate shocks are drawn.

There are two possible responses in cases like this:
1. Define all such parameters as part of the "model"
1. Partition each of the categories mentioned above into subcategories
   * Fundamental parameters (like, relative risk aversion)
   * Parameters where there is a limiting solution but for which the specific implementation requires specific choices
      * Like, number of points in the discretization of the solution
      * Or, number of agents to simulate in the simulation

sbenthall:
I think I'm following you but I'd like to be sure...
Could you elaborate on what "true" means in this case?

mnwhite:
By "true", he means that we *say* some distribution is lognormal in the
model, but we don't actually integrate a function distributed over a
lognormally distributed RV.  We discretize the RV by one method or another
and compute a numeric integral.  The limit of that method as the number of
discretization points approaches infinity is the "true" function.

Likewise, a lot of macro models assume there's a continuum of agents.  When
those agents are meaningfully heterogeneous, you can't actually track an
infinite number of idiosyncratic states. We simulate a large-ish finite
number of agents, for very small values of "large".

sbenthall:
> By "true", he means that we *say* some distribution is lognormal in the model, but we don't actually integrate a function distributed over a lognormally distributed RV. We discretize the RV by one method or another and compute a numeric integral. The limit of that method as the number of discretization points approaches infinity is the "true" function.

To be clear, are you referring to the simulation process here?

Or as part of the solution?

I think these are both covered by my original proposed categories.

[...]

PR #665, opened 2020-04-28
solver refactoring

sbenthall:
This PR is for refactoring of the solution code, as per #658 

The intention here is to change as little functionality as possible while improving readability.

The first commit intersects with #664 , in that it tries to provide and use a general way to access parameters that may-or-may-not be time-varying.


sbenthall:
@mnwhite This PR is in a good state for your review.

This PR introduces a new `core` function, `onePeriodOOSolver`, which takes a Solver class as an argument and returns a solve* method. This can be called during ConsumerType initialization.

This is also simplifying the solver method arguments: the solver now knows about the agent, and takes its parameter values from the agent rather than as arguments. Because it is sometimes picking from time-varying arguments, the solver also is aware of its current time period `t`.

The one odd wrinkle I wasn't sure about was the way `Rfree` is used in the Kinky solver. It looks like the Kinky solver object was getting passed `Rboro` for the superclass's `Rfree` argument; I'm not sure this is consistent with the (voluminous, and now removed) documentation. Not sure what's going on here.

The aim of this PR is to reduce the amount of code (mostly comments, but very significantly so) and centralize the solver logic so that it is easier to follow. All the methods that defer solution logic to an Object can now be controlled from one block of code.

HARK/core.py
```diff
@@ -832,6 +842,33 @@ def solveOneCycle(agent, solution_last):
...
+    Returns a function that solves a single period consumption-saving
+    problem.
+
+    Parameters
+    ----------
+    solver_class : Solver
+        A class of Solver to be used.
+
+    -------
+    solution_now : Solution
```
sbenthall, line 856: this doc_string needs to reflect the new output (a solver function)

HARK/core.py
```diff
@@ -809,19 +808,30 @@ def solveOneCycle(agent, solution_last):
...
+                temp_dict['solution_next'] = solution_next
+                these_args.remove('solution_next')
+
+            # Make a temporary dictionary for this period
+            temp_dict.update({name: agent.value_at_t(name, T - 1 - t)
+                              for name
+                              in these_args})
+        else:
+            temp_dict['agent'] = agent
+            temp_dict['t'] = T - 1 - t
```
sbenthall, line 833: Sending this time period to the solver is not elegant.

HARK/ConsumptionSaving/ConsPrefShockModel.py
```diff
@@ -257,58 +258,28 @@ class ConsPrefShockSolver(ConsIndShockSolver):
...
 
         Returns
         -------
         None
         '''
-        ConsIndShockSolver.__init__(self,solution_next,IncomeDstn,LivPrb,DiscFac,CRRA,
-                      Rfree,PermGroFac,BoroCnstArt,aXtraGrid,vFuncBool,CubicBool)
-        self.PrefShkPrbs = PrefShkDstn.pmf
-        self.PrefShkVals = PrefShkDstn.X
+        ConsIndShockSolver.__init__(self,agent, t, solution_next)
```
sbenthall, line 280: should be self.agent here (no comma)

HARK/ConsumptionSaving/ConsPrefShockModel.py
```diff
@@ -502,129 +410,27 @@ class ConsKinkyPrefSolver(ConsPrefShockSolver,ConsKinkedRsolver):
...
-        unconditionally on the shock, just before it is revealed.
-    '''
-    solver = ConsKinkyPrefSolver(solution_next,IncomeDstn,PrefShkDstn,LivPrb,
-                             DiscFac,CRRA,Rboro,Rsave,PermGroFac,BoroCnstArt,
-                             aXtraGrid,vFuncBool,CubicBool)
-    solver.prepareToSolve()
-    solution = solver.solve()
-    return solution
+        ## SB: This move has some bad juju
+        agent.Rfree = agent.Rboro
```
sbenthall, line 433: Matt vetoes

HARK/core.py
```diff
@@ -809,19 +808,30 @@ def solveOneCycle(agent, solution_last):
         else:
             solveOnePeriod = agent.solveOnePeriod
 
-        these_args = getArgNames(solveOnePeriod)
 
-        # Update time-varying single period inputs
-        for name in agent.time_vary:
-            if name in these_args:
-                # solve_dict[name] = eval('agent.' + name + '[t]')
-                solve_dict[name] = agent.__dict__[name][T - 1 - t]
```
sbenthall, line 818: keep this place for selecting the time varying params here

HARK/core.py
```diff
@@ -832,6 +842,33 @@ def solveOneCycle(agent, solution_last):
...
+    Parameters
+    ----------
+    solver_class : Solver
+        A class of Solver to be used.
+
+    -------
+    solution_now : Solution
+        The solution to this period's problem.
+    '''
+    def onePeriodSolver(agent, t, solution_next):
```
sbenthall, line 859: `make` onePeriod

sbenthall:
I bit off too much in this PR.
I'm going to close it and file a new PR with _only_ the `makeOnePeriodOOSolver` replacement for the `solve*` functions.
That seemed to have @mnwhite 's favor.

I now think that reconciling the interfaces on the OO solver classes needs to be done, if ever, only _after_ something like #660.

PR #674, opened 2020-05-06
change _hist constructions to history namespace

sbenthall:
Towards #495 ...

In the current functionality, during a simulation the state history is stored on the model object in fields named X_hist.
These state variables are being accessed with `getattr` and set with `setattr`.

This PR replaces storage on the object with storage in a dictionary.
So, X_hist is replaced with `history[X]`.

There are several reasons for this change:
 - Dictionaries are faster than objects. https://gist.github.com/jessicald/2861038
 - The use of objects for storage of model state is part of a design that relies on a lot of switching between representing model variables as strings and representing them as attributes. This design has encouraged a lot of hard-coding of functionality that would ideally be implemented on more general data structures. This is a small step in that direction.

Note: the PR currently passes tests, but it looks like it may break the functionality of ConsAggShockModel. New tests for ConsAggShock are introduced in #654. Please merge that PR first.


sbenthall:
I guess this doesn't break the AggShockModel

mnwhite:
Interesting.  I will try to look at this and merge tomorrow.

sbenthall:
If this change looks good, then the next very low hanging fruit is to replace the setattr and getattrs used for the "now" values to also be replaced with a dictionary, maybe called `state`.

mnwhite:
I'm ok with this, but not sure about the name "state".  Some of the "Now"
variables are states, some are controls, and some are something else
entirely (like the MPC, the derivative of a control with respect to a
state).

sbenthall:
Hmmm. Ok, I definitely like the idea of putting them in separate namespaces.

I'll work that out in a different PR, after this one is merged.

sbenthall:
Feature passes the review, need to fix it in examples.

mnwhite:
Try to make examples work with this.  CDC also wants DEMARKS to work.

sbenthall:
I can't make DemARKs work with code that is not yet committed to HARK.
This is an order of operations issue.

mnwhite:
Yeah, let's get the examples functional and then merge this.

sbenthall:
Latest commit should fix examples

mnwhite:
Excellent.

mnwhite:
...I really thought I merged this right after the examples were working.

PR #680, opened 2020-05-07
makeOnePeriodOOSolver

sbenthall:
This PR replaces the standalone object oriented solve* functions throughout ConsumptionSaving with a call to `makeOnePeriodOOSolver`, which takes a solver class as an argument.

This reduces the number of lines of code--especially redundant comments--by a lot.

It moves the selection of the basic or cubic ConsIndShock solver to the initializer of the type, to make the solver methods more consistent.

This version of the code is doing something a little funky by storing the solver class's initialization arguments on an attribute of the generated solver method. This is to accommodate the fact that in the core solution code, which this PR leaves unchanged, the solution loop is finding the values to pull from the Agent object based on the argument names of the onePeriodSolver. This design is, in my opinion, weird. But changing it would require deeper and more controversial refactorings.

The scope of this PR is intended to a simple, 'Pareto improving' step in code reduction.


sbenthall:
It occurs to me that storing the OO solver constructor arguments on the solver function as I have done in this PR may be too much of a stretch.

I gather that Matt and Chris are committed to the following:
 - a one period solver method should know only about the value of present-time model parameters and the next solution

Currently, there is the following design assumption:
 - the arguments to the solver method correspond to attributes on the model class (though perhaps a time-varying attribute that requires a subscript before being passed to the solver)

In that case, maybe I was barking up the wrong tree trying to generalize around the OO solver.
On further inspection, it looks like the OO solver design is mainly there to break up a lot of otherwise complex functionality into submethods. But these methods operate pretty much in a sequence. I suppose this is why Matt's been favoring a non-OO, single function design in more recent solvers.

In other words, the main thing making the solve code so complex is that it's currently caught between two different design paradigms: functional, and object oriented. It may be that the best step forward would be to complete the transition towards the functional design, throughout HARK.

The main thing I'm not clear on yet is to what extent there is a generalizable logic to the backwards induction solvers. They are currently written in a way that all the state and control variables are hard-coded separately. But my understanding of the backwards induction technique is that it generalizes.

Maybe the best way to make headway would be to start trying to break down the OO solver into a single function. A new module `solver.py` could contain more of the functions that currently compose the OO solvers. In the process of working on that, maybe some redundancies would expose themselves.

I understand that these solvers are in a sense the core of the economic research being done here. Maybe they are truly so complex that there's really no way to tame them, at least without graduate level education in the subject matter. On the other hand, it seems like Matt has some misgivings about the current design and I'm definitely willing to try to fix things up, especially if he can provide direction about how to do that. I've been working with this code for months now and these solvers are still almost entirely opaque to me (the notation is not helping me....) and that's rather frustrating.

sbenthall:
@mnwhite I believe this PR passed your review on Thursday and you were just waiting for tests to finish running after the update to merge.

mnwhite:
Yes, will merge momentarily.

PR #682, opened 2020-05-08
Fix typo in parallel.py

mnwhite:
parallelNelderMead still refers to maxcores in its code, but the parameter is maxthreads; not sure how this even worked before.  Now fixed.


MridulS:
@mnwhite I tried looking for `parallelNelderMead` in HARK/REMARK/DemARK but couldn't find a reference.

This should have a test.

PR #691, opened 2020-05-14
set seed on distribution initialization, not draw. 

### LINKED ISSUE #619: discrete distributions should get a random seed on initialization
`DiscreteDistribution.drawDiscrete` takes a `seed` argument.

Instead, this seed can be an argument to the DiscreteDistribution upon initialization.

sbenthall:
fixes #619


sbenthall:
It looks like Windows handles maximum integer length differently from MacOS and Ubuntu.
My understanding was that with Python 3, ints are of unbounded length.
But on Windows it looks like they are still int32.

I suppose it's worth articulating what I'm trying to do in the failing code...

HARK seems to be designed so that a model performs identically, though pseudorandomly, when ran twice. That's done through default random seeds.

Before, the seeds were passed to the distribution code when the distribution was drawn from.
The change here is to give a distribution a seed when it's initialized.
That means that the draws from a distribution won't by default have identical values.

One issue raised by this new design is whether and how to the seeds from a continuous distribution to the new distribution created by its discretization.

In this code, it's drawing that seed pseudorandomly from the continuous distribution's generator.

sbenthall:
Debugging this is hard because I can't run the WIndow build natively.

But I'm trying these fixes, in light of the fact that RandomState `seed`s must be convertible to unsigned 32-bit integers:
 - explicitly setting 'int32' dtype on the integer generators within the distribution code
 - setting the maximum at `2**31-1` because one bit is used for the positive/negative sign, so when converted to an unsigned integer the bit will be redundant...

See: https://github.com/numpy/numpy/issues/4085

mnwhite:
Add quick reset() method to set RandomState (to same value).

sbenthall:
When does a reset() method need to be called by the AgentType class?

Currently, most distributions used in solution and simulation are constructed just before they are used, then discarded.

I would like to eventually change that so that the distributions are provided when the model is initialized, but things don't work like that yet.

mnwhite:
Oh!  That could definitely be the source of the performance slowdown.
reset() would be called during a Market's solve() loop, right before the
model is simulated again.

sbenthall:
`reset()` method added in last commit

mnwhite:
Should we merge this?

mnwhite:
Ugh, we'll have to adjust the AggShock economy test targets, as this PR changes the random seed.  The targets change because the tests don't actually solve the model, but instead quit after 3 passes; the idiosyncratic shocks thus matter.

mnwhite:
I thought I merged this.  Pretty sure it's ready for merge after these checks run.

PR #692, opened 2020-05-15
adding distributeParams 

sbenthall:
See #669


HARK/core.py
```diff
@@ -938,6 +938,36 @@ def __init__(self, agents=None, sow_vars=None, reap_vars=None, const_vars=None,
...
+        # Distribute the parameters to the various types, assigning consecutive types the same
+        # value if there are more types than values
+        replication_factor = len(self.agents) // param_count 
+            # Note: the double division is intenger division in Python 3 and 2.7, this makes it explicit
+        j = 0
+        b = 0
+        while j < len(self.agents):
+            for n in range(replication_factor):
+                self.agents[j](AgentCount = int(self.Population*param_dist[0][b]*self.TypeWeight[n]))
+                exec('self.agents[j](' + param_name + '= param_dist[1][b])')
```
sbenthall, line 967: These two lines, copied from what I assume is ancient cstwMPC code, scare me.
It hardly looks like valid Python to me.

I think also it's using [0] and [1] to access the distribution values, which I'll need to fix for sure.

sbenthall:
This PR now does something a little different from the original method.

- It's a standalone function in `core`, not a method on `Market`.
- Rather than make any assumptions about the total population in the market, it now works by cloning and splitting an AgentType, given in an argument, into N ex ante heterogeneous copies.
- It takes a (continuous) distribution as an argument.
- the AgentCount of the original agent is divided between the new types in proportion given by the discretized distribution.

mnwhite:
I need to look at this, but I like your description.

mnwhite:
This is on my task list to merge.

PR #693, opened 2020-05-19
Move neldermead code to a new file and clean up parallel 

MridulS:
- Clean up imports in parallel, add rudimentary type hints support to Parallel (https://github.com/econ-ark/HARK/issues/664)
- Add base tests for parallel
- move neldermead code to neldermead.py

@mnwhite I couldn't find any public files in econ-ark which use `parallelNelderMead`. To test it out `from HARK.neldermead import parallelNelderMead ` should work.


mnwhite:
It should be moved to HARK.estimation, not a new file.

mnwhite:
Great, thanks.

Issue #699, opened 2020-05-26
checkConditions runs automatically and prints output to screen


mnwhite:
Solving an IndShockConsumerType automatically prints output to screen.  This is absolutely not what we want to happen.

sbenthall:
I believe this is connected to #283 
At the end of that issue is a proposal for a new design for logging that has passed @llorracc 's review.

mnwhite:
Whoops, I forgot we had already discussed this.

sbenthall:
it's all good. Thanks for reminding me I need to fix it

sbenthall:
@mnwhite I think you and @llorracc may have a design disagreement here?

https://github.com/econ-ark/HARK/issues/283#issuecomment-600279552

I think that @llorracc wants checkConditions to normally print _some_, but not all, of the messages when IndShockConsumerType is solved(). He has asked that the "quiet" option be something that can be set for individual notebooks.

If I understand you correctly, I think you are saying here that you think that turning on the print statements for checkConditions should be opt in.

From my perspective, this is a matter of deciding which of three proposed logging settings (normal, "verbose", or "quiet") is the default. I have no opinion on this.

I'll get the flexible logging framework described in #283 in as a PR and let you and @llorracc hash out the details here, since it seems to be mainly a question of how best to address the tensions between pedagogical and research use cases.

mnwhite:
This was fixed long ago, closing.

PR #701, opened 2020-05-26
make HARK less "verbose" -- introduce logger 

### LINKED ISSUE #283: replace print with logging, with 'quiet' and 'verbose' logging configuration options.
PerfForesightConsumerType now has attributes called verbose and quiet.  I'm not sure when these were added.

1) It seems like only one of these should exist, as they sort of mean the opposite of each other.

2) They are both currently set to default to False, which feels weird.  This instance is neither verbose nor quiet?

sbenthall:
…with variable configuration levels. Fixes #283

The way you can change the verbosity level at the top of a notebook is now, e.g.:

```
import HARK.logger as logger
logger.verbose()
```

or

```
import HARK.logger as logger
logger.quiet()
```

I've shown this at the top of one of the example notebooks to demonstrate.

This PR so far only introduces the logger's functionality for the PerfForesightConsumerType and IndShockConsumerType, which have implemented checkConditions() methods.
If the design passes review, I can change other print statements to logging.info statements.


MridulS:
I probably missed this conversation regarding this but shouldn't
``` python
import HARK.logger as logger
```
be used in ConsIndShockModel.py?

Users should use something like
```
IndShockConsumerType(verbose=True)
```
instead of creating a "global" loggers while creating an example/notebook?

sbenthall:
@MridulS No, I don't think so.

In this, PR, ConsIndShockModel.py imports `logging`, and then uses `logging.info`, `logging.warning`, or `logging.debug` instead of `print` statements.

This effectively puts the logging module, which is a Python standard library, as an intermediary between these statements and STERR. (With a little more sophisticated tuning, it could pipe these messages to STOUT, but this is the simplest implementation currently.)

This is done "globally" because that's the simplest implementation. If we needed something more complex, we could do it. But that would require more expressing logging configuration.

I think it's not smart to have verbosity be a model parameter.

Logging configuration _should_ be done at the global level because really it's a configuration of the Python environment's I/O. Specifically, the O. Trying to patch it through the program logic creates a lot of messy interaction between model, controller, and view logic. It's harder to maintain and makes it difficult to give the user the ability to turn verbosity on or off.

sbenthall:
Oh, hmmm. @MridulS you might be right about something actually.

Really, it depends on what the default behavior. This is disputed. See #699

sbenthall:
I'll add the logger to the __init__ file as per this PR from Mridul

https://github.com/pymc-devs/pymc3/pull/1416/files

sbenthall:
In the latest commit to this PR, the logging configuration is moved to `HARK.__init__`
Now this code is executed when HARK packages are imported.

For library code, the HARK logger needs to be invoked explicitly. I.e.:
```
from HARK import _log

_log.info("An informative message")
```

The HARK user does not need to do anything extra to run this code! This is good.

However: in the current PR:
 - Most (but not all) of the checkConditions messages are at the INFO level
 - This INFO level is logged to the user by default for HARK.
 - checkConditions is run automatically before `solve()` is called.

This means that for the current PR, the user will automatically see the checkConditions() messages.
This I believe satisfies the requirements of #283 but contradicts #699 

This looks like a difference of opinion between @mnwhite and @llorracc and I defer to their consensus on the final implementation.

mnwhite:
I need to check the boolean flags for verbosity the default dictionaries; there is no disagreement otherwise.

llorracc:
This is superseded by #714

PR #698, opened 2020-05-26
MNW cleanup May 2020

mnwhite:
Various small comment and functionality fixes for ConsAggShockModel examples and FashionVictim.

PR #690, opened 2020-05-14
Tim Munday's original work on nan_bool

mnwhite:
Reissued PR to dev branch so I can sort out the argument order issue.

Not ready to merge, need to update.

PR #702, opened 2020-05-26
Fix simulation performance of AggShockMarkovConsumerType

mnwhite:
Implementation of DiscreteDistribution in AggShockMarkovConsumerType was incomplete/awkward, leading to extreme performance problems for simulation.  Now fixed.

Also removed references to timeFwd() in one of the examples, and maybe a couple other comment or whitespace tweaks that happened to be in my active changes.


mnwhite:
This includes my other two recent PRs.  See those threads for other explanation.

sbenthall:
By "other recent PRs" you mean #690  and #698  ?

sbenthall:
There are some edits to `examples/ConsumptionSaving/example_ConsAggShockModel.py` and `examples/ConsumptionSaving/example_ConsIndShock.py`.

There are also `.ipynb` versions of these examples.

Ideally, the two are synced up. I'm a little confused at this point as to why they aren't. We've been using jupyter lab/jupytext to keep those working together.

sbenthall:
If the performance issue has been fixed, could you also reenable the automated test?

mnwhite:
I'll try to do that, yes.  But I don't think it was ever enabled?  Maybe it
was.

Yes, those are the PRs.  That's why there's random stuff like edits to
examples in this PR-- I pulled in MNWcleanup so I could run the relevant
example.

I don't know anything about the notebooks.  Are we supposed to generate
notebooks from py files, or vice versa?

MridulS:
Those are supposed to be synced up using jupytext. (Install it using pip install jupytext if not already present)

To sync a paired notebook with a .py file ->

jupytext  --sync path_to_python_file.py

sbenthall:
Let's put the jupytext thing into the contributor guidelines

mnwhite:
We'll need to adjust the target for one AggShock test due to the change in the RNG.  This test doesn't fully solve, but instead "quits early", so the distribution of shocks matters for the "result".

mnwhite:
Tests are now passing, will be ready to merge soon.

mnwhite:
This is ready to be merged again.

PR #723, opened 2020-06-22
[WIP] old numba HARK: Will be replaced by piecemeal bits

alanlujan91:
This PR adds basic implementation of numba just-in-time compiling to `PerfForesightConsumerType` and `IndShockConsumerType`. 

Specifically, I've added `ConsPerfForesightSolverNumba `and `ConsIndShockSolverBasicNumba`. 

Additionally, I've created a "flat" solver for `PerfForesightConsumerType `as `PerfForesightNumbaModel`.


alanlujan91:
## `ConsPerfForesightSolverNumba`: A Numba solver for  `PerfForesightConsumerType`

The class [`PerfForesightConsumerTypeNumba`](https://github.com/alanlujan91/HARK/blob/f9927820d0d9d972e27be1689dbbc8acc9cb64e0/HARK/ConsumptionSaving/ConsIndShockNumba.py#L544-L548)is a child class of `PerfForesightConsumerType`, inheriting most of the existing structure but changing the solver to a ``numba`` implementation called [`ConsPerfForesightSolverNumba`](https://github.com/alanlujan91/HARK/blob/f9927820d0d9d972e27be1689dbbc8acc9cb64e0/HARK/ConsumptionSaving/ConsIndShockNumba.py#L321). 

The class [`ConsPerfForesightSolverNumba`](https://github.com/alanlujan91/HARK/blob/f9927820d0d9d972e27be1689dbbc8acc9cb64e0/HARK/ConsumptionSaving/ConsIndShockNumba.py#L321) itself inherits from the standard `ConsPerfForesightSolver` class, but modifies certain methods to allow for `numba` just-in-time compiling. In particular, the methods [`makePFcFunc`](https://github.com/alanlujan91/HARK/blob/f9927820d0d9d972e27be1689dbbc8acc9cb64e0/HARK/ConsumptionSaving/ConsIndShockNumba.py#L327-L361) and [`addSSmNrm`](https://github.com/alanlujan91/HARK/blob/f9927820d0d9d972e27be1689dbbc8acc9cb64e0/HARK/ConsumptionSaving/ConsIndShockNumba.py#L363-L390) are computationally intensive and would benefit most from jit-compiling. 

First, I created an global jitted function called [`makePFcFuncNumba`](https://github.com/alanlujan91/HARK/blob/f9927820d0d9d972e27be1689dbbc8acc9cb64e0/HARK/ConsumptionSaving/ConsIndShockNumba.py#L44-L146) that takes only primitives and (``numpy``) arrays. While most of the code can stay as in `makePFcFunc`, `numba` implementations of `numpy` ufuncs can only take `numpy` arrays, and not lists. This is easily adaptable by replacing any `[x,y]` with `np.array([x,y])` , as in [this instance](https://github.com/alanlujan91/HARK/blob/GPUHARK/HARK/ConsumptionSaving/ConsIndShockNumba.py#L107-L110). Any assignments to objects or code that can not be implemented in `numba` are left to the class function to take care of, such as creating a `LinearInterp` instance for `cfunc` [here](https://github.com/alanlujan91/HARK/blob/f9927820d0d9d972e27be1689dbbc8acc9cb64e0/HARK/ConsumptionSaving/ConsIndShockNumba.py#L361). 

Implementing [`addSSmNrmNumba`](https://github.com/alanlujan91/HARK/blob/f9927820d0d9d972e27be1689dbbc8acc9cb64e0/HARK/ConsumptionSaving/ConsIndShockNumba.py#L159-L174)took a bit more work. This method uses calls to `cfunc`, which is an instance of `LinearInterp`, to find roots using `scipy.optimize.newton`. Again, `numba` does not (yet) support user defined classes (although there is `@jitclass`, which could be interesting to look into) or non-jitted functions such as those from `scipy`, so we need alternatives to both. As a replacement to `LinearInterp`, I used interpolation.py from [EconForge](https://github.com/EconForge). This replacement is straightforward; instead of cfunc(m), I take the parameters that created `cfunc` and use [`interp(mNrmNow, cNrmNow, m)`](https://github.com/alanlujan91/HARK/blob/GPUHARK/HARK/ConsumptionSaving/ConsIndShockNumba.py#L154) which returns the interpolated value of the consumption function at `m`. 

In the standard (non-`numba`) implementation of `ConsPerfForesightSolver`,  `scipy`'s newton takes 2 parameters, `func` and `x0`. This means, according to the documentation, that without `fprime`, `scipy` will default to a newton secant algorithm. Fortunately, I know just the library that has jitted optimization functions, [QuantEcon](https://quantecon.org/). I thus use [`quantecon.optimize.newton_secant`](https://github.com/QuantEcon/QuantEcon.py/blob/ca8fb06b52b3cb9b5a53702fbd1090eb7a6dcbd5/quantecon/optimize/root_finding.py#L196-L274) in place of `scipy`'s newton for root-finding. Another shortcoming of `numba` is that lambda functions can not be passed as arguments to jitted functions, such as quantecon's `newton_secant`. Thus I create an additional [`searchSSfuncNumba`](https://github.com/alanlujan91/HARK/blob/f9927820d0d9d972e27be1689dbbc8acc9cb64e0/HARK/ConsumptionSaving/ConsIndShockNumba.py#L149-L156) to be used by `newton_secant`.

alanlujan91:
## `ConsPerfForesightNumbaSolver`: A "flat" Numba solver for  `PerfForesightConsumerType`

The file [**PerfForesightNumbaModel.py**](https://github.com/alanlujan91/HARK/blob/GPUHARK/HARK/ConsumptionSaving/PerfForesightNumbaModel.py) contains the class [`ConsPerfForesightNumbaSolver`](https://github.com/alanlujan91/HARK/blob/GPUHARK/HARK/ConsumptionSaving/PerfForesightNumbaModel.py#L182) which implements a "flat" solver for 

`PerfForesightConsumerType`. This "flat" solver uses a single, jitted function, to create a solution to the problem presented, along with a helper function. The function [`solveConsPerfForesightNumba`](https://github.com/alanlujan91/HARK/blob/f9927820d0d9d972e27be1689dbbc8acc9cb64e0/HARK/ConsumptionSaving/PerfForesightNumbaModel.py#L55) performs the same optimizations as `ConsPerfForesightSolverNumba`, but does so without calling as many additional and helper methods along the way. The code still needs a helper function, [`searchSSfunc`](https://github.com/alanlujan91/HARK/blob/GPUHARK/HARK/ConsumptionSaving/PerfForesightNumbaModel.py#L37-L51), to find steady state market resources.

MridulS:
This is great Alan!

To test if we can swap in the Numba alternative with the current switch all instances of `PerfForesightConsumerType` in with something like below at the top of the files to make sure this will not break anything inside HARK.
```
from HARK.ConsumptionSaving.ConsIndShockNumba import  PerfForesightConsumerTypeNumba
PerfForesightConsumerType = PerfForesightConsumerTypeNumba
```

HARK/ConsumptionSaving/PerfForesightNumbaModel.py
```diff
@@ -0,0 +1,379 @@
...
+    m_init_guess = mNrmMinNow + ExIncNext
+
+    try:
+        mNrmSS = newton_secant(
+            searchSSfunc,
+            m_init_guess,
+            args=(PermGroFac, Rfree, ExIncNext, mNrmNow, cNrmNow),
+        )[0]
+    except:
+        mNrmSS = None
```
MridulS, line 166: check if it possible just to use the converge information from `newton_secant` to rewrite the try, except code block.
Do check when newton_secant fails (so that we know the reason why there is this try, except code block)

llorracc: I'm actually a bit surprised that newton_secant does _not_ fail for the PF model, since the "secant" for a line is degenerate.  But it seems to have gotten the right answer.

alanlujan91:
## INDSHOCKCONSUMERTYPENUMBA

Since the last commit I've worked on a `numba` version of `IndShockConsumerType` in [IndShockNumbaModel.py](https://github.com/alanlujan91/HARK/blob/GPUHARK/HARK/ConsumptionSaving/IndShockNumbaModel.py). The class [`IndShockConsumerTypeNumba`](https://github.com/alanlujan91/HARK/blob/31fe021a58746d0156842d4b6fd2b3f47b46c3c2/HARK/ConsumptionSaving/IndShockNumbaModel.py#L935) is the same as `IndShockConsumerType`, but now I use a "flat" `numba` optimized solvers called [`ConsIndShockNumbaSolverBasic`](https://github.com/alanlujan91/HARK/blob/31fe021a58746d0156842d4b6fd2b3f47b46c3c2/HARK/ConsumptionSaving/IndShockNumbaModel.py#L451) and [`ConsIndSchockSolverNumba`](https://github.com/alanlujan91/HARK/blob/31fe021a58746d0156842d4b6fd2b3f47b46c3c2/HARK/ConsumptionSaving/IndShockNumbaModel.py#L757).

### `ConsIndShockNumbaSolverBasic`

The method [`prepareToSolve`](https://github.com/alanlujan91/HARK/blob/31fe021a58746d0156842d4b6fd2b3f47b46c3c2/HARK/ConsumptionSaving/IndShockNumbaModel.py#L628) implements the machinery of `ConsIndShockSetup`. Beyond the steps that have been described before to translate code from standard python to `numba`, this method didn't need much alteration.

#### `solveConsIndShockNumba` in `ConsIndShockNumbaSolverBasic`

The method [`solve`](https://github.com/alanlujan91/HARK/blob/31fe021a58746d0156842d4b6fd2b3f47b46c3c2/HARK/ConsumptionSaving/IndShockNumbaModel.py#L686) implements the machinery of `ConsIndShockSolverBasic`. I will describe the few modifications that were needed in detail.

The first hurdle was due to the fact that `numba` does not yet have optimized implementations of [`np.tile`](https://github.com/alanlujan91/HARK/blob/31fe021a58746d0156842d4b6fd2b3f47b46c3c2/HARK/ConsumptionSaving/IndShockNumbaModel.py#L40-L54) and [`np.insert`](https://github.com/alanlujan91/HARK/blob/31fe021a58746d0156842d4b6fd2b3f47b46c3c2/HARK/ConsumptionSaving/IndShockNumbaModel.py#L57-L73). With the hopes that these methods are implemented in the future, I wrote two basic 'placeholders' that take the same parameters in the exact same format, and instead uses a combination of other `numpy` methods that `numba` does support. For example, the way `np.insert` is being used is to insert `values` into `array` before index 0, which can be done instead by converting `values` to an array, then appending `array` at the end. If and when `numba` implements these two methods, the code can be changed by adding `np.-` back to the call.

The next big hurdle was to replace the call to [`vPfuncNext`](https://github.com/alanlujan91/HARK/blob/31fe021a58746d0156842d4b6fd2b3f47b46c3c2/HARK/ConsumptionSaving/IndShockNumbaModel.py#L221-L223), a HARK object, in `calcEndOfPrdvP`. After inspection, it appears that `vPfuncNext` is a `MargValueFunc`, which is itself an instance of `LowerEnvelope`. What `MargValueFunc` does is take a `cFunc` and apply the function `c**-gam`, so I can just do that myself, assuming I can calculate `c` first. This is where `LowerEnvelope `comes into place, since `cFunc` is an instance of `LowerEnvelope` and thus composed of several functions. Assuming I can get `cFunc.x_list` and `cFunc.y_list` (for now, will touch on this later), I can linearly interpolate `cFuncNext` given `mNrmNext`, using `interp` from interpolation.py.

Doing this, however, was not as easy as expected. I may have found a bug (or perhaps a feature request) on the method `interp`. Calling `interp(x_list,y_list,values)` where `values` is an `ndarray` (more than 1 dimension), returned only an array output (1 dimension) which appeared to only interpolate along the first value of axis 0. More clearly, if `values.shape = (x,y)`, then `output.shape = (x,)`. Clearly, this is not what I need, so to fix this issue, I flatten `values`, then `reshape` the output. The final code looks like this:

```
cFuncNext = interp(sn_cFunc_x_list, sn_cFunc_y_list, mNrmNext.flatten())
vPfuncNext = (cFuncNext ** -CRRA).reshape(mNrmNext.shape)
```

As I mentioned before, I assumed the existence of `cFunc.x_list` and `cFunc.y_list`, which are inputs to the `solveConsIndShockNumba` method. Since `cFunc` is an instance of `LowerEnvelope`, these attributes do not exist directly, therefore, I need to create them. For `cFunc.x_list`, I simply used `self.aXtraGrid`, which is the resource grid. For `cFunc.y_list` I unfortunately need to call [`cFunc(cFunc_x_list)`](https://github.com/alanlujan91/HARK/blob/31fe021a58746d0156842d4b6fd2b3f47b46c3c2/HARK/ConsumptionSaving/IndShockNumbaModel.py#L700-L701), which uses the internals of `LowerEnvelope` to linearly interpolate over `cFunc_x_list`. This linear interpolation is not `numba` optimized, and may be a bottleneck to the code.

MridulS:
> interp(x_list,y_list,values) where values is an ndarray (more than 1 dimension), returned only an array output (1 dimension)

You can open an issue at https://github.com/EconForge/interpolation.py regarding the interp dimension issue.

mnwhite:
aXtraGrid is *not* the same thing as the grid of market resources on which
we're interpolating.

alanlujan91:
> aXtraGrid is *not* the same thing as the grid of market resources on which we're interpolating.

I had tried this instead and it didn't give correct results:

```
        if isinstance(self.solution_next.cFunc, LowerEnvelope):
            sn_cFunc_x_list = self.solution_next.cFunc.functions[0].x_list
        else:
            sn_cFunc_x_list = self.solution_next.cFunc.x_list
```

I think in my solution it worked because `BoroCnstArt` = 0. Perhaps `aXtraGrid + BoroCnstNat` would be more general?

mnwhite:
What happened when you used x_list as the list of interpolation nodes?  I
mean, it *is* the list of interpolation nodes.

mnwhite:
Ah ok, I've read that more carefully now.  Are you trying to create a
single linear interpolation that represents the consumption function,
rather than the "two piece" style we use with LowerEnvelope?  So you want a
set of interpolation nodes (m_t values) and corresponding function levels
(c_t values), is that correct?

llorracc:
> aXtraGrid is *not* the same thing as the grid of market resources on which we're interpolating.

Per the [NARK](https://econ-ark.github.io/HARK/Documentation/NARK/#RomalLetters), 

Per [NARK/RomanLetters](https://econ-ark.github.io/HARK/Documentation/NARK/#RomanLetters)
1. a is \textit{A}ssets \textit{A}fter \textit{A}ll \textit{A}ctions \textit{A}re \textit{A}ccomplished (end of period)
1. h is human wealth

Per [NARK/Modifiers](https://econ-ark.github.io/HARK/Documentation/NARK/#Modifiers)
1.  aXtraGrid is the list of points added to the minimum value of a

mnwhite:
Yeah, it turns out he wasn't using it incorrectly, I don't think.  He was
using it as an ex-post grid of mNrm values, and then getting the
corresponding grid of cNrm values as cFunc(that_grid).

alanlujan91:
> Ah ok, I've read that more carefully now. Are you trying to create a single linear interpolation that represents the consumption function, rather than the "two piece" style we use with LowerEnvelope? So you want a set of interpolation nodes (m_t values) and corresponding function levels (c_t values), is that correct?

Correct. If cFunc were just a simple LinearInterp, I could extract x_list and y_list directly. For this case, setting on an x_list and interpolating the corresponding y_list seemed like the most appropriate approach, but I'm open to suggestions!

alanlujan91:
I worked on a `numba` version of [`ConsIndShockSolver`](https://github.com/alanlujan91/HARK/blob/e2486f3da3dde8707ca4f8af0a3177e6ccc1e2ef/HARK/ConsumptionSaving/IndShockNumbaModel.py#L757) in `IndShockNumbaModel.py`. For this, I created two `numba` helper functions, [`solveConsIndShockCubicNumba`](https://github.com/alanlujan91/HARK/blob/e2486f3da3dde8707ca4f8af0a3177e6ccc1e2ef/HARK/ConsumptionSaving/IndShockNumbaModel.py#L255) for solving using a cubic approximation for the consumption policy function, and [`addvFuncNumba`](https://github.com/alanlujan91/HARK/blob/e2486f3da3dde8707ca4f8af0a3177e6ccc1e2ef/HARK/ConsumptionSaving/IndShockNumbaModel.py#L366) for calculating the value function during the solution.

### `solveConsIndShockCubicNumba`

In the previous post, I described the process of linearly interpolating the value function using `interp` from i`nterpolation.py`. For the cubic version, I needed to use a similar cubic interpolator that worked with `numba`, but using tools from `interpolation.py` did not seem straightforward. However, I had created a basic `numba` cubic interpolator for my own purposes previously, and added these functions to [`HARK.numba`](https://github.com/alanlujan91/HARK/blob/e2486f3da3dde8707ca4f8af0a3177e6ccc1e2ef/HARK/numba.py#L371). A more thorough overview of what these do is [presented here](https://gist.github.com/alanlujan91/21975ca59f866eed10b759e0d5cab79e), but below I briefly explain. 

The function [`splrep`](https://github.com/alanlujan91/HARK/blob/e2486f3da3dde8707ca4f8af0a3177e6ccc1e2ef/HARK/numba.py#L371-L402) taxes 1-d vectors `x` and `y`, where `y=f(x)` for some function `f`, and creates interpolating coefficients using a complete cubic spline method. The function returns a 1-d vector of coefficients, which we can call `z`. The function [`splevec`](https://github.com/alanlujan91/HARK/blob/e2486f3da3dde8707ca4f8af0a3177e6ccc1e2ef/HARK/numba.py#L407-L438) then takes 1-d vectors `x0`, `x`, `y`, and `z`, where `y=f(x)`, `z` are interpolating coefficients, and `x0` contains values where we need an approximation for f, and returns an approximation `y0=g(x0)`, where `g` is the cubic spline approximation. The [final code](https://github.com/alanlujan91/HARK/blob/e2486f3da3dde8707ca4f8af0a3177e6ccc1e2ef/HARK/ConsumptionSaving/IndShockNumbaModel.py#L314-L318) looks like this:

```python
cFuncCoeffs = splrep(sn_cFunc_x_list, sn_cFunc_y_list)
cFuncNext = splevec(mNrmNext.flatten(), sn_cFunc_x_list, sn_cFunc_y_list, cFuncCoeffs)
vPfuncNext = (cFuncNext ** -CRRA).reshape(mNrmNext.shape)
```

The same process is used to calculate [`vPPfuncNext`](https://github.com/alanlujan91/HARK/blob/e2486f3da3dde8707ca4f8af0a3177e6ccc1e2ef/HARK/ConsumptionSaving/IndShockNumbaModel.py#L335-L339). 

### `addvFuncNumba`

This section uses the same process above to calculate a cubic approximation of [`vfuncNext`](https://github.com/alanlujan91/HARK/blob/e2486f3da3dde8707ca4f8af0a3177e6ccc1e2ef/HARK/ConsumptionSaving/IndShockNumbaModel.py#L391-L394) on current resources. The only novel part of this section is the use of `@vectorize` on the different kind of utility functions. Instead of importing utilities from `HARK.utilities`, I have created basic `numba` implementations of utilities by simply adding `@vectorize` decorators in [`HARK.numba`](https://github.com/alanlujan91/HARK/blob/e2486f3da3dde8707ca4f8af0a3177e6ccc1e2ef/HARK/numba.py#L18). In the future, these vectorized functions could be easily adaptable to GPU processing by adding [`target="cuda"`](https://numba.pydata.org/numba-doc/latest/cuda/ufunc.html) as a parameter. Instead of creating lambda functions with a set CRRA parameter, however, I use the full call [`f(c,crra)`](https://github.com/alanlujan91/HARK/blob/e2486f3da3dde8707ca4f8af0a3177e6ccc1e2ef/HARK/ConsumptionSaving/IndShockNumbaModel.py#L429) for all `f` in utilities, which is used to reduce overhead and prevent `numba` recompiling. The solver appears to work, although there will be numerical differences given different interpolators that need to be explored further.

MridulS:
@alanlujan91 This can be closed now right?

alanlujan91:
yes

PR #749, opened 2020-07-02
Deal with cases when the agent wants to short

Mv77:
This implements the solution proposed by @mnwhite  in #741 for dealing with cases where the agent wants to hold less than a 0% share of the risky asset.

I did some testing and it seems to be well behaved. The approach assumes that as long as the FOC at s=0 is negative, the agent wants to hold less than 0. I guess it will break if one considers non-concavities in the value function?

I got rid of the try-except block as @mnwhite suggested.

### DIFF
--- a/HARK/ConsumptionSaving/ConsPortfolioModel.py
+++ b/HARK/ConsumptionSaving/ConsPortfolioModel.py
@@ -778,18 +778,20 @@ def solveConsPortfolio(solution_next,ShockDstn,IncomeDstn,RiskyDstn,
         FOC_s = EndOfPrddvds
         Share_now   = np.zeros_like(aNrmGrid) # Initialize to putting everything in safe asset
         cNrmAdj_now = np.zeros_like(aNrmGrid)
-        constrained = FOC_s[:,-1] > 0. # If agent wants to put more than 100% into risky asset, he is constrained
-        Share_now[constrained] = 1.0
+        constrained_top = FOC_s[:,-1] > 0. # If agent wants to put more than 100% into risky asset, he is constrained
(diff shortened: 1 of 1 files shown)


llorracc:
> I guess it will break if one considers non-concavities in the value function?

Yes; but anything will break under those circs.  We may need to handle this in the future (esp when trying to have a discrete-continuous problem jointly with the portfolio problem) but that's a problem for another day.

mnwhite:
Yeah, the code assumes a single-crossing property / monotone FOC, but
that's consistent with most of our other solvers, which assume that the
(marginal) value function being fed to it from solution_next came from the
solution to the same model and thus is guaranteed to have certain
properties.

A while back I wrote a "robust" ConsIndShock solver that used Giu Fella's
method for handling non-concave value functions.  Just like my attempt at
implementing his discrete labor supply choice model with that method, it
put the "bust" in "robust".  18 months ago I began work with Fedor Iskhakov
on an improved method.

Anyway, thanks for doing this so fast, I'll take a look soon.

Mv77:
> 
> 
> > I guess it will break if one considers non-concavities in the value function?
> 
> Yes; but anything will break under those circs. We may need to handle this in the future (esp when trying to have a discrete-continuous problem jointly with the portfolio problem) but that's a problem for another day.

Yes, exactly. Since I know we will eventually have to deal with this, I just want to keep a mental list of where things might break.

mnwhite:
Yes, this is exactly correct.  And I guess I had already handled the 100% constraint, now labeled as `top_constraint`; that might be why I thought I dealt with the 0% constraint as well.  Nice work.

PR #703, opened 2020-05-27
add black pre commit hook

MridulS:
This PR adds pre commit hooks for black (python auto-styling tool) and will fix up issues about python code style choices https://github.com/econ-ark/HARK/issues/310, https://github.com/econ-ark/HARK/issues/262 .

pre commit hooks adds the ability to auto run `black` on changed files locally before creating a commit and pushing it. This can be extended to run other auto processes like a linter or a type checker like mypy if and when we have type checks in our codebase.

This is how it looks like:
``` 
(econ) ➜  HARK git:(precommit) ✗ vim HARK/dcegm.py
(econ) ➜  HARK git:(precommit) ✗ git add HARK/dcegm.py
(econ) ➜  HARK git:(precommit) ✗ git commit -m 'test change'
black....................................................................Failed
- hook id: black
- files were modified by this hook

reformatted HARK/dcegm.py
All done! ✨ 🍰 ✨
1 file reformatted.
```

As the black test failed, it auto ran black for all the files in the staging area (files added through `git add`) and another add and commit is required on top of this. Which will look like something like this.

```

(econ) ➜  HARK git:(precommit) ✗ git add HARK/dcegm.py
(econ) ➜  HARK git:(precommit) ✗ git commit -m 'suggested black changes'
black....................................................................Passed
[precommit 02beafd] suggested black changes
 1 file changed, 36 insertions(+), 16 deletions(-)
```

As this requires a change to personal development style to make sure our codebase is slowly and steadily black-compliant, please do look through this @llorracc @mnwhite @sbenthall .

To install `pre-commit` locally and add these hooks the following steps are required:
```
$ pip install precommit
$ pre-commit install # in the HARK/ directory 
```

[Even if this is merged in, if precommit isn't installed this will have no effect on local development environments]

PR #757, opened 2020-07-11
Some RNG fixes

mnwhite:
One of our DemARKs was malfunctioning because our simulator wasn't replicating the same sequence of random shocks after initializeSim() was called.  This PR fixes that problem in all of our models (I think) by implementing extensions to AgentType.resetRNG() that reset the RNG of each discrete distribution used by that type.


HARK/ConsumptionSaving/ConsIndShockModel.py
```diff
@@ -1991,16 +1991,41 @@ def update(self):
...
+        -------
+        None
+        '''
+        PerfForesightConsumerType.resetRNG(self)
+        
+        # Reset IncomeDstn if it exists (it might not because resetRNG is called at init)
+        if hasattr(self, 'IncomeDstn'):
+            T = len(self.IncomeDstn)
+            for t in range(T):
+                self.IncomeDstn[t].reset()
```
sbenthall, line 2026: possible rewrite:

for distribution in self.IncomeDstn:
  distribution.reset()

HARK/ConsumptionSaving/ConsMarkovModel.py
```diff
@@ -782,6 +783,32 @@ def initializeSim(self):
...
+        ----------
+        None
+
+        Returns
+        -------
+        None
+        '''
+        PerfForesightConsumerType.resetRNG(self)
+        
+        # Reset IncomeDstn if it exists (it might not because resetRNG is called at init)
```
sbenthall, line 804: I wonder if this could be written in a more general way once rather than having more complexity added via inheritance.

This could be a time to think through how exogenous shocks get stored on the Agent object.

dolo, for example, puts all these into a collection names `exogenous`

HARK/ConsumptionSaving/ConsMedModel.py
```diff
@@ -714,7 +714,31 @@ def updatepLvlGrid(self):
...
+        ----------
+        None
+        
+        Returns
+        -------
+        None
+        '''
+        PersistentShockConsumerType.resetRNG(self)
+        
+        # Reset MedShkDstn if it exists (it might not because resetRNG is called at init)
```
sbenthall, line 736: Again: if all the exogenous shock distributions were contained in a single collector, then this method could be written more efficiently by looping over that collection.

sbenthall:
Great progress.

I have some notes on this which I don't think should block the PR but which I think would, ideally, be addressed or at least marked for improvement:

1. It looks like some of the changes made in this PR are purely formatting/whitespace issues, like when there are line breaks. I understand that there are often strong aesthetic feelings on this point. I also feel that objectively, these differences are trivial. I also believe that `black`, the code formatting tool that is now a pre-commit hook, is very opinionated about this, and that @MridulS merged a PR some time ago with the `black` changes made to the ConsumerSaving classes.

I just want to confirm that the changes in this PR are consistent with `black`, and not a reversion to a pre-black standard.

2. The `resetRNG` is in this PR inherited and modified by each subclass according to the name of its exogenous shocks. It would simplify the code and inheritance structure, as well as bring the software closer in line with Dolo, if the exogenous shocks were stored in some collection, such as a dictionary stored at `self.exogenous`, instead of added as attributes to the base Agent object. This would enable the method to be written generically like so:

```
for distribution in self.exogenous:
  self.exogenous[distribution].reset()
```

I wonder if now is the time to do this kind of refactoring, or if it should be slated for a different issue, given the urgency of this work.

mnwhite:
Yes, I have strong opinions about whitespace, and I don't know what black
says.

We can have an attribute like `exogenous` or `distributions` if you want,
but I think making the code work as intended is important.  However, your
code snippet won't work, as sometimes the elements of `IncomeDstn` are in
nested lists.  See resetRNG() for MarkovConsumerType.

sbenthall:
Good point about the code snippet not working.
Ideally, I'd like to see IncomeDstn implemented in such a way that it was never a nested list, but always a Distribution, even if it was a complex, multivariate distribution.
I'd see that as another step in this proposed refactoring.

I suppose I'll flag this as something to refactor later, unless you have an objection to this in principle.

As for the whitespace issues, while we have not passed an EEP about it, @MridulS and I at least are in agreement that following `black` conventions for whitespace is a good ida.
https://github.com/econ-ark/governance/issues/5

Since there is now the pre-commit hook installed ( see #703 ) I think that creates de facto pressure to standardize this way.

Running the `black` on some of the changed files in this PR, I see that indeed the tool does some reformatting.

I'm not certain but I believe some of these reformattings are ones you may have reverted in this PR.

It would be extremely silly if some of us used a reformatting tool as part of our contributions, and some of us reverted them back to a different format whenever they worked on the code.

mnwhite:
In this case, it's not a multivariate distribution, but instead one that
varies by age (t) and by a discrete state (n).  So there's no single
representation of the distribution within a period.

sbenthall:
Hmm. Then maybe using the Dolo idiom of a `Process` would be better here.

I'm not wedded to it being a Distribution subclass.

But I think it would be better to wrap this construct in an object rather than using a generic data type like a list because there is functionality over it that could be encapsulated by a method (for example, reseting all of them at once)

sbenthall:
Early bird catches the worm

PR #758, opened 2020-07-11
[WIP] Find value function crossings in dcegm.calcMultilineEnvelope

Mv77:
This PR augments dcegm.calcMultilineEnvelope by finding the exact coordinates of the kinks generated in the value function when computing its upper envelope. This lets us find the exact position of secondary kink points in DCEGM.

There are a couple of coding decisions on how to use these points that I would like to consult:

1. Should these points be incorporated to the m-grid _inside_ the calcMultilineEnvelope function or returned for other methods to do that and also use them for other purposes (e.g. plotting)?. My guess is do both.

2. What is your preference on how to handle the discontinuity of interpolating functions at these points? My idea is: if the crossing happens at some m, then add (m) and ( m+ np.finfo(float).eps) to the grid, with the corresponding values of c and v. Then generate the interpolating functions. Is there an approach that you would prefer?


llorracc:
@Mv77 could you make this an option rather than an unavoidable part of the code, and make a PR for a modification to the DemARK which illustrates the point.  (The issue is basically whether the time cost in finding those points is worth the increased accuracy; it would be nice to do some timing tests at some point, but those are not needed now).

Mv77:
The last commit makes computing the crossing points optional through an added boolean argument `findXings`. If `findXings == True`, the crossing points are computed and added to the gridpoints, and their `m` coordinate returned.

This feature is illustrated in https://github.com/econ-ark/DemARK/pull/126 . That PR won't work until this one is merged, but you can see the results in my version of the notebook here: https://github.com/Mv77/DemARK/blob/upperEnvelopeExample/notebooks/DCEGM-Upper-Envelope.ipynb (at the very end).

Mv77:
While checking for backwards compatibility of this PR, I noticed the endogenous-retirement REMARK is now broken. It seems to have to do with the change in how distributions are handled.

However, the dcegm test passed.

Mv77:
> 
> 
> While checking for backwards compatibility of this PR, I noticed the endogenous-retirement REMARK is now broken. It seems to have to do with the change in how distributions are handled.

The distributions were the trouble indeed. Fixed in https://github.com/econ-ark/REMARK/pull/70

Mv77:
The last commit adds a function for computing primary kinks. This is completely optional and done outside of `calcLogSumChoiceProbs`, taking its result as input.

PR #759, opened 2020-07-15
Small tweaks to CPM example file

Mv77:
@llorracc requested minor changes to the example_CPM file regarding the Merton-Samuelson exercise:
- To get rid of the CGM calibration and simply use defaults.
- To have a more informative legend in the plots.
- To use more extreme numbers of discretization points for the risky returns, e.g., 5 vs 200.

This PR introduces those changes.


Mv77:
Here is how the results are looking at the moment:
image
image

The gap between the exact and the no-discretization limits still looks quite large even for 200 points. I tried 500 without any notable change. @mnwhite says here https://github.com/econ-ark/REMARK/pull/68 that this is due to equiprobable approximations undershooting the variance of the continuous return distribution. Is the size of the gap consequent with this hypothesis?

Issue #760, opened 2020-07-16
grouping exogenous shocks into collection


sbenthall:
As per discussion in #757, try combining the exogenous shock distributions (and/or processes) into a collection and finessing the data types in the cases where they are age or state varying.

mnwhite:
I think we should talk about this (and AgentType stuff) right now on call.

sbenthall:
Fixed with #803

PR #762, opened 2020-07-20
Add proper Krusell-Smith model

mnwhite:
CDC has long wanted HARK to be able to solve the Krusell-Smith (1998) model verbatim; this PR adds that capability.  Previously, we had "fake KS", as our Markov ConsAggShock model didn't have serially correlated unemployment.

This PR includes new AgentType and Market subclasses in ConsAggShockModel.py, as the Krusell-Smith model is so specialized that trying to shoehorn trivial versions of existing HARK model objects would just overcomplicate things.  The one compromise with HARK is that the aggregate saving rule (regressing log A_t on log M_t conditional on the state of the economy) is still the equilibrium object, rather than a straight log K_t on log K_{t-1} like in the Krusell-Smith paper.  HARK's existing models use the "A on M" approach because we have aggregate risk other than the binary state of the macroeconomy, so K_{t-1} is not a sufficient statistic for the aggregate state in period t.

Example has been added to the ConsAggShock example file, replacing the old "fake KS" example.  Tests have also been added.  Default parameters are set to those in the KS paper, with the exception of LbrInd (\tilde{\ell}), whose value is not provided in the paper as far as I can tell.  Their results are nearly identically replicated.

This code does not handle the KS version of discount factor heterogeneity, as CDC did not reply to my query about whether he wanted that included.


alanlujan91:
I have ell tilde in my class notes as 0.3271, if that helps

mnwhite:
This does help, as I had hand-calibrated it to 0.325 to fit the KS results.
I was worried that they used 1/3 or 0.33 and there was still an error in
the code.

Given that all the other calibrated numbers are very rough and the entire
model has no scale, do you know where/why they used this specific number to
4 digits? Is it in the paper and I missed it?

alanlujan91:
I don't think the number is on the paper. As far as I remember, the source of this number is my prof. emailing one of the authors.

mnwhite:
Thank you very much for confirming this.

HARK/ConsumptionSaving/ConsAggShockModel.py
```diff
@@ -550,11 +550,326 @@ def getMrkvNow(self):  # This function exists to be overwritten in StickyE model
...
+        '''
+        Get each agent's consumption given their current state.'
+        '''
+        Mnow = self.Mnow*np.ones(self.AgentCount)
+        cNow = np.zeros(self.AgentCount)
+        
+        # This can be written more concisely, but is intentionally explicit for
+        # the sake of making it very easy to read
+        if self.MrkvNow == 0: # Bad macroeconomic conditions
+            these = np.logical_not(self.EmpNow) # Unemployed agents
```
sbenthall, line 844: maybe name this variable `unemployed_agents` instead of `these`. 
Save you having to explain it in comment :)

mnwhite: I suppose, but this is used throughout HARK for binary indexing arrays.

HARK/ConsumptionSaving/ConsAggShockModel.py
```diff
@@ -550,11 +550,326 @@ def getMrkvNow(self):  # This function exists to be overwritten in StickyE model
...
+        '''
+        params = init_KS_agents.copy()
+        params.update(kwds)
+
+        AgentType.__init__(self, pseudo_terminal=False, **params)
+
+        # Add consumer-type specific objects
+        self.time_vary = []
+        self.time_inv = ['DiscFac', 'CRRA',]
+        self.poststate_vars = ['aNow','EmpNow']
```
sbenthall, line 589: Not a critique of this code but a general question...

I see that the poststate_vars are listed explicitly here.
Would it be silly to list (a) the regular states, (b) the control variables, and (c) the shocks in lists like these?
I mean, in a future implementation.

That would make it easier to do things like #760 and #761

HARK/ConsumptionSaving/ConsAggShockModel.py
```diff
@@ -550,11 +550,326 @@ def getMrkvNow(self):  # This function exists to be overwritten in StickyE model
...
+    def preComputeArrays(self):
+        '''
+        Construct the attributes X, which will be used by the one period solver.
+        '''
+        # Get array sizes
+        aCount = self.aGrid.size
+        Mcount = self.Mgrid.size
+        
+        # Make tiled array of end-of-period idiosyncratic assets (order: a, M, s, s')
+        aNow_tiled = np.tile(np.reshape(self.aGrid, [aCount, 1, 1, 1]), [1, Mcount, 4, 4]) # shape (aCount, Mcount, 4, 4)
```
sbenthall, line 663: In general, I think the code formatting of this PR is really nice.
I was going to ask if you used `black` for it?

I'm surprised by this long line (in terms of number of characters) though.

HARK/ConsumptionSaving/ConsAggShockModel.py
mnwhite: That can be done.  They're listed here and in all HARK models because initializeSim() initializes them.

HARK/ConsumptionSaving/ConsAggShockModel.py
mnwhite: No, I didn't use black.  All of my comments with the shape of arrays can be removed (will do that now).  I put those in while explaining the code the Mridul.

mnwhite:
Ack, just caught that I forgot to put pre-computed arrays into the solveKrusellSmith docstring.

HARK/ConsumptionSaving/ConsAggShockModel.py
sbenthall: fair enough. just a suggestion.

sbenthall:
I've made a few minor comments.
Otherwise, this looks good to me.
I'm probably not the best person to judge the economics of it, but I'd be happy to merge if there's no objections, after Matt gets in his "ack!' fix.

HARK/ConsumptionSaving/ConsAggShockModel.py
sbenthall: Cool. I think I can demo how this would look in a PR for the next weekly meeting.

HARK/ConsumptionSaving/ConsAggShockModel.py
```diff
@@ -550,11 +550,325 @@ def getMrkvNow(self):  # This function exists to be overwritten in StickyE model
...
+        self.preComputeArrays()
+        
+    def update(self):
+        '''
+        Construct objects used during solution from primitive parameters.
+        '''
+        self.makeGrid()
+        self.updateSolutionTerminal()
+        
+    def getEconomyData(self, Economy):
```
sbenthall, line 604: Should be lower case -- see how the GitHub syntax highlighting misformats this as a class name because it is capitalized.

I see this is common to other uses of getEconomyData though.

HARK/ConsumptionSaving/ConsAggShockModel.py
```diff
@@ -550,11 +550,325 @@ def getMrkvNow(self):  # This function exists to be overwritten in StickyE model
...
+        self.CapShare = Economy.CapShare             # Capital's share of production
+        self.LbrInd = Economy.LbrInd                 # Idiosyncratic labor supply (when employed)
+        self.UrateB = Economy.UrateB                 # Unemployment rate in bad state
+        self.UrateG = Economy.UrateG                 # Unemployment rate in good state
+        self.ProdB = Economy.ProdB                   # Total factor productivity in bad state
+        self.ProdG = Economy.ProdG                   # Total factor productivity in good state
+        self.MrkvIndArray = Economy.MrkvIndArray     # Transition probabilities among discrete states
+        self.MrkvAggArray = Economy.MrkvArray        # Transition probabilities among aggregate discrete states
+        self.addToTimeInv('Mgrid', 'AFunc', 'DeprFac', 'CapShare', 'UrateB', 'LbrInd',\
+                          'UrateG', 'ProdB', 'ProdG', 'MrkvIndArray', 'MrkvAggArray')
```
sbenthall, line 632: See #764 . If those aspects of the economy that needed to be brought into the Agent were available in a list (as most of them are here), then it would be possible to write this method generically as a loop over those values, rather than writing out each one individually.

This is common across all uses of getEconomyData

mnwhite:
But not all of them are just "take this attribute of that and make it an
attribute of self".  Some of them change names or do a very small bit of
coding.  Yes, that can be done in the lines below.

Change it as you see fit.

sbenthall:
Made #766 for that issue.
I'll work on it after this PR is in so as not to block it.

sbenthall:
Is this PR ready to merge, from your perspective, @mnwhite ?

mnwhite:
Yes, I think so.

mnwhite:
Ah, the one thing left is to make the ipynb example match the py.  There's
an automated thing for that?

sbenthall:
! Oh, perhaps I merged too soon !

Yes, I think what you are looking for is `jupytext`

https://jupytext.readthedocs.io/en/latest/using-cli.html

mnwhite:
We can just do it in another branch.

PR #765, opened 2020-07-21
light touch Market class refactoring

sbenthall:
CDC has asked me to familiarize myself with the Market code because of a certain project.

In this PR, I'll try to do some light maintenance based on previously established design principles.

Some considerations:
 - A performance benefit for namespaces/dictionaries for state storage rather than as object attributes (see #674)
 - By packaging alike variable together, it makes it easier to write generalized functions (see #760 #761)
 - The more mathematical values can be represented as key/value pairs in a dict-like data structure, the less conflict there is between specialized mathematics formatting desiderata and Python style conventions (see https://github.com/econ-ark/governance/issues/5 )

The current step in this PR, meant to be illustrative, is moving the storage of initial and state values for `sow_vars` into dictionaries. The trajectory of this work is to make similar adjustments for the `reap_vars`, etc. The Market class is a nice place to do these changes because the variables are already categorized in the class constructor. (I'd recommend similar changes to the Agent classes, but these are not yet written in as abstract a way.)

The current PR passes the `test_ConsAggShockModel.py` tests. I'd like to note that this is without changes to e.g. the `SmallOpenEconomy` that would be necessary to retain functionality. This indicates a gap in the current test suite. If this PR's direction gets approval, I would write tests for those other classes as part of this PR.


sbenthall:
Another reason to store Market state in this way is that this kind of representation is needed by the `mill` method anyway. Currently, this is done via awkward ad hoc construction. The trajectory of this PR is to have the Market natively store the `reap` and `const` variable values in dictionaries, making this ad hoc construction unnecessary.

https://github.com/econ-ark/HARK/blob/master/HARK/core.py#L1058-L1065

HARK/ConsumptionSaving/ConsAggShockModel.py
```diff
@@ -1678,12 +1680,12 @@ def update(self):
         self.KtoLnow_init = self.kSS
```
sbenthall, line 1680: did I miss one here? i think so

HARK/core.py
```diff
@@ -1126,7 +1135,10 @@ def store(self):
         none
         '''
         for var_name in self.track_vars:
```
sbenthall, line 1137: This gets more complex with more namespaces for variables.

HARK/core.py
```diff
@@ -1058,20 +1073,15 @@ def mill(self):
...
 
         # Run the millRule and store its output in self
         product = self.millRule(**mill_dict)
-        for j in range(len(self.sow_vars)):
-            this_var = self.sow_vars[j]
-            this_product = getattr(product, this_var)
-            setattr(self, this_var, this_product)
+
+        for sow_var in self.sow_state:
+            this_product = getattr(product, sow_var)
```
sbenthall, line 1083: This simplifies if `millRule` returns a dict-like: `self.sow_state.update(product)`

HARK/ConsumptionSaving/ConsAggShockModel.py
```diff
@@ -2247,14 +2264,14 @@ def update(self):
...
-        self.Mnow_init = self.MSS
-        self.Aprev_init = self.KSS
-        self.Rnow_init = self.RSS
-        self.Wnow_init = self.WSS
+        self.sow_init['KtoLnow'] = self.KtoLSS
+        self.sow_init['Mnow'] = self.MSS
+        self.sow_init['Aprev'] = self.KSS
+        self.sow_init['Rnow'] = self.RSS
+        self.sow_init['Wnow'] = self.WSS
         self.PermShkAggNow_init = 1.0
```
sbenthall, line 2272: Missed these?

HARK/ConsumptionSaving/ConsAggShockModel.py
```diff
@@ -1825,6 +1847,15 @@ def __init__(self,
...
                  act_T=1200,
+                 sow_vars=['MaggNow',
+                           'AaggNow',
+                           'RfreeNow',
+                           'wRteNow',
+                           'PermShkAggNow',
+                           'TranShkAggNow',
+                           'KtoLnow',
+                           'MrkvNow' # This one is new
+                 ],
```
sbenthall, line 1858: When all the sow_vars are "Now", "Now" can be left out of the name...

HARK/ConsumptionSaving/ConsAggShockModel.py
```diff
@@ -2455,81 +2526,6 @@ def calcAFunc(self, Mnow, Aprev):
         return AggShocksDynamicRule(AFunc_list)
 
 
-class CobbDouglasAggVars(HARKobject):
```
sbenthall, line 2458: replaced these container classes with tuples same size/order as sow_vars

HARK/ConsumptionSaving/tests/test_ConsAggShockModel.py
```diff
@@ -48,6 +48,9 @@ def test_macro(self):
         self.assertAlmostEqual(self.economy.AFunc.slope,
                                1.116259456228145)
 
+        self.assertAlmostEqual(self.economy.history['MaggNow'][10],
```
sbenthall, line 51: These tests are added in #773, which should be merged first

HARK/core.py
sbenthall: chose for `millRule` to return a tuple

sbenthall: (closer to Dolo, supposed to be good for numba)

HARK/ConsumptionSaving/ConsAggShockModel.py
sbenthall: No, not missing because these are not sow_vars in KS

HARK/ConsumptionSaving/ConsAggShockModel.py
sbenthall: But I don't dare.

HARK/core.py
sbenthall: fixed

sbenthall:
This PR is now ready for review.

llorracc:
Can use regular dicts instead of OrderedDict for Python 3.7+

sbenthall:
0.10.6 supports Python 3.6:
https://github.com/econ-ark/HARK/blob/master/setup.cfg#L42

I think in our meeting today we agreed 0.10.7 would drop Python 3.6 support

sbenthall:
OrderedDict replaced with dict in recent commit

sbenthall:
I expect this to be merged AFTER the 0.10.7 release, since there seems to be a feature freeze there.

PR #772, opened 2020-07-23
remove deprecated Solution code, use HARKobject in the codebase

MridulS:
fixes https://github.com/econ-ark/HARK/pull/772


MridulS:
@mnwhite Can you have a look here and merge this? There were a couple of places where the `ModelSolution` Class inherited from `Solution`, I have changed it to inherit `HARKobject` now.

Issue #612, opened 2020-04-07
Equivalency between 2 models/agents. 


MridulS:
Should HARK have the `__eq__` method to compare 2 HARK objects?

For example

```
example_1 = PerfForesightConsumerType(**PF_params)
example_2 = PerfForesightConsumerType(**PF_params)
```

`example_1 == example_2` should be True (it will give it False) but currently HARK doesn't have the ability to test this.

sbenthall:
That sounds nice.

A couple thoughts:
 - The `distanceMetric` functionality seems to do some of this. That functionality is old code. It could probably be reimplemented in a more pythonic way. Maybe that would be connected to this issue.
 - Would an unsolved and a solved AgentType be equivalent? What if a simulation had been ran, with different parameters? The meaning of 'equivalence' might be more intuitive if some of the functionalities currently handled in AgentType were disentangled, see #495

MridulS:
> Would an unsolved and a solved AgentType be equivalent? What if a simulation had been ran, with different parameters? The meaning of 'equivalence' might be more intuitive if some of the functionalities currently handled in AgentType were disentangled, see #495

Different parameters should mean the agent are not equivalent. There isn't any way to get all the parameters of an agent right? something like `AgentType.parameters`

mnwhite:
distanceMetric is used for comparing solutions, to see if it has converged
on successive iterations.  I'm not sure there's a use case for comparing
instances of AgentType subclasses for equality.

sbenthall:
> There isn't any way to get all the parameters of an agent right? something like `AgentType.parameters`

I don't believe there is.
But this is high on my refactoring wishlist.

It's a little complicated to do this... see #492

sbenthall:
#760 #761 are related

sbenthall:
See also #660

sbenthall:
> distanceMetric is used for comparing solutions, to see if it has converged on successive iterations. I'm not sure there's a use case for comparing instances of AgentType subclasses for equality.

This makes sense to me.

I have wondered for a while why AgentType inherits from HARKObject, since there is no use case for using its distanceMetric function.

I think what it's using currently is `assignParameters`.

Does `assignParameters` ever get used on a solution object? If not, then I think it would be safe to split HARKObject into two classes.

It would be safe to do that anyway, since the two classes can be combined with multiple inheritance when necessary.

PR #774, opened 2020-07-27
adding PerfForesightConsumerTypeFast 

alanlujan91:

Would like suggestions on the boxes not checked. 

This PR creates `ConsIndShockFastModel` which for now only contains `PerfForesightFastConsumerType`, a `numba` implementation of `PerfForesightConsumerType`. 

~~I have also added basic `__eq__` for comparison of models and solutions as in #612.~~

## `ConsIndShockFastModel`

The goal of this PR is to create a mirror to `ConsIndShockModel.py`, but that implements faster solutions using `numba` just-in-time compiling and optimization.

### `PerfForesightSolution`

The first step is to create a solution object called `PerfForesightSolution` that contains only python primitives and `numba` arrays. This object takes the place of `ConsumerSolution`, which creates additional HARK objects and can lead to unnecessary time overhead in the solution iteration. The `distance_criteria` for this solution is `["cNrm", "mNrm"]`, which is identical to the `distance_criteria` for `ConsumerSolution` but more straight forward. This `ConsumerSolution` replacement is faster because instead of having to create HARK objects such as `cFunc`, `vFunc`, etc. as `LinearInterpolator`s for every solution iteration, the solution method or algorithm only really needs the primitives that compose these objects. Below I will describe in more detail how this object is used in the solver. 

### `ConsPerfForesightFastSolver`

The next step is to create a solver that replaces `ConsPerfForesightSolver`, aptly called `ConsPerfForesightFastSolver`. The object uses a `numba`-optimized helper function called `solveConsPerfForesightNumba`. In this implementation, the outputs of this helper method are packaged into a `PerfForesightSolution` object instead of a `ConsumerSolution` object. As I mentioned, this avoids the needless creation of other HARK objects in the iteration, leaving the creation of a proper HARK solution to the end of the iteration. An additional feature that has been removed from the solution iteration is the searching of steady-state normalized market resources with `addSSmNrm()`. As the iteration does not depend on this result as an intermediate input, this can be safely moved to the end of the solution method, or rather to `postSolve()`. 

### `PerfForesightFastConsumerType`

Finally, I create an agent that can use the above solution method called `PerfForesightFastConsumerType`. Since now the solution iteration uses `PerfForesightSolution`, I first override `solution_terminal_ = PerfForesightSolution()`. Then, on `__init__`, I override `solveOnePeriod` to `ConsPerfForesightFastSolver`. Since I am overriding the solution_terminal_, I might still need a `ConsumerSolution` terminal in the case of a non-infinite problem. So, I override the method `updateSolutionTerminal()` to save an internal `ConsumerSolution` version of the `solution_terminal`. The final step is to implement a `postSolve()` method to wrap every `PerfForesightSolution` parameter into a HARK object and to run `addSSmNrmNumba()` only on the final solution. As I had mentioned, there is no need to create these objects at every iteration step of the solution, and it is thus more efficient to leave them to `postSolve()`. Now, ideally, the output of `solve()` is the same between the standard HARK approach and the fast approach, a list of `ConsumerSolution` objects with `cFunc`, `vFunc`, etc. as HARK `LinearInterpolator`s.


HARK/ConsumptionSaving/ConsIndShockFastModel.py
```diff
@@ -0,0 +1,427 @@
...
+        self.MPCmax = MPCmax
+
+
+# =====================================================================
+# === Classes and functions that solve consumption-saving models ===
+# =====================================================================
+
+
+@njit(cache=True)
+def searchSSfunc(m, Rfree, PermGroFac, mNrm, cNrm, ExIncNext):
```
MridulS, line 130: For code-style and documentation needs, can you make these independent helper functions start with `_`.

Example:
``` python
def _searchSSfunc(.........):

def _addSSmNrmNumba(........):

def _solveConsPerfForesightNumba(........):

```
By doing this we can make sure we don't pollute the local namespace, documentation on ReadTheDocs and functions that users won't be accessing directly should start with `_`.

HARK/ConsumptionSaving/ConsIndShockModel.py
```diff
@@ -99,20 +138,33 @@ def __init__(self, cFunc=None, vFunc=None,
...
-        self.MPCmin       = MPCmin
-        self.MPCmax       = MPCmax
+        self.mNrmMin = mNrmMin
+        self.hNrm = hNrm
+        self.MPCmin = MPCmin
+        self.MPCmax = MPCmax
 
-    def appendSolution(self,new_solution):
-        '''
+    def __eq__(self, solution: HARKobject) -> bool:
```
MridulS, line 153: This should ideally be present in the parent class `HARKobject` (similarly for `AgentType`).

MridulS:
> Would like suggestions on the boxes not checked.

For the changelog one, let's wait till we make a 0.10.7 release first, then you can add this to the changelog for 0.10.8. You would need to edit this file https://github.com/econ-ark/HARK/blob/master/Documentation/CHANGELOG.md

For the documentation one, as you have added new functionality and files you also need to add this to the documentation folder so that the documentation site (https://hark.readthedocs.io/en/latest/) is able to scrap the new model class.
Go to `Documentation/reference/ConsumptionSaving` and create a new file `ConsIndShockFastModel.rst` with the content

```
ConsIndShockFastModel
----------------------------

.. automodule:: HARK.ConsumptionSaving.ConsIndShockFastModel
   :members:
   :undoc-members:
   :show-inheritance:
```

Sphinx will do rest of the work to scrap through the class docstrings and generate a documentation page for it.

HARK/ConsumptionSaving/ConsIndShockFastModel.py
```diff
@@ -0,0 +1,427 @@
...
+   2) A consumption-savings model with risk over transitory and permanent income shocks. #todo
+   3) The model described in (2), with an interest rate for debt that differs
+      from the interest rate for savings. #todo
+
+See NARK https://HARK.githhub.io/Documentation/NARK for information on variable naming conventions.
+See HARK documentation for mathematical descriptions of the models being solved.
+"""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
```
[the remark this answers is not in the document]
alanlujan91: ok, I will remove from ConsIndShockModel.py and ConsIndShockFastModel.py

I see that `__future__` imports are peppered throughout the codebase, might we want to remove these as well?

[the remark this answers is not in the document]
MridulS: yeah go for it, but preferably keep them in another PR :)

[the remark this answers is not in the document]
alanlujan91: will do! suggestions on my PR and commit patterns are appreciated! 

is `builtins` also for compatibility? should I remove those?

HARK/ConsumptionSaving/ConsIndShockModel.py
alanlujan91: This makes sense, although I put them in PerfForesight Model and Solution since these implementations might not be ready for generic models and generic solutions. Do you think I should proceed to moving them to HARKobject and AgentType as is? or maybe generic `__eq__` comparison should be on another PR?

MridulS: Yeah, let's do that. Can you remove the __eq__ tests from this PR and move it to another branch?

alanlujan91: removed

llorracc:
My preference for naming is to add extra things to the beginning or the end of the name of an existing object; so, PerfForesightConsumerTypeFast

HARK/ConsumptionSaving/ConsIndShockModel.py
alanlujan91: I have removed everything related to `__eq__` and will do this in own PR

alanlujan91:
I've removed everything relating to `__eq__` to move it to its own PR later on. @MridulS I think this is ready to review and merge as only adding PerfForesightConsumerTypeFast

PR #779, opened 2020-07-28
fix distribution seed in tractablebufferstock

MridulS:


sbenthall:
I have to defer to @mnwhite on this.

In some of the ConsumptionSaving models, there was a choice to set seeds in an unconventional way.
I could never tell if this was a deliberate design. (I would like to think it _isn't_, and would otherwise merge this PR. It looks good to me.)

Issue #457, opened 2019-12-12
Generalise unpackcFunc, make a general unpacking function


MridulS:

mnwhite:
Strongly endorse this.  This is a good task for someone new to HARK-- the new method is like 3 lines long.

PR #784, opened 2020-07-30
generalised unpack function after a model is solved

MridulS:
fixes https://github.com/econ-ark/HARK/issues/457

- [N/A ] Tests for new functionality/models or Tests to reproduce the bug-fix in code.


sbenthall:
Is the idea that this is already covered by existing tests?

MridulS:
nope, this is supposed to fail for the time being. I want to discuss this in the call today.

Currently we use `unpackcFunc` and this PR introduces `unpackFunc` (notice the missing 'c') so we would need to update the code examples where we use `model.unpackcFunc()` with something like `model.unpackFunc('cFunc')` to have a generalised unpacking function. `model.unpackFunc('cFunc')`  does look clunky so maybe we can have something like `model.unpack('cFunc')`

llorracc:
@MridulS 

Thanks for fixing then merging this.

PR #789, opened 2020-08-06
Fix checkRIC typo

Mv77:
There is a typo in the checkRIC() method of the perfect foresight agent, which says that the condition is satisfied when it fails.

This PR fixes it.


mnwhite:
Will merge shortly.

PR #792, opened 2020-08-06
Tex setup fix - windows problems with paths (?)

Mv77:
`setup_latex_env_notebook` creates the path of 'latexdefs.tex', a file of symbol definitions to be read when rendering figures.

However, paths in Python use '\\\\', which is unrecognized by (at least my version of) LaTeX. Substituting '\\\\' for '/' fixes the issue.


MridulS:
> However, paths in Python use '\\', which is unrecognized by (at least my version of) LaTeX. Substituting '\\' for '/' fixes the issue.

Is this on windows?

Mv77:
Yes, that problem came up on my Windows machine.

llorracc:
Should add a test for whether running and windows and change separator accordingly

sbenthall:
The problem is that using string manipulation to specify a path is not cross-platform, for the reasons you mentioned.

However, Python has internal cross-platform path construction tools.

https://docs.python.org/3/library/os.path.html

Try building the path with `os.path.join`

Mv77:
I'm working on this. The current version results in LaTeX being passed things like this
image
in Windows, so it breaks.

Maybe my separator from the point of view of python is '\\' but then when it writes the tex files it interprets that as an escaped '\' and then latex sees what should be a separator as a command... I have to concede this to @llorracc as a point against Windows.

Mv77:
The last commit fixes the issue.

I tested the resulting 'make' tool on Windows by running [the BST notebook](https://github.com/llorracc/BufferStockTheory/blob/master/Code/Python/BufferStockTheory-Problems-and-Solutions-Source.ipynb). It completed without issues.

Is there a way to make sure I did not break the functionality in other OS? Do the automated Travis test check the `make_figs` function?

Mv77:
Latexdefs needs to be in the same place as the notebook anyway. So just try to get rid of the cwd solution altogether.

PR #799, opened 2020-08-17
tests in the Lognormal class to make sure mu and sigma have the same …

sbenthall:
This PR adds a couple exception handlers to the Lognormal class to check to make sure that if mu or sigma is a list, they are both lists, and are of the same length.

### DIFF
--- a/HARK/distribution.py
+++ b/HARK/distribution.py
@@ -57,6 +57,20 @@ class Lognormal(Distribution):
         # Set up the RNG
         super().__init__(seed)
 
+        if isinstance(self.mu, list) != isinstance(self.sigma, list):
+            raise Exception(
+                "mu and sigma must be of same types. They are %s, %s" % (
+                    type(self.mu), type(self.sigma)
+                    ))
+
+        if isinstance(self.mu, list):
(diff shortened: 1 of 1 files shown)


HARK/distribution.py
```diff
@@ -57,6 +57,20 @@ def __init__(self, mu = 0.0, sigma = 1.0, seed = 0):
...
+                    type(self.mu), type(self.sigma)
+                    ))
+
+        if isinstance(self.mu, list):
+            if len(list(self.mu)) != len(list(self.sigma)):
+                raise Exception(
+                    "mu and sigma must be of same length, are %d, %d" % (
+                        len(list(self.mu)), len(list(self.sigma))
+                    ))
+
```
MridulS, line 72: what should happen if there are not lists?

llorracc: Like, if they are scalars?  Then they should be converted into length-1 lists, I guess ...

llorracc: The internal representation should be numpy arrays, not lists.  If the user provides a one-dimensional input, mu and sigma, those should be converted to two 1-element np arrays.

These topics should be moved to a different issue.  @MridulS or @sbenthall will create the new issue

Issue #809, opened 2020-08-20
Need instructions for how to get Black running locally


llorracc:
@alanlujan91 will review the docs that Mridul writes

MridulS:
It's already available at https://github.com/econ-ark/HARK/blob/master/Documentation/contributing/CONTRIBUTING.md#stylistic-guidelines

PR #797, opened 2020-08-13
applying black to HARK

sbenthall:
This PR applies `black` to reformat all the code in the main HARK module.

- [x] Update CHANGELOG.md with major/minor changes.

PR #813, opened 2020-08-24
[WIP] Add a Github action to run pre commit on PRs

MridulS:


MridulS:
Created a Github action which makes https://github.com/econ-ark/HARK/issues/809 redundant as black and any future pre commit tools will run using GitHub Actions and it will create a new commit in the PR with the message "pre-commit fixes".

This should be only merged in if https://github.com/econ-ark/HARK/pull/797 is merged.

This action ensures the code that is merged into master always follows the black style guidelines, even if the contributor hasn't installed it locally.

PR #819, opened 2020-08-25
[WIP] Add education choice model

DuanYiran1990:
A consumer type with idiosyncratic discount factor, age-path of education effort cost, direct monetary cost
    of education, with non-separable utility function (derive utility from consumption and experience disutility
    from periods spent receiving education). In each discrete time period, an agent makes a continuous decision over
    how much of their market resources to consume and how much to retain in a risk free asset; If they received
    education in the prior period, they also choose how much costly effort to exert to remain in school.


mnwhite:
I have edited the title of the PR.  Please provide a description of your work in the initial comments so that others can understand what you're contributing.

mnwhite:
The file should not have [WIP] in front of its name. That's just a tag for the pull request, not the file.

alanlujan91:
@mnwhite is there  a write up for the problem this is trying to solve?

mnwhite:
Yes, I can email you notes later.

PR #821, opened 2020-08-25
Add unemployment search model with non-separable disutility of search

fangli-DX3906:
This PR adds non-separable disutility of search to unemployment search model. There is only one unemployment status. Agent makes a decision about how much of his resources to consume and how much to retain in an interest bearing asset; when he is unemployed, he also chooses his probability of becoming re-employed s_t, but higher probabilities induce greater disutility.
When employed, his income is subject to permanent and transitory income shocks and an exogenous probability of becoming unemployed (job separation rate). When unemployed, the agent receives unemployment benefits proportional to his permanent productivity.


mnwhite:
It looks like you deleted the prior ConsLaborModel.py; please don't do that.  Your PR should *modify* ConsLaborModel.py, not delete it and replace it with the similarly name ConLaborModel.py.

Also, you don't need to close the PR.  Just continue making commits and push them.  They will appear here.

Issue #678, opened 2020-05-07
automated test for finite horizon cyclical model


sbenthall:
include an automated test for the finite horizon cyclical case
derived from an example perhaps

PR #835, opened 2020-09-14
Time varying solver

Mv77:
While working on #826, I noticed that time-varying `solveOnePeriod` methods, passed in a list, are read going _forwards_ in the list. All other parameters in `timeVary` are read backwards from the list, so I believe this is a bug.

(Alternatively one could create the ``solveOnePeriod` list going backwards in time, but I don't think that's the desired behavior)

This PR fixes the issue.


mnwhite:
Yikes, let me take a look at this.  That would be a pretty fundamental error.

sbenthall:
As @mnwhite is very busy, I've been assigned to review this PR.

My first observation is that the code change has not modified any automated tests, and has not caused any automated tests to fail.

That implies that the functionality effected by the change is not currently covered by an automated test.

Therefore, I do not think this PR is ready to be merged.

@Mv77 would you be able to write an automated test for the functionality being addressed?
I am happy to meet and work with you to introduce you to the automated testing framework if that would be helpful.

sbenthall (changes requested):
Needs automated test of the functionality being changed.

Doing this will clarify whether and to what extent the original software behavior included a bug.

Mv77:
> 
> 
> As @mnwhite is very busy, I've been assigned to review this PR.
> 
> My first observation is that the code change has not modified any automated tests, and has not caused any automated tests to fail.
> 
> That implies that the functionality effected by the change is not currently covered by an automated test.
> 
> Therefore, I do not think this PR is ready to be merged.
> 
> @Mv77 would you be able to write an automated test for the functionality being addressed?
> I am happy to meet and work with you to introduce you to the automated testing framework if that would be helpful.

Yes, the reason why no behavior is modified is that (I think) no example or test in HARK uses the time-varying solver functionality.

Sure, I could take a stab at writing a test. We'd need to talk about how the test would look like however, since none of the agent types currently in HARK lends itself to this functionality very easily.

sbenthall:
It looks like there is already an issue for this here: #678 If this problem interests you, I suggest you claim it. May we add you as a collaborator on the repository?

Perhaps @mnwhite and @llorracc  have some downstream code that demonstrates its usefulness that could be converted into an example and test combination.

One the other hand, I believe this functionality is also being staged for a redesign.
The discussion in #798 about periods and frames pertains mainly to model definitions and includes a discussion of "cyclic" problems. There has been a rather large amount of difference of opinion about the best way to handle 'cycles', 'lifecycles', etc.

One part of that redesign which I believe is not yet fully specified is how to organize methods for backwards induction at each frame.

See also #495 for background discussion.

mnwhite:
Writing a test for this would be rather cumbersome: he would have to add an
entire new model to HARK. Mateo is doing this anyway, which is how he found
this error. This PR is independent of that model.

mnwhite:
I just verified: It's a one line change. He fixed an indexing issue that came up when timeFlip() was removed from the code.  The model Mateo is working on will use this feature, but I don't think we need a PR where 5 characters are changed to fix an index, and then 2000 lines of code are added to show that this is correct.

sbenthall:
Sounds like you're the reviewer now, @mnwhite .
I defer to you.

PR #836, opened 2020-09-14
namespace for states; transition function for #761

sbenthall:
Addresses #761 

Still working on this.


sbenthall:
Tests for this PR are passing locally.

The implementation is still a bit rough, especially in light of the dealing with "poststates", which are slated to be substantially revised (see #798).

But I think it is at a state now where it makes sense to request a review from @mnwhite .

sbenthall:
Next steps for this PR:
 - replace "post state" storage with a step at the beginning/end of a period which stores values from `states_now` in `states_prev`, and put None/NaN values in all the `states_now`
 - include a configurable `ante_vars` with functions for initializing the state at time = 0.

sbenthall:
Ah, working on this reminded me of the other reason for the poststate designation.

The poststates notionally need to be computed after a control variable is determined via a decision rule.

I believe that this wrinkle will get smoothed over once we use the topological order of model variables, or some user-defined framing, to determine execution order.

In the short term, I'm going to have to keep using the somewhat vague and overloaded 'poststate' category:
 - getPoststates will calculate state values that depend on controls....
 - ... but store these values in state_now momentarily...
 - ... before transferring them to state_prev

sbenthall:
I have almost all the bugs worked out in the new implementation (i.e., most tests are passing) but I have a sticky issue I need feedback from @mnwhite about.

It has to do with how the AgentType classes, now organized with `state_now` and `state_prev` namespaces, interact with the Market class.

Some things I notice:
 - While the market `solve()` method uses agent simulation as part of the `makeHistory()` step, I don't believe `initializeSim` on the agents is ever called. This makes for a different path for seeding the antecedent state (i.e., the `state_prev` values needed to simulate a period.)
 - I guess the pushing and pulling of state from the market to the agent objects is done in the `sow()` and `reap()` methods.
   - these used to work by setting agent attributes
   - now, we need to make the decision of whether they `sow` to `state_prev` and `reap` from `state_now`, or what.
   - It looks like the `marketAction` command `simulate`s one period, which would cycle `state_prev` on time period--so maybe the market reads from and writes to `state_prev`?
      - does this use the same mortality mechanics as a regular simulation?

Thanks in advance for any thoughts on this.

sbenthall:
Tests are now passing.

There remains a design issue which has not be resolved in a satisfactory way yet.

Model classes currently use a combination of explicitly defined state variables and inherited state variables.

In the current pattern, the lists of 'state' variables are inherited in such a way that many downstream classes have dummy states used by ancestor's models, but unused themselves. This is messy.

There is also a problem that in some places, the 'state variable' is defined by assignment to two dictionaries (`state_now` and `state_prev`) in the model constructors. This is untidy and can lead to bugs in future models.

sbenthall:
There seem to be some inconsistencies across the model classes in how inheritance is done. This makes it harder to have a consistent design for these things.

I'm wondering if there's a specific reason why the constructor for `AggShockConsumerType` calls the `AgentType` constructor, rather than its superclass `IndShockConsumerType`'s constructor. It takes several steps to reproduce some of the aspects of IndShockConsumerType after the call to the `AgentType` constructor.

sbenthall:
Latest commit cleans up some of the `state_var` inheritance issues.

However, locally this change breaks the Fast models developed by @alanlujan91

@alanlujan91 do you have any insight into why this may be happening? I have not touched the Fast classes directly in this PR. The main thing the most recent change does to the core classes is create a `state_vars` (no underscore) variable on the `AgentType` class and has this variable be modified directly in subclasses like `ConsIndShockModel`

sbenthall:
Hmm. I see tests are passing online. So the issue with the Fast models must be a local issue.

In that case, I request a review from @mnwhite

alanlujan91:
> Latest commit cleans up some of the `state_var` inheritance issues.
> 
> However, locally this change breaks the Fast models developed by @alanlujan91
> 
> @alanlujan91 do you have any insight into why this may be happening? I have not touched the Fast classes directly in this PR. The main thing the most recent change does to the core classes is create a `state_vars` (no underscore) variable on the `AgentType` class and has this variable be modified directly in subclasses like `ConsIndShockModel`

@sbenthall if this issue persists, I can take a look later this week.

mnwhite:
Quick reactions:

1) It looks like getStates() now returns the new state variables, but doesn't actually record them in a history.  AgentType.simOnePeriod() also doesn't try to collect any output from getStates.

2) Did you mean to rename getStates() to transition()?  ConsGenIncProcessModel has this renamed, but it's not changed in core; not sure if this is WIP.

3) The fact that time advances and everything in state_now should be copied to state_prev should be the *first* thing that happens in the period, not the last thing.  As written, any time the user simulates and then looks at state_now, she would expect to see the current values of variables at the end of the simulation... but instead she sees empty arrays.

4) That dictionary move/copy should be part of simOnePeriod itself, not something the user has to ask for by calling AgentType.getPostStates() (or any other method).  It's part of the machinery.

sbenthall:
Thanks for this, @mnwhite 

> 1. It looks like getStates() now returns the new state variables, but doesn't actually record them in a history.  AgentType.simOnePeriod() also doesn't try to collect any output from getStates.

I believe getStates(), getControls() etc are being phased out, but not yet in this PR.

In `core`, getStates() wraps the `transition()` method, which is modeled on Dolo's `transition` functions. I was going to split this up into a frame-specific transition function in a future step. getStates (for now) stores the results of the transition in memory.
https://github.com/econ-ark/HARK/pull/836/files#diff-250396a11470a27af0e8cc6a3474b16cR660-R663

Corresponding with Dolo, the `transition` method returns the values of the changed variables.

In some cases because of the irregularity between different models, I was unable to liquidate the getStates variable completely. Separately, it may be that I have missed some revisions to the new style.

I'll check over this again.

> 3\. The fact that time advances and everything in state_now should be copied to state_prev should be the _first_ thing that happens in the period, not the last thing

> 4\. That dictionary move/copy should be part of simOnePeriod itself, not something the user has to ask for by calling AgentType.getPostStates() (or any other method).  It's part of the machinery.

Ah, these are good points. I will make this change. Thanks!

mnwhite:
Crap, I missed that AgentType.getStates calls AgentType.transition.  My bad.

sbenthall:
As a consequence of this, I will need to change `simBirth` so that it populates `state_now` instead of `state_prev`.

The upshot of this is that agents will be born into t = 0 without the normal transition logic.
This is a slightly different flavor than before, where `simBirth` populated `poststate_vars` which was a kind of "yesterday".

sbenthall:
Ah, hmm. But that has some friction with `getMortality`, which creates `newborns` with different t-1 state?

mnwhite:
getMortality should just call simDeath and simBirth, I think.  getMortality
can be called after state_now is moved to state_prev. simBirth should
overwrite state_prev entries of everyone who is flagged in who_dies,
creating a new "previous state" for newborns that describes how they come
into the world; any state variables that aren't specified in simBirth
should be left as NaN.

sbenthall:
This is how it worked before.

However, since simBirth is also called by initializeSim, then those birth conditions get erased if the state moves from _now to _prev at the beginning of the period.

sbenthall:
I think I need to put the moving for state from _now to _prev in between getMortality and the model transition.

mnwhite:
Just put a conditional on getMortality then:

`if self.t_sim > 0:`
`    self.getMortality()`

Then it gets skipped only if this is the very first simulated period in
history.

sbenthall:
That is not as elegant.

sbenthall:
Oh dear. Now the shock logic and the state logic are quite divergent.

I don't think I'll be able to fix that in this PR.

The main issue I see, which I'll bring up again, is that I think that the default population/mortality model is an extension of the MDP formalism, but I haven't yet seen a mathematical description of its form. It's a messier process when the single source of truth is raw code.

sbenthall:
I'm going to write up some math about this. I think it's needed for the documentation at the very least.
I expect there's also some venue (maybe not ones you are interested in yourself) where it is publishable.

mnwhite:
We could also split mortality so that simDeath happens at the end of the
period, drawing boolean flags for who_dies as the last step.  simBirth
would then happen at the beginning of the next period.

At the beginning of simOnePeriod, state_now automatically gets moved back
to state_prev *and* state_prev values for indices marked in who_died get
set to empty.  Then simBirth acts as it usually does, filling in state_prev
for who_dies.

We could then remove simBirth from initializeSim and instead just set
who_dies to all True.  This means that the initial population can't be
examined after calling initializeSim, but maybe we need to live with that.

Or we can just have a 1 line conditional that says the very first period of
time is slightly special.

mnwhite:
If you really want mortality in an MDP, the model is essentially that there
is some probability that each period, rather than experiencing the normal
intertemporal transition, your state instead gets randomly drawn from some
distribution that is exogenous to your idiosyncratic state. Moreover, you
weight future utility flows in this event by factor zero.

sbenthall:
These are all possibilities.

I wonder which one is the cleanest or best motivated?
I would prefer for this PR to not depend on that design decision any more than necessary since it's already doing a lot.
But I think it's a discussion worth having and getting right for sure.

The reason why I think the population and mortality aspects of HARK models are extensions to/beyond the MDP formalism is that:
 - MDPs tend to be single agent
 - we're talking about simulations with large AgentCounts and tracking the population dynamics over time (such as steady state distributions of income)
 - the assumption that the population is of constant size, and that a person who dies is immediately replaced with a newborn, is a strong assumption

I think I see how mortality can be internalized to the MDP for the purpose of determining optimal behavior.

mnwhite:
The death-and-replacement assumption is mostly for computational efficiency
/ pre-allocating memory.  It works best in a perpetual youth model, but
it's also fine with age heterogeneity as long as you don't want to do any
experiments with exogenously killing off part of the population, and
procreation is exogenous.

In models with endogenous birth or death, this method doesn't work.

sbenthall:
I'm making an issue to discuss this specifically now... I'd like to continue there because I think it is a deep and important topic but also a little far afield from trying to keep state variables in a namespace.

sbenthall:
I have been hung up on a difficult bug in the KrusellSmith classes.

One of the tests for the KS economy _solver_ was not passing.
I added tests, based on values I was getting from `master`, which has isolated the issue to something happening in the `solveAgents` step within the economy class.

But I'm a little stumped where to go from here.

sbenthall:
I believe the issue has to do with how the way the KS model is written in master, it's ambiguous whether `EmpNow` is a shock or a state.

I think it's technically an exogenous process but it's time dependent and so tracked as a "post-state".

In an earlier PR, I added a namespace for shocks. But I did not track `shocks_now` and `shocks_prev` separately. (I believe in some cases the shock history is tracked separately).

sbenthall:
Using fine-grained tests ( see #850 ) I believe I've been able to isolate the bug to

`KrusellSmithType.preComputeArrays()`

There is a changed numerical value in the `mNextArray` created in `preSolve()`.

I'm surprised, because as far as I know I'm not touching any of this stuff in the refactoring of the model variables for simulation.

@mnwhite any guesses what could be causing things to change this early on?

sbenthall:
Progress: More testing has confirmed that _before_ `preSolve()`, the `AFunc` that is being given to the KS agent has changed in this PR.

sbenthall:
Ok. I figured out the problem.

I had been building out the test suite in an incorrect way.

The way it had been organized, the new tests I was writing (see #850) were running `test_economy` first, then later `test_methods`, using the _same_ `Economy` object.

This was causing it to start with an already converged `AFunc` value, which was being distributed to the agent when `getEconomyData` was called.

I believe that the way `reset()` is written on the KS economy class and the Market class more generally, `reset()` will not change `AFunc` back to its original value! This seems like a bug to me.

What I'm doing now is rewriting the test cases in the recommended way: the `setUp` function is defined in one `TestCase` class, which is then subclassed for each more specific test.

Separately, I wonder if @mnwhite agrees that this behavior of `reset()` is not ideal. If so, I'll add an issue for it. This seems related, but not identical, to #849.

sbenthall:
Nope, it's not that easy...(!)

Now I've isolated the problem to `updateDynamics()`.
`AFunc` is not being set the same value across the two branches.

Worse, it looks like there's some stochastic variation in `updateDynamics` such that running the same test multiple times on the _same_ branch comes up with different `AFunc[0].slope` values.

sbenthall:
The calculation of the `AFunc` seems to depend on some attributes of the market, `intercept_prev` and `slope_prev`. These are as far as I know neither `sow` or `reap` variables on the market, but it looks like they are neverthless some kind of stored Market state that is iterated on.

What's happening here?

sbenthall:
The KS economy class had a bug.
It used a `copy()` of the parameters dictionary, instead of a `deepcopy()`.
Because the economy class stores the previous slope/intercept by mutating the lists, this was changing the default intercept and slope values.

This bug is now fixed in this PR.

sbenthall:
There was an error in my transitioning of KS into the new state namespace.
I didn't properly change `mNow`. Tests are passing locally now!

sbenthall:
ugh. One more case to fix: the Portfolio model.

sbenthall:
Ok. Tests are passing locally. **Phew**

The issue was with how I was dealing with initialization and birth of `AdjustNow` and `ShareNow`.
In the current Portfolio model, these are anomalous: they need to be set as "post states" for a period's transition function, but they are a shock and a control variable respectively. As such they are not handled yet in this new treatment of `state_now` and `state_prev`.

I'll need to revisit this design while working on #838 (namespacing the control variables).

Please merge #853 , new tests for ConsPortfolio model, before this PR.

Issue #845, opened 2020-09-30
TractableConsumerType does not have all default parameters needed to solve()


sbenthall:
Related to #844 

Trying to write this automated simple automated test of basic `solve` and `simulate` functionality for TractableConsumerType:
https://gist.github.com/sbenthall/771121ec30afeaefdf24a9b7ac109997

This test fails on the `solve()` step with the following error:

```
        # Define some useful constants from model primitives
>       self.PermGroFacCmp = self.PermGroFac / (
            1.0 - self.UnempPrb
        )  # "uncertainty compensated" wage growth factor
E       AttributeError: 'TractableConsumerType' object has no attribute 'PermGroFac'

TractableBufferStockModel.py:362: AttributeError
```

The problem seems to be that there is no value of PermGroFac passed through to the constructor upon initialization by default.

MridulS:
There is no startup dictionary for `TractableConsumerType` available.

Issue #844, opened 2020-09-29
TractableBufferstock example contains functionality not in automated tests


sbenthall:
Tests are passing while the TractableBufferstock example is not.
This means that some new tests covering the TractableBufferstock functionality should be written.

PR #848, opened 2020-10-01
Update tractable buffer stock model with changes to shocks and add initial base parameters

MridulS:
Fixes https://github.com/econ-ark/HARK/issues/845, https://github.com/econ-ark/HARK/issues/844

### DIFF
--- a/HARK/ConsumptionSaving/tests/test_TractableBufferStockModel.py
+++ b/HARK/ConsumptionSaving/tests/test_TractableBufferStockModel.py
@@ -11,10 +11,13 @@ class testPortfolioConsumerType(unittest.TestCase):
 
     def test_simulation(self):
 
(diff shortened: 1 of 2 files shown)


HARK/ConsumptionSaving/tests/test_TractableBufferStockModel.py
```diff
@@ -0,0 +1,23 @@
+from HARK.ConsumptionSaving.TractableBufferStockModel import TractableConsumerType
+import unittest
+
+
+class testPortfolioConsumerType(unittest.TestCase):
```
sbenthall, line 5: I think this class needs to be renamed?

sbenthall (changes requested):
rename test class to fit TractableBufferStock case

PR #850, opened 2020-10-08
some method-specific tests for KS

sbenthall:
just a few more tests.


mnwhite:
I'll merge when it's green.

sbenthall:
I'm adding a little more to this PR as I still need more tests to debug this KS issue in #836

Issue #851, opened 2020-10-08
rewrite parallel.py without exec


sbenthall:
parallel.py uses exec statements.
Can it be rewritten without them?

PR #852, opened 2020-10-09
Use getattr to execute commands instead of exec

MridulS:
Fixes https://github.com/econ-ark/HARK/issues/851

The code still contains a "hack" to strip the last two characters `()` from all the methods passed to `multiThreadCommands(Fake)`. Ideally going forward we should only pass in the name of the method in the list of commands. `[solve, initializeSim, simulate]` instead of `[solve(), initializeSim(), simulate()]`.

Should I refactor instances of this in the codebase for the next release or should we wait atleast for a 0.10.8 release.

If we move to `[solve, initializeSim, simulate]` this will break a bunch of REMARKs.

### DIFF
--- a/HARK/parallel.py
+++ b/HARK/parallel.py
@@ -27,7 +27,8 @@ def multiThreadCommandsFake(
     """
     for agent in agent_list:
         for command in command_list:
-            exec("agent." + command)
+            # TODO: Code should be updated to pass in the method name instead of method()
+            getattr(agent, command[:-2])()
 
 
 def multiThreadCommands(agent_list: List, command_list: List, num_jobs=None) -> None:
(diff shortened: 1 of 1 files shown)


sbenthall:
I would go ahead with it if I were you.
Are we sure that the methods can't be passed around directly?

MridulS:
Yeah as getattr will try to find "solve()" as a model attribute/decimation rather than "solve".

sbenthall:
What I mean is that if the method is passed directly, it can just be called with `agent` as `self`.

I wasn't sure if class methods could as `lambda`s for the purpose of pickling in this context? (Since they aren't anonymous)

mnwhite:
This looks good.

Issue #862, opened 2020-11-05
Simplest possible portfolio model with frames


sbenthall:
An implementation of framing in the simplest possible portfolio model, where reallocation is possible every period.

sbenthall:
This is going to be even "simpler" from the perspective of adapting code and be a rewrite of the original HARK portfolio model.

I need to do it this way because the first such model will use the frames for simulation only.

Using frames, with first order condition equations and the like, in the solver depends on some unresolved architecture issues see #856

PR #865, opened 2020-11-08
PortfolioFrameModel and FrameAgentType

sbenthall:
This PR addresses issue #862, with some movement towards #660 and #798 

The goal is to provide a general architecture for models that use the "frame" method of defining a model.

The main changes are:
 - `FrameAgentType`, a subclass of `AgentType` that changes the `simBirth` and `simOnePeriod` methods to make use of model configuration objects `birth_values` and `frames`
 - A `ConsPortfolioFrameModel` which is a conversion of the existing `ConsPortfolioModel` to use `FrameAgentType`.

The current implementation passes all automated tests, which have been converted from the original ConsPortfolioModel tests to use the new class. What that implies is that strictly speaking, this is just a refactoring of the old class code!

I have a few open design questions:
 - How should the `FrameAgentType` class deal in general with the setting of aggregate variables such as `PLvlAggNow`?
 - In order to make it as consistent as possible with the original code, I've defined the transition methods so that they mutate the underlying model variable storage. I personally think it's cleaner/more mathematical/more aligned with Dolo if the transition methods return the variable values, and the `FrameAgentType` handles the storage of the values in memory. I wanted to run this option by you.


sbenthall:
This PR now had CHANGELOG edits, updated docs, and full tests.
I think it's ready to merge.

sbenthall:
@Mv77 can you review this code?

Mv77:
Yeah, absolutely. On it now.

Mv77:
This looks very good to me. What I like most about it is that it allows policy functions (or any transition function, really) to depend on things that have not been defined at the start of the period.

As an example: in the basic portfolio model, the agent makes two decisions -- his consumption and his portfolio share. In the current implementation, both `cFunc` and `ShareFunc` are defined over start-of period market resources `m`. This is okay and might even be convenient in the simple setting. However, in theory, one can define some `ShareFuncAlt` over 'assets after consumption' `a`, because the portfolio decision depends only on end-of-period resources. In such case and if the solutions were perfect and not numeric, we would have `ShareFunc(m) = ShareFuncAlt(m - cFunc(m))` for all `m`.

This implementation of the simulation method allows the user to take the `ShareFuncAlt` approach if he so desires. This was not possible before, because `Share` --a control-- would have been needed to be computed before `a`, a post-state.

Comments:
- I know you were thinking about the solution side of this. We can chat about it whenever you'd like.
- One thing to think about is the relationship between simulation and solution frames. I do not know where is the internal discussion on this matter at the moment. Will simulation and solution frames have to correspond with each other? What happens to simulation frames in which the agent is passive (e.g.  `('RiskyNow','AdjustNow')`)?
- I am not sure I follow the naming convention of `frames`' entries. Is there a reason why `('aNrmNow', 'aNrmNow')` is repeated?

sbenthall:
Super. Thanks for this @Mv77 

> * I know you were thinking about the solution side of this. We can chat about it whenever you'd like.

My understanding of the roadmap for now is that we anticipate the solution side to not be ready for HARK 1.0, which has a sharp deadline due to the Sloan grant.

I believe it is feasible to bring in the simulation side by the HARK 1.0 deadline, hence the push on this PR.

@llorracc has opined that it would be best if we could have the simulation side anticipate the solution side as much as possible to make it easier to combine the two down the line.

So that's a major material reason to sync up on the design.

> * I am not sure I follow the naming convention of `frames`' entries. Is there a reason why `('aNrmNow', 'aNrmNow')` is repeated?

Nope, that's a bug. Thanks for catching it.

The intention was for those "labels" to be the model variables that are updated in that frame. In the current implementation those labels are not tied to functionality so it's easy for a bug to slip through.

I'm not wedded to this convention. But I do like it. I'm open to a better idea.

> * One thing to think about is the relationship between simulation and solution frames. I do not know where is the internal discussion on this matter at the moment. Will simulation and solution frames have to correspond with each other? What happens to simulation frames in which the agent is passive (e.g.  `('RiskyNow','AdjustNow')`)?

I think this is the crux.

My understanding is that the solution "stages" each correspond to the construction of one or more decision rules (such as `cFunc` or `ShareFunc`), and that these decision rules are connected to particular control variables in the model.

What if it was organized like this:
 - The model's frames are a partition of its model variables. [with some conditions on the partition]
 - Each frame contains a transition method for updating its model variables
    - In the future, this can be reduced to transition equations
 - If a frame contains a control variable, then it also contains a solution method
    - In the future, this can be reduced to a set of equations used by a generic solution technique, like the first order condition.

Then the solver should be able to search this list of frames, filter out the frames without solution methods, and solve them by backwards induction.

How does that sound to you?

sbenthall:
Hmm. Ok, thinking it through, I see a path forward on this.

I'll roll in a conservative subset of the changes from #866 into this PR. (Stopping short of the functional "equation" forms for the transition methods, which introduce a lot of complications).

Then I'll ping @Mv77 again to sync up on a call. In the meantime I'm looking at #832 and seeing how things line up.

Mv77:
> How does that sound to you?

Yes, that sounds exactly right. Keeping track of frames where solutions are computed is crucial as you say, for backward induction purposes.

> In the future, this can be reduced to a set of equations used by a generic solution technique, like the first order condition.

I am less optimistic about this, but maybe it is my lack of knowledge. I'm not sure we'll get to a generic solution technique. The equations used to represent controls might be complicated, to the point of just being optimization problems like `c = argmax { u(c) + \beta V(a)} S.t. ...`. This is because we already have problems like the ones in DCEGM where FOCs are not sufficient. So I'm not sure a simpler representation in these instances is available.

llorracc:
Reading through this briefly, let me reiterate strongly:  A key aspect of "frames" is that they should be formulated in such a way as to contain both the information necessary for forward simulation and the information needed for backward induction/solution.  I'd almost say that there's not much point in all the work of constructing frames if we don't do it that way: Their key advantage is to be the "single source of truth" about what goes on as you move forward and backward in time, and probably 80 percent of time that people currently spend in debugging these kinds of models occurs because there is some subtle difference in how they have specified the backward induction vs the forward simulation.  I'd rather post something extremely simple, but that satisfies the criteria above (like maybe the simplest portfolio model), than something much more complex that handles only one of the two stages (solution, or simulation).

This information is 90 percent overlapping.  An example of something that is not overlapping is that, if you are simulating, and this is the first period of the simulation, you probably need some method to specify default initial values of the variables.  And, if you are solving, you need some method to define the gridpoints at which the end-of-period problem should be solved.  But ONLY those things that are required to be special should be allowed to be special.

sbenthall:
Hmmm.  A lot of food for thought here.

@llorracc That's an interesting idea about implementing a framed version of an even simpler model first.

To me, the hardest part about working with HARK is how the solution code is written across the library. It makes it difficult to _find_ the model information, let alone replace it with information from a different model. I think we are on the same page about how ideally, this would be more modular.

I was going to try to tackle that by refactoring the solution code until @mnwhite went on a break from the project, because I would naturally want to check in with him to make sure I wasn't missing something important.

Without having refactored solutions which tease out the model information from the solution strategy, I don't think it's possible to inject that model information in from a frame definition.

There is a bit of a chicken and egg problem here.

However, I still think it's worthwhile to scaffold what a framed model looks like because it gets away from the Shocks/States/Controls/PostStates paradigm, which is confusing.

@Mv77 I think one thing that is not ideal about HARK currently is that each AgentType specifies a singular `solveOnePeriod` method. I think the way `dolo` does it is better: you define a problem, and then you can apply one of several solution algorithms to solve the problem.

https://dolo.readthedocs.io/en/latest/algos.html

The way `dolo` works is that you can only use an algorithm on a model if it has the appropriate equations defined. So for example, if you have transition equations and an arbitrage (FOC) equation, you can run a time iteration algorithm. But if you have the felicity function instead of arbitrage, you can use value iteration.

In my view, it would be best if the frame architecture supported several different solution algorithms. I think it would also be appropriate for it to support _sui generis_ solution methods where no known algorithm exists. The path to implementation I would take would be:
 - to repackage the current solution methods to be model-specific solution implementations within a Frame architecture
 - then work on refactoring those solutions to extract the model information (like the forward transition function) and wrap them in more general, frame-based solution algorithms

One tricky thing I have not figured out a good solution for yet is this: we would like the Frames to be a "single source of truth" for both the simulation and the solution, but so far our best plan is to write and store the forward and backward versions of the transition function separately. It would be much better if we could programmatically invert the transition function.

sbenthall:
> I'd rather post something extremely simple, but that satisfies the criteria above (like maybe the simplest portfolio model), than something much more complex that handles only one of the two stages (solution, or simulation).

Reading this again ...

The reason why I started with a refactoring of the portfolio model currently in HARK is that it is already very well covered by automated tests.

If what you are chiefly interested in is an example of a model, defined in terms of frames, with a solution method that draws on the model definition as an SST, I would be inclined to start with something even simpler, like the ConsIndShock model or even the PerfectForesight model, as that would keep the solution complexity from getting in the way of the rearchitecting.

sbenthall:
I'm beginning work on this PR again.
The design will be informed by many meetings with Chris and Pablo.
I'm merging material from #866 into this PR, as now the scope for this PR will be a much more full-featured AgentType model.

The goal for this PR is:
 - Configurable SST for model variable names and transition equations
 - Monte Carlo forward simulator
 - Value function iteration solver

This will be designed to handle the range of models considered in scope for HARK 1.0's "micro" models.

The main test case will be the original Portfolio agent class, which is now very well understood. It demonstrates the case of multiple sequential controls.

sbenthall:
This PR is now passing tests locally.
I submit it for review.

I expect this to be of interest to anybody interested in "frames" or "stages" as a way of breaking down a problem. So perhaps especially @Mv77 and @llorracc would be interested in this work. It also performs an alternative implementation of many original HARK internals and so may interest @mnwhite . Certainly any review would be very appreciated!

This PR has changes that implemented a FramedAgentType, which allows for a different kind of model specification.
It uses a variation of the Portfolio model as its example.

This deals with model specification and simulation only. Currently, these classes depend on some other solver to create the solution function. An anticipated next step after this PR is to develop a general solution method for framed agents. That is out of scope of this PR.

There are a few noted areas for improvement in the current PR that I haven't gotten to yet. One is to make 'aggregate' state variables and shocks, which are a bit awkwardly part of the AgentType in some cases, into a special kind of frame (i.e, have the Frame class take an optional parameter, `aggregate` and for this to be handled well). There's also a lot of cleanup to be done to reduce the transition functions into something lighter-weight.

HARK/ConsumptionSaving/ConsPortfolioFrameModel.py
```diff
@@ -0,0 +1,286 @@
...
+        PortfolioConsumerType.__init__(
+            self, **kwds
+        )
+
+        # Set the solver for the portfolio model, and update various constructed attributes
+        self.solve_one_period = solveConsPortfolio
+        self.update()
+
+        ## TODO: Should be defined in the configuration.
+        self.aggs = {'PermShkAggNow' : None, 'PlvlAgg' : None} # aggregate values
```
Mv77, line 75: I like what you have done here, fully specifying states, shocks, controls and aggregates.

When creating new models, a frustration of mine has been the inconsistency with which that specification is done across different HARK agent types. Its sometimes done, sometimes not, sometimes inherited so one has to go hunting for the original class' attributes, sometimes fully specified.

It is further complicated by two different objects  `shock_vars` and `shock_vars_` ([see e.g.](https://github.com/econ-ark/HARK/blob/10fe1efba5d6f558fc47433325ac987acfc09077/HARK/ConsumptionSaving/ConsIndShockModel.py#L1583)) with no indication of what each of them does. I think I have seen this for states too.

HARK/ConsumptionSaving/ConsPortfolioFrameModel.py
```diff
@@ -0,0 +1,286 @@
...
+                      # -- handled differently because only one value each per AgentType
+        self.shocks = {'Adjust' : None, 'PermShk' : None, 'Risky': None, 'TranShk' : None}
+        self.controls = {'cNrm' : None, 'Share' : None}
+        self.state_now = {
```
Mv77, line 79: was the `_now` not out of fashion now?

HARK/ConsumptionSaving/ConsAggShockModel.py
```diff
@@ -1022,9 +1022,6 @@ def get_shocks(self):
         # TODO: replace poststate_vars functionality with shocks here
         EmpNow = self.state_now["EmpNow"]
 
-        if employed.sum() != emp_permute.size:
```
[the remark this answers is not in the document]
sbenthall: pdb is the 'python debugger'. It looks like somebody (probably me) was debugging and accidentally got this breakpoint pushed into the master branch by mistake.

I could remove it in a separate PR, which would be cleaner, but did so here so I didn't forget.

HARK/ConsumptionSaving/ConsPortfolioFrameModel.py
sbenthall: These are doing some logical work as per the current HARK simulation code.
There, there's both state_now and state_prev, because transition functions can depend on the previous period's state.

I can see how we could do an implementation where the state memory array was duplicated each period, and updated frame-by-frame.

That's an implementation detail, as far as I'm concerned, not a matter of model logic. I'm not opposed to making the implementation consistent for shocks, controls, and states, as another PR. But there may be benefits to keeping state_prev.

HARK/ConsumptionSaving/ConsPortfolioFrameModel.py
sbenthall: What I would like to do ultimately is move all of this information into the frame definitions.
Then have the constructor code build these dictionaries based on those frames.
Ultimately, I want all model information in the frames dictionary, and nothing baked into a model class.
This is sort of an intermediary implementation.

HARK/core.py
```diff
@@ -863,6 +866,242 @@ def clear_history(self):
...
+    in a specified order, rather than assuming that they
+    are resolved as shocks -> states -> controls -> poststates.
+
+    Attributes
+    ----------
+
+    frames : [Frame]
+        #Keys are tuples of strings corresponding to model variables.
+        #Values are methods.
+        #Each frame method should update the the variables
```
Mv77, line 906: the the

HARK/ConsumptionSaving/tests/test_ConsPortfolioFrameModel.py
```diff
@@ -0,0 +1,146 @@
...
+
+        self.assertTrue(
+            np.any(self.pcct.shocks['Adjust'][0])
+        )
+
+        self.assertAlmostEqual(
+            self.pcct.state_now["mNrm"][0],
+            self.pcct.state_prev["aNrm"][0] \
+            * self.pcct.Rfree / self.pcct.shocks['PermShk'][0] \
+            + self.pcct.shocks['TranShk'][0] \
```
sbenthall, line 58: I want to note one thing in particular about the tests here.

I found that with the new architecture I was getting a slightly different sampling order for shocks. This is of course fine for the library, but it breaks the extremely brittle tests we've had which check for a specific numerical value.

In many places in this test suite, I've changed to this more resilient and accurate style of test. I'm now testing the relationships between model state/shock/etc. values, rather than particular numerical values.

That will mean that if the code is working properly, some change to the distribution sampling order won't break all the tests.

We might try to do this elsewhere in the library before doing a big random number generation shifting change, such as #1026

HARK/ConsumptionSaving/tests/test_ConsPortfolioFrameModel.py
```diff
@@ -0,0 +1,146 @@
...
+            self.pcct.controls["cNrm"][0],
+            self.pcct.solution[0].cFuncAdj(self.pcct.state_now['mNrm'][0])
+        )
+
+        self.assertAlmostEqual(
+            self.pcct.state_now['aNrm'][0],
+            self.pcct.state_now["mNrm"][0] - self.pcct.controls["cNrm"][0]
+        )
+
+class SimulatePortfolioConsumerTypeTestCase(PortfolioConsumerTypeTestCase):
```
Mv77, line 75: Something that would make me confident about the PR would be to have a test where a `ConsPortfolio` and `ConsPortfolioFrame` agent types were created and solved with the default parametrization and their solutions and simulations were compared.

Mv77 (commented):
I like the direction of the PR and how it makes the simulation code clearer.

I suppose the solution side is the next hurdle to be cleared.

llorracc:
@sbenthall,

Reviewing this seems like something that would be best accomplished, at least in the initial stage, via Zoom.  In particular, if you have a notebook as well as the code so you could do something like what I did last time with the pre-alpha HARK 2.0 post, where the notebook was demonstrating the substantive outcome but I also had the code up and Spyder running so we could look at the internals, that would be great.  

@Mv77 is in Bonn now (I think) communing with Fedor and the Great Spirits of dynamic structural estimation (or some of them), and it would probably be good to have him on the call.  So maybe a week from Thursday would be a good target?

HARK/ConsumptionSaving/tests/test_ConsPortfolioFrameModel.py
sbenthall: That's an interesting idea.
I don't think a comparison between the solutions will be interesting because it's literally just calling the same solution code.
Though I could do a few tests to verify that it's working as expected.

Comparing the simulations would need to take into account a couple things: 
 - differences in sampling. i suppose these could be done by averaging many cases.
 - the fact that in this model the simulation is computing a stochastic transitory shock for the 0th period of life, as opposed to the current default code that for some reason makes this shock 1.0.

Still, a side-by-side plot would be easy to do.

sbenthall:
Thanks for the review @Mv77 
I'll make some edits based on your comments, and try to move forward on a couple things I've mentioned (e.g. aggregation defined in frames).

That would make it polished for August 19th, which @llorracc you're suggesting as a review date.

I don't know if there's much in terms of new substantive results to this; it's mainly a refactoring job. But I can think about what would be useful for demo purposes. I am mainly hoping to have this PR go smoothly so that I can start working on generic solution code.

Mv77:
@llorracc I'm still in the US! I leave on the 14th and come back on the 23rd. I can chat this week if you would like.

An issue is that I have my pre-trip covid test on Thursday morning, so I can't make the usual HARK call.

HARK/ConsumptionSaving/tests/test_ConsPortfolioFrameModel.py
llorracc: @sbenthall 

That's great, thanks, you are right that our existing tests are much too finicky and I'm glad to learn that you are robustifying them.  To tell the truth, even 6 decimal places (which is what we have dialed down to from a default of 13) is far more than makes much sense for most of our tests.  3 or 4 digits is probably more than enough.

sbenthall:
Ok, I'm glad I compared the ConsPortfolioModel and the new PortfolioFrameModel simulation results directly.
I see now I'm getting some buggy behavior when sampling the shocks.
I hope to add repairs to the PR soon.

sbenthall:
I have repaired this PR.
Now `Risky` is an aggregate shock and `Rport` is computed as a separate frame.
(This shows, by the way, how this new architecture frees us from the misnamed `get_Rfree` method that was needed for backwards compatibility.)

The notebook in `examples/FrameAgentType` shows a comparison between a simulation with the new class and the old class.

Requesting review (again... thanks!) from @Mv77

Mv77:
No worries, I am a committed committer now. I'll slowly but surely get through it. :)

Mv77 (approved):
This looks good to me.

My only question just to make sure is: the simulations in the example do not *exactly* match just because of the RNG, right? The frame type is sampling different shocks at different points of execution with potentially different seeds from those of the regular portfolio type?

sbenthall:
Thanks @Mv77 

Yes, I currently believe that the change in random seed accounts and some of the sampling steps account for the discrepancies.

For example, the FramedAgent using IndexDistributions does not sample values for newborns twice (see #1021).

However, it is possible that there are remaining bugs in my code that result in other differences. These are naturally difficult to detect. If we can't find them, I hope we wouldn't let the perfect be the enemy of the good.

The new example notebook is designed to show and compare the two simulations.

Mv77:
Yes, for sure. Just wanted to make sure I understood. I approve!

llorracc:
small suggestion: effective way to detect bugs in this kind of context is often to dial down the variance of the shocks to zero, or very close to zero.  Bugs (if they exist) will often leap out at you when you do that.  Then, dial up one but not the other of the shocks.  etc.

PR #873, opened 2020-11-16
Make distanceMetric work on dictionaries

Mv77:
This PR adds support for dictionaries in core's `distanceMetric`.

`distanceMetric` already knows that the distance between two lists is the maximum of the element-wise distances between their elements. The same logic can be applied to dictionaries: the distance between two dictionaries is the maximum element-wise distance between their elements. This PR implements this logic.

An example of why computing the distance between two dictionaries might be useful:
-@sbenthall suggested that solutions to problems with 'sub-stages' could be kept in dictionaries, so that keys can be monikers of the sub-stages. I like that setup and, for its implementation in infinite horizon problems, it would be useful to compute the distance between two dictionaries of sub-stage solutions.


sbenthall:
@Mv77 This implementation makes some assumptions that are not perfectly general.

It assumes that the listed values of `dictA` will be commensurable with the listed values of `dictB`. However, while Python dictionaries are ordered, it is possible that two dictionaries with the same keys might have different element orders.

Consider two dictionaries: `dictA = {'x' : 0, 'y' : 1}` and `dictB = {'y' = 0, 'x' = 1}`. In the currnet logic, the distance between these dictionaries will be `0`. However, under the distance function's current logic (distance is maximum distance of the elements), the distance should arguably be `1` here.

I believe this implementation will also break if for some reason dictionaries with lengths are compared.

A good way to catch these kinds of cases ahead of time is to expand the test suite to include those cases. This is also a good way to work out what functionality you are really looking for.

Mv77:
Thanks for the feedback.

I wonder then what the desired behavior should be in terms of comparing things that are not alike. For instance, `distanceMetric` handles lists of different length by returning the difference in their lengths. This is would also apply to dictionaries of different lengths under the current implementation.

Should dictionaries with different keys and lengths be comparable? Should the program just crash? What about same length but different keys?

(Is this worth thinking about or should I just go on not passing dictionaries to `distanceMetric`?)

llorracc:
@Mv77, let's do two things:

1. Have it return a warning (but not an error) message when the objects are of different dimensionality;
1. But have it return the difference in the dimensionality (like, if one of the objects is 2 and the other is 3 dimensions, the difference is 1)

Mv77:
>     1. Have it return a warning (but not an error) message when the objects are of different dimensionality;
> 
>     2. But have it return the difference in the dimensionality (like, if one of the objects is 2 and the other is 3 dimensions, the difference is 1)

I added 1. and 2. in the last commit. The warning is also printed when comparing lists of different length.

There is still the question of what the distance of dictionaries with different keys should be:
- What is `distance_metric({'a': 3}, {`b`:3})`?
- What is `distance_metric({'a': 1, 'b': 2}, {'b': 2, 'a': 1})`?

My proposal, which I think is in the spirit of the current code, would be:

`distance_metric(dict_1, dict_2) = distance_metric(dict_1.keys, dict_2.keys) + distance_metric(dict_1.values, dict_2.values)`

Using the [discrete metric](https://en.wikipedia.org/wiki/Discrete_space) for strings (keys). This solution would work in the sense of being defined for any pair of dictionaries and being 0 if and only if the keys and elements match and are in the same order.

Mv77:
@llorracc @sbenthall

Mv77:
>     * What is `distance_metric({'a': 3}, {'b':3})`?
Log a warning.
>     * What is `distance_metric({'a': 1, 'b': 2}, {'b': 2, 'a': 1})`?
0
@llorracc

Mv77:
The previous commit incorporates all the feedback above:
- If dictionaries have different length, a warning is printed and the distance is the difference in lengths.
- If dictionaries have the same length but different keys, a warning is printed, and items are compared pairwise after sorting by key.
- If dictionaries have the same length and the same keys but in different order, items with matching keys are compared.

sbenthall:
@sbenthall to review and merge

Issue #130, opened 2017-08-31
Make better / more flexible format for warnings


mnwhite:
Since the very first version of HARK, we've had a "monkey patch" to the warnings module to make warnings display in a single line, rather than a big angry chunk of red text (in Spyder).  Some recent changes to numpy have increased the number of warnings it throws, for things like dividing by zero or using NaNs basically anywhere.  As a bunch of methods in HARK exploit numpy's handling of division by zero (1./0. = np.inf), it's probably a bit worrisome to new users when they see a bunch of "invalid value encountered in divide" messages come up.  I would like to turn off these warnings where we know that they're not relevant (because dividing by zero or carrying around some NaNs is our intended functionality), but our current warning format gives no opportunity for traceback.

Our monkey patch needs to evolve into a human patch, roughly speaking.  The definition in HARKutilities should be greatly expanded to offer easy options.  I'm not sure what's possible, but ideally I would like the option to:

- Turn off all warnings, period
- Nicely display the file and line number that generates the warning
- Display the warning with an N-level traceback.
- Pause execution until keyboard input when encountering a warning.

I'm assigning this to Nathan because he wrote the initial warnings patch and knows the most about this.

sbenthall:
See discussion in #283 

There is a standard and flexible `logging` package in Python.
https://docs.python.org/3/howto/logging.html#logging-basic-tutorial

I highly recommend using this standard functionality instead of creating new logging functionality.

sbenthall:
Responding to this, which may have been meant for this issue:
https://github.com/econ-ark/HARK/issues/283#issuecomment-597810736

The `warnings` controls in Python are managed through this package:
https://docs.python.org/3/library/warnings.html

sbenthall:
I think we are using the Python logging library now. So maybe it's a good time to revisit this?

MridulS:
Already implemented for ConsIndShockModel, to be extended across HARK.

mnwhite:
I think this was handled long ago. Warnings about intended HARK behavior are suppressed.

PR #885, opened 2020-12-16
replace print statements with calls to HARK logger, see #130

sbenthall:
Addresses #130 by using the HARK logger more widely in the library


sbenthall:
@sbenthall will fix conflicts and then pass back to @MridulS for review?

PR #888, opened 2020-12-17
Centralizing Value and MargValue functions

Mv77:
The trick of working with value and marginal value functions in 'inverse-space' is now applied in many models in HARK. This happens mostly through the `ValueFunc` and `MargValueFunc` classes.

However, the general idea of these classes is not captured in a centralized place. Different versions have been defined as models have been added: the 1-d version is in `ConsIndShockModel`, the 2-d version in `ConsGenIncProcessModel`, and my new portfolio model will use a 3-d version.

This makes it so that if I'm working on a new, say, 2-d model, I have to go check whether this has already been implemented in some of the existing models. Or if I need both 1-d and 2-d versions (as `ConsPortfolioModel` does), I need to import them from different models. I think this will lead to multiple versions of these objects being defined in different places.

This pr aims to construct general n-dimensional versions that can be imported from `interpolation` (or somewhere else?).

Would this be useful or is there a reason for these objects to be defined in different places?


Mv77:
The previous commits centralize `ValueFunc`. All of its uses now point to a single version in `interpolation.py`.

If you like this idea, I can continue with `MargValueFunc`, which is trickier as some instances of its use need its derivatives. Still, I think the task is possible and worthy.

sbenthall:
In my view this is exactly the right direction.

Mv77:
The previous commits implement n-dimensional `MargValueFunc` and `MargMargValueFunc`, and replace their previous uses and definitions.

sbenthall:
Would you like me to review this? I see it is not marked as WIP

Mv77:
Yes, please

sbenthall:
Beautiful work. Thanks for doing this. It passes review.

Could I ask you to please add a commit with an edit to the CHANGELOG describing the change and linking to this PR?

https://github.com/econ-ark/HARK/blob/master/Documentation/CHANGELOG.md

I believe it counts as a Major change. It should make it into the next release, which may be early this week.

sbenthall:
One other question: I've noticed this related code in the PerfectForesight model which is involved in constructing value functions.

https://github.com/econ-ark/HARK/blob/b45cabdd5b3f4d079f1d11660e08167311dc9036/HARK/ConsumptionSaving/ConsIndShockModel.py#L438

Is it possible to streamline that code any further given the changes you are making here?

Mv77:
> Is it possible to streamline that code any further given the changes you are making here?

Does not seem to me like it is. There are four lines of code, two using `ValueFunc` and `MargValueFunc` (that now will come from interpolation.py), the other two just defining a pair of points that are needed for ValueFunc in the special perfect foresight case.

Mv77:
> 
> 
> Beautiful work. Thanks for doing this. It passes review.
> 
> Could I ask you to please add a commit with an edit to the CHANGELOG describing the change and linking to this PR?
> 
> https://github.com/econ-ark/HARK/blob/master/Documentation/CHANGELOG.md
> 
> I believe it counts as a Major change. It should make it into the next release, which may be early this week.

Ready! (in the last commit)

Mv77:
Call them ___CRRA.

Mv77:
Since @sbenthall's review, I have:
- Added the change to the changelog.
- Changed the name of the functions to `[az]*ValueFuncCRRA`, to make it clear that CRRA utility is hardcoded in them.

Issue #625, opened 2020-04-16
Simplify expectation taking code


sbenthall:
The code that gets the expected values of distributions in the solver can be simplified by making generic methods on the distribution classes.

https://github.com/econ-ark/HARK/blob/master/HARK/ConsumptionSaving/ConsAggShockModel.py#L633-L667

sbenthall:
duplicate #621

sbenthall:
> Mathematica syntax is: 
> Expectation[f[x],x \[Distributed] LogNormal[mu,sigma]]

sbenthall:
This code is calculating the expectation of:
 * a mathematical function: `...` ?
 * over the Income distribution.... `IncomeDstn`

sbenthall:
Try to mirror the Sympy syntax:

sbenthall:
https://gist.github.com/mnwhite/a77387cf2633f0a2404342d39fdb5284

mnwhite:
I'm going to take a stab at this tonight or tomorrow morning.  It's easier
than I thought.

llorracc:
By "this" you mean working out sympy syntax for some expression(s) in your gist code?

mnwhite:
I'm making a non-sympy expectations function.  Here's the docstring, which won't format correctly here:

`
def calcExpectation(func,values,dstn):
    '''
    Calculate the expectation of a stochastic function at an array of values.
    
    Parameters
    ----------
    func : function
        The function to be evaluated, which can take M+N identically shaped arrays
        as arguments and returns 1 array as an output (of the same shape).  The
        first M arguments are non-stochastic, representing the inputs passed in
        the argument values.  The latter N arguments are stochastic, where N is
        the dimensionality of dstn.
    values : np.array or [np.array]
        One or more arrays of input values for func, representing the non-stochastic
        arguments.  If the array(s) are 1D, they are interpreted as grids over
        each of the M non-stochastic arguments of the function; the expectation
        will be evaluated at the tensor product set, so the output will have shape
        (values[0].size,values[1].size,...,values[M].size).  Otherwise, the arrays
        in values must all have the same shape, and the expectation is computed
        at f(values[0],values[1],...,values[M],*dstn).
    dstn : DiscreteDistribution
        The distribution over which the function is to be evaluated.  It should
        have dimension N, representing the last N arguments of func.
        
    Returns
    -------
    f_exp : np.array
        The expectation of the function at the queried values.
    '''
    pass
`

[...]

PR #893, opened 2020-12-25
using calcExpectation in ConsIndShock, see #625

sbenthall:
This PR is for using `calcExpectation` in the core model code where it's possible to use it.

This is to improve code readability, making reading the code more like reading the math from which it is derived.

Ultimately, this should reduce the number of lines of code as many `np.tile` and `np.sum` operations will be performed by the distribution code.

At this time, this PR makes a few small changes to the ConsIndShock solver.

**PROBLEM**: These changes introduce a small difference to the test results from their original target. This could easily be due to a small change in how the `calcExpectation` code works which is irrelevant to whether it is "working", or it could be due to a bug.

I have to request a review from somebody who is very familiar with the solution mathematics to assess whether the subsitutions I've made using `calcExpectations` here are sensible.


sbenthall:
I'm going to work on this in smaller pieces.

PR #897, opened 2020-12-27
rewrite calcExpectation to use numpy methods, see #625

sbenthall:
This rewrites `calcExpectation` to use `numpy.apply_along_axis`.
It is a _much_ more succinct implementation.

This rewritten version does not compute the tensor product in cases of multiple 1D value inputs.
I have argued in #625 that this should be handled by a different method outside of the `calcExpectation` function.

What this PR lacks in terms of functionality compared to the first implementation, it makes up for in automated tests.
This PR includes tests, including for multivariate distributions with both array and constant non-stochastic values.

### DIFF
--- a/Documentation/CHANGELOG.md
+++ b/Documentation/CHANGELOG.md
@@ -16,7 +16,7 @@ Release Data: TBD
 
 * Adds a constructor for LogNormal distributions from mean and standard deviation (#891)[https://github.com/econ-ark/HARK/pull/891/]
 * Uses new LogNormal constructor in ConsPortfolioModel (#891)[https://github.com/econ-ark/HARK/pull/891/]
-* calcExpectations method for taking the expectation of a distribution over a function (#884)[https://github.com/econ-ark/HARK/pull/884/]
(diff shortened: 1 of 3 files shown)


sbenthall:
@Mv77  please review!

Mv77:
This looks good and useful to me.

I am relying on the tests that you provide for my review for now, and they look awesome.

A more complete test would be to use this functionality in the solution of an actual model. I can try it in #832 as you have requested when I get back to working on it (I'll have to update it soon with all the cool stuff that's been added). I might have more useful things to say then.

But since this does not break anything and is a step in the right direction to potentially be improved upon in the future, I would say it passes review.

sbenthall:
Thanks for this review, @Mv77 
I'll merge this now.

As you try to work calcExpectations in to #832 , keep an eye out for this kind of tensor product manipulation:
https://github.com/econ-ark/HARK/pull/897/files#diff-29545ab437dbe5ab93d947afb0a70205323f0f9eba1d565d965adb5ceed1c732L1052-L1071

It's likely we will should be supporting this sort of operation in a reusable way, but that there's a simpler implementation of it that leverages numpy better. See discussion in #625

Issue #679, opened 2020-05-07
Model parameter getters and setters modeled on NetworkX methods


sbenthall:
Currently, HARK users are encouraged to dynamically reset model parameters on the model class like so:
```
    def test_another_solution(self):
        self.agent_alt.DiscFac = 0.90
        self.agent_alt.solve()
        self.assertAlmostEqual(
            self.agent_alt.solution[0].cFunc(10).tolist(),
            3.9750093524820787)
```

Here `DiscFac` is being reset by the user after initialization.

This use of a public attribute is in fact Pythonic. (I was incorrect earlier about using getters and setters in Python). But there are ways to set up hooks when the value is set using the `@property` decorator:

https://www.python-course.eu/python3_properties.php

This might be useful for testing conditions (for example, that `Rboro > Rsave`) when the model is being modified.

It may also be useful for storing the values in a way that is easier to programmatically access later. Ideally, there would be a way to export or display all the currently set model parameters as a dictionary (or some other format) and this is currently impossible. See #446

sbenthall:
This requirement is, on its surface, at odds with the idea of namespacing parameters #660
It would be possible in principle to redundantly store model properties in a namespace.
But the implementation implied by the python `@property` decorator stores the property value as an object attribute.

At this time when this issue was written, this user interface was considered important.

What is troublesome from a software design perspective about this is that there are currently multiple ways to reset a parameter. There is this way: setting it via an attribute setter.

There is also the way the `HARKobject` class is given a `__call__()` method mapped to `assignParameters()`

https://github.com/econ-ark/HARK/blob/master/HARK/core.py#L172-L177

This violates the Zen of Python: o"There should be one-- and preferably only one --obvious way to do it." (see [Zen of Python](https://www.python-course.eu/python3_history_and_philosophy.php))

We need to give up one of these ways of doing things. Which will it be?

We can either:
 1. Allow parameters to be assigned to attributes/properties of a Model class. This will require hard-coding the parameters of a Model as `properties`.
 2. Allow parameters to be provided in a more flexible way, but in a dedicated data structure. This is the direction other parts of the library have been moving in, as this matches the more general architecture of Dolo.

Because `assignParameters` works with any dictionary, there's no way to keep track of those parameters without storing the dictionary itself. This is what is being done in solutions to issues like #612 see #903

sbenthall:
This is related to #640 . The use case is when the user wants to change the standard deviation of the income distribution and rerun the model.

[...]

Issue #479, opened 2020-01-29
Show model equations for inspected HARK AgentType and Market classes


sbenthall:
For comparison, the `dolo.compiler.model.Model` class instances, when inspected in the REPL, provide some nice metadata and the equations for the model.

https://gist.github.com/sbenthall/54772209ab5108801a5fe166cff12b5b

Currently, with HARK AgentType and MarketClasses, it is difficult to find the equations of the models by inspecting the code or the objects.

The equations could potentially be put in the docstrings of these classes.
Or they could be printed to the user when the object is inspected. 

This may be related to #446 -- this could be a way to communicate the parameters used in the object definition in a structured way.

sbenthall:
depends on #761

sbenthall:
Related to #660 #612 

- If we refactor to have parameters grouped/ available , it will be possible to display them/

We don't have a SST on transition equations. The closest we'll get to that by 1.0 is FrameAgentType.

sbenthall:
Moving to 1.1 now that this is resolved for parameters.
need frame/equations explicit for next stage

sbenthall:
See #1085 
Will be feasible with #1427

mnwhite:
Closing because this is a major part of the HARK 1 roadmap, and is actively being worked on.

PR #903, opened 2021-01-06
replace HARKobject with MetricObject and Model

sbenthall:
This PR implements a change brought up in issue #612 

It replaces `HARKobject` with two classes, `MetricObject`, which has the distance metric functionality, and `ParameterizedObject`, which has the parameter assignment functionality.

The motivations for this are:
 - there is no use case for using the distance functionality on agents; this is only used for elements of a solution
 - the uses of `assignParameters` in solution elements could be written in the same number of lines using standard Python constructor formatting
 - there are many reasons why parameter assignment for models may need to be handled in a special way (see #612 #660) that is not assigning them to object attributes (which is normally what constructors do). So preserving the ParameterizedObject superclass will make it easier to maker these other changes.


sbenthall:
Latest commits here continue on this trajectory of work:

 - renamed `ParameterizedObject` class to `Model`; this now corresponds to the class with the same name in Dolo: https://github.com/EconForge/dolo.py/blob/master/dolo/compiler/model.py#L523
 - Adding a `__repr__` to `Model` so that parameters are printed upon the inspection of the object. Similar to Dolo: https://github.com/EconForge/dolo.py/blob/master/dolo/compiler/model.py#L641-L698 This addresses #479 as much as possible without Frames.

sbenthall:
Now, `Model` class has equivalency method `__eq__`, fixing #612 
Asking @MridulS to review

sbenthall:
The following tickets are being resolved or blocked by this PR:
 - #479
 - #660 
 - #679 
 - #612 
 - #889

MridulS (approved):

PR #909, opened 2021-01-13
Fix DCEGM/calcCrossPoints' return fields 

Mv77:
DCEGM's `calcCrossPoints`, which I implemented a while ago, has an annoying behavior.

The number of things it returns depends on whether or not it finds crossing points in the value function. This produces bugs when one calls the function, as one can not know how many things will be returned. This tiny PR fixes that issue.


sbenthall:
Makes sense to make this change.
A nitpick: to keep the type signature consistent, the second returned element should be an `ndarray` not a list.

This can be done with `np.array([])` or `np.empty(0)`

Mv77:
Great point. Fixed in the previous commit.

Issue #878, opened 2020-12-08
Document time scaling of parameters


sbenthall:
From @llorracc 

> the point that the baseline version of HARK is calibrated so that a "period" is a year, and therefore the interest rate and rates of return and size of stochastic shocks are all at a yearly rate, is something that should be in the documentation somewhere

> "cstwMPC scale the yearly calibration to a quarterly calibration, and so if you want to think of the highest frequency with which any consumer rebalances their portfolio as being once a quarter, you could use their income calibration)."

I believe this raises questions more generally about how to document default parameters and their provenance.

Here is an example of where the default parameters in HARK are set. It is currently part of the default Model definition.

https://github.com/econ-ark/HARK/blob/master/HARK/ConsumptionSaving/ConsIndShockModel.py#L2008-L2035

Maybe there is a way to given these parameter objects docstrings.
It would be helpful if the _units_ of the parameters were documented ($$$, $$$/day, etc.).

I don't have a lot of insight into how these parameters were chosen; has @mnwhite recorded that anywhere?

sbenthall:
On this topic, for this parameter:

`"aNrmInitMean": -6.0,  # Mean of log initial assets`

It would be best to specify the base of the log.

mnwhite:
All models now have default parameter dictionaries, and they're all fully defined in the model file for each AgentType subclass. *Within* a file, dictionaries can be copied and adjusted (only a couple instances of this), but all the information is always there in the file.

As for how values were chosen: They're arbitrary or represent well known plausible values. E.g. no one is shocked by a risk free interest factor of 2% or 3%, and any CRRA between 1 and 5 is reasonable / defensible.

I believe all parameters in HARK are properly unitless.

Issue #592, opened 2020-03-22
`init_lifecycle` should be replaced with a real life cycle calibration


llorracc:
At present, the `init_lifecycle` calibration does not reflect a realistic description of a life cycle; it's just a 10 period finite-horizon model.  We should replace it with something based on the calibration in SolvingMicroDSOPs, but updated to be a more realistic calibration (the SolvingMicroDSOPs calibration dates from 1997, and a lot has been learned since then).   

I didn't go ahead and do this myself right now because:

1. `init_lifecycle` is imported via for a couple of tests, and so I want @sbenthall, who created those tests, and @mnwhite who wrote the original `init_lifecycle` to be involved in the rewrite.  
   * When done, we should also use the new `init_lifecycle` in the `LifeCycleModel` notebook in `HARK/examples`.  
1. I want to do a more serious job of parameterizing the model.  In particular, we need 
   * a better representation of the age profile of transitory and permanent shocks, from Sabelhaus and Song or Panousi et al or some other source
    * To incorporate aggregate growth
    * To construct a plausible calibration of distributions at age of workforce entry of:
        * Permanent income
        * Assets
    * To have a parallel calibration for a version of a `persistent shock` version of the model
        * lp_{t+1} = \gamma_{0} + \gamma_{1} lp_{t} + \psi_{t+1}
        * ly_{t+1} = lp_{t+1} + l\theta_{t+1}
        * where l indicates log, \psi and \theta are IID, and typical calibrations would have \gamma_{1} \approx 0.97

This would be a nice mini-project for an econ PhD student.
```
$ cd HARK ; grep -rs --include="*.py" 'init_life' * | grep import 
HARK/tests/test_simulate.py:from HARK.ConsumptionSaving.ConsIndShockModel import IndShockConsumerType, init_lifecycle
HARK/ConsumptionSaving/tests/test_modelInits.py:from HARK.ConsumptionSaving.ConsIndShockModel import IndShockConsumerType, init_idiosyncratic_shocks, init_lifecycle
```

sbenthall:
Is this a duplicate of #586 ?

llorracc:
It's an elaboration of #586

Mv77:
I will start working on this. I will start out simple. Here is my plan (we'll see how far I get this winter).

- [ ] Start with the calibration from SolvingMicroDSOPs. Add that one to HARK and make sure it can be called by the examples in an easy enough way.
- [ ] Add the age-varying shock variances.
[...]

PR #921, opened 2021-01-18
Sabelhaus & Song (2010) age profiles of income volatility

Mv77:
This PR adds tools to produce age-profiles of income volatilities from Sabelhaus & Song (2010). It uses the original estimates from the paper, which were generously shared by John Sabelhaus.

The code is pretty much ready but I'm looking for feedback on:
- **Where should the estimates used to produce the profiles go?** They are a small enough set of coefficients that I just put them in the python script. Would it be better to have them in external csv files?
- **The location of the files within HARK:** I see these tools being used in multiple places and given the flexibility afforded by implementing a function with different specifications, I thought they deserved their own file. Do you agree? Would you prefer all of this to just be written at the bottom of `ConsIndShockModel.py`? Would this be considered a dataset and thus be better placed at `HARK/datasets` (re: @sbenthall 's comment in #592)?

I also need to better document the time units (#878).


Mv77:
This PR is now ready. I'd like to request review @sbenthall @llorracc .

I am still looking for feedback on the location and format of these tools.

Yet, I am very happy with the results. @llorracc I added smoothing using a polynomial by default as you requested, but kept the option to get the original results (see the example file).

I am particularly satisfied with the tests: they check the outputs against the coefficients from the original paper. Also, they check that smoothed versions are not more than 10% off from the original results.

I also added a README in the folder crediting John Sabelhaus.

PR #922, opened 2021-01-19
[WIP] Initial distributions of permanent income and mNrm

Mv77:
#592 asks for "a plausible calibration of distributions at age of workforce entry of: -Permanent income -Assets". In discussions with @llorracc, we determined the best place to get such a calibration would be the SCF. This PR works on adding calibrations of these two distributions coming ---as directly and transparently as possible--- from the SCF.

An issue is that we clearly do not want to have the full SCF in the toolbox. The solution that I propose here is:
- Keep only a relatively light table with all (most of?) the summary statistics we will possibly need: `WealthIncomeStats.csv`. It currently contains survey-weighted means and standard deviations at multiple age-year-education combinations. I can easily expand it to contain, e.g., deciles in case one wants to avoid lognormality.
- Link to [a public repo that produces the table directly from raw SCF files](https://github.com/Mv77/SCF-IncWealthDist).
- Have functions that parse the contents of the table to produce the exact parameters that HARK's models need.

This clearly need more documentation, which I will add. I first wanted to check whether you (@llorracc @sbenthall ) like this solution of keeping only the table and linking to the external repo that reproduces it.


Mv77:
@llorracc Chris, not sure if you merged this by accident, but it was work in progress. It's a bit far from complete. I just wanted initial feedback on how you all felt about the external repo solution. It needs at least:
- Documentation.
- A way to transform the moments to real dollars for any year one wants. I wanted to discuss the best way to do this in our call today.

PR #930, opened 2021-01-21
Module for inflation adjustments based on CPI

Mv77:
This PR adds a module to `HARK.datasets` that contains the CPI research series from BLS.

It contains tools to read the raw dataset and extract "deflators" from it, which can then be used to rebase U.S. dollar quantities to any year from 1977 to 2019.

This is useful when combining calibrations from different papers/sources that use different base years.


Mv77:
Currently failing because the BLS file is distributed as an `xlsx` and Pandas does not have the dependencies to read from excel in the dependencies. Will consult whether to manually transform to CSV.

Mv77:
Manually converted file from `xlsx` to `csv` as discussed in our call.

This is now ready for review (after tests complete).

@sbenthall could you review?

PR #932, opened 2021-01-23
Initial distributions of permanent income and mNrm

Mv77:
(This revives #922, which was merged accidentally and reverted)

#592 asks for "a plausible calibration of distributions at age of workforce entry of: -Permanent income -Assets". In discussions with @llorracc, we determined the best place to get such a calibration would be the SCF. This PR works on adding calibrations of these two distributions coming ---as directly and transparently as possible--- from the SCF.

An issue is that we clearly do not want to have the full SCF in the toolbox. The solution that I propose here is:
- Keep only a relatively light table with all (most of?) the summary statistics we will possibly need: WealthIncomeStats.csv. It currently contains survey-weighted means and standard deviations at multiple age-year-education combinations. I can easily expand it to contain, e.g., deciles in case one wants to avoid lognormality.
- Link to a public repo that produces the table directly from raw SCF files.
- Have functions that parse the contents of the table to produce the exact parameters that HARK's models need.


Mv77:
@llorracc requested that [the repo where the table comes from](https://github.com/Mv77/SCF-IncWealthDist) be able to download all the data to produce the table. I've done this now.

If one clones that repo and runs `do_all.R` it should download and compute everything from 0.

Mv77:
This is marked [WIP] only because it relies on #930, so that should ideally be merged in first.

Issue #920, opened 2021-01-18
remove Now from model variable names


sbenthall:
their place in time is now determined by data structures and namespaces

sbenthall:
Turns out this is hard to do without doing 933 first.

PR #935, opened 2021-01-28
MrkvNow into model state shocks['Mrkv'] in AggShockMarkov and KrusellSmith models

### LINKED ISSUE #933: refactor MrkvNow into model state variable in ConsAggShockModel
In ConsAggShockModel, the model variable MrkvNow is currently stored as an object attribute.

Is it a state? A shock?

Either way, it should be changed into a namespaced variable for consistency with the new style.

sbenthall:
Fixes #933.

In the AggShockMarkov and KrusellSmith agents and associated economies, the `MrkvNow` property was being set as an object attribute, despite being a model variable. It's an aggregate property in both cases.

This PR replaces these with a shock `Mrkv`. This is small progress towards #920, which otherwise can be done with simple search-and-replace operations.

`Mrkv` is a shock and not a state in this implementation because the way Market interactions with the Agent currently works, if the Market sets and agent state in its `sow` step, it will be rolled over into the `state_prev` namespace during the simulation in the `cultivate` step. Because here the `Mrkv` state really is exogenous to the agent, `shocks` seems more appropriate, though I hope we can scaffold this a bit better in, say, v1.1


sbenthall:
Needs CHANGELOG. Merge after.

Issue #120, opened 2017-07-10
New functionality: correlated discretized RVs


mnwhite:
As is, our only method for automatically generating discretized joint distributions is `combineIndepDstns`.  HARK.utilities should also contain methods for making discrete approximations to correlated multivariate distributions.  Three requests:

1) To start really simple: continuous uniform distribution with linear relationships.  
    * In 2D, the graph of the continuous PDF is a parallelogram; the discrete approximation would be approximately a parallelogram but with jagged edges

2) Discretized multivariate normal distribution.  Take in vector array of means and matrix array of varcov, return joint normal approximation.

3) Same as (2), but lognormal with underlying lognormal distribution as specified.

NB.  We plan to move toward a system like that of [dolo](https://github.com/EconForge/dolo) for discretizing distributions.  The best way to do this task might be to figure out how to construct a front end to dolo's [existing tools for doing this](https://dolo.readthedocs.io/en/latest/model_api.html?highlight=discretize).

This is related to #119 and #121

econ-ark:
A full description of what is desired, with the math, is at [Chris Carroll's public lecture notes site](http://econ.jhu.edu/people/ccarroll/public/LectureNotes) in MathFacts/Equiprobable-Returns.  The problem is solved and graphs produced by Mathematica code that is part of an [archive here](http://econ.jhu.edu/people/ccarroll/public/LectureNotes/MathFacts/Equiprobable-Returns.zip]

mnwhite:
compumetrika's (NMP) comment on #119 is probably more relevant for this issue; I think he was responding to the last line of my original post in that issue.

sbenthall:
@Mv77

Mv77:
I see this marked as a 1.0.0 milestone, but I have less than a week left and Chris told me to prioritize the calibration stuff.
I might get to it eventually if needed for my dissertation (likely), but that's a bit far down the line.

sbenthall:
ok, i just moved it because I see that #623 was marked for 1.0.

I guess all this can be punted to 1.x

llorracc:
Yes, it can be punted.  Mateo will indeed need to do it eventually but the calibration stuff nailed down is definitely higher priority.

sbenthall:
After #640 this will depend on a representation of a correlated multivariate (e.g. lognormal) distribution.

[...]

PR #948, opened 2021-02-10
Multivariate normal with discrete approximation

Mv77:
#120 Requests methods to deal with correlated discretized RVs.

This PR adds a multivariate normal distribution class, with a discretization method.

The idea for approximating an N-variate normal is:
- Construct a discretization for N independent standard normals.
- Join them as in `CombineIndepDstns`, taking the product of their grids (so you get an N-dimensional grid) and multiplying their pmfs.
- Rotate the grid using the [eigendecomposition](https://en.wikipedia.org/wiki/Eigendecomposition_of_a_matrix#Real_symmetric_matrices) so that the resulting grid has the VCOV matrix that you want.

I have not found a citation for this, and I'm not familiar with the literature, so I'm not sure if this is inaccurate or inefficient. It seems to work well in my tests.

A multivariate log-normal like the one requested in #120 should be easily derivable from this.

@llorracc @sbenthall what do you think?


Mv77:
Here's an example of how it looks:
image

The black dots are draws and the red are the discretized X's. Both from a negatively correlated multivariate normal.

llorracc:
@Mv77 this looks great!  I'd propose that a good test of it would be to implement the multivariate lognormal you mention as an easy application, because it should be easy to compare that to something that surely must exist which is some kind of tool for calculating the CDFs of continuous multivariate lognormals.

llorracc:
@Mv77, I'd like to merge this in if you will resolve the conflicts.

Mv77:
> 
> 
> @Mv77, I'd like to merge this in if you will resolve the conflicts.

Great, I've just resolved them. This should be ready to merge after tests pass.

PR #957, opened 2021-02-20
Add csv's to MANIFEST.in

Mv77:
My recent calibration tools added a couple datasets in csv format to the toolbox.

However, I had no idea these needed to be in the `MANIFEST.in` file to get collected when one installs HARK.

I noticed the issue while messing around with some conda environments for a cluster, and wondering why my tools were not working after installing HARK.

This tiny pr fixes the issue. It is my first time working with `MANIFEST.in`, could someone review?

### DIFF
--- a/Documentation/CHANGELOG.md
+++ b/Documentation/CHANGELOG.md
@@ -42,6 +42,7 @@ interpolation for problems with CRRA utility. See [#888](https://github.com/econ
 * AgentType simulate() method now returns history. [#916](https://github.com/econ-ark/HARK/pull/916)
 * Rename DiscreteDistribution.drawDiscrete() to draw()
 * Update documentation and warnings around IncShkDstn [#955](https://github.com/econ-ark/HARK/pull/955)
+* Adds csv files to `MANIFEST.in`. [957](https://github.com/econ-ark/HARK/pull/957)
 
 ### 0.10.8
 
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -11,7 +11,7 @@ include MANIFEST.in
 include setup.cfg
 
 # include data files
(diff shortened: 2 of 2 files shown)


llorracc:
@MridulS, this looks like one for you.

MridulS (approved):

MridulS:
@mv77 yup, need to put in these data files to make sure its packaged properly :)

MridulS:
@Mv77 Do you think it's possible to get a more flat structure in the datasets directory? Like moving all the csv files to the `HARK/datasets/data` folder and the data prep functions to `HARK/datasets`, next to the load_data.py file?

MridulS:
@Mv77 just wanted to bump this up.

Mv77:
I tried to get feedback on the location and format of the tools in my PR's (See #921, #922) and calls with @llorracc .

I proposed a format for datasets in which each had its folder with:
- Data in as raw of a format as possible/sensible.
- tools to parse the data into model params.
- A readme explaining the format, variables and tools.

@llorracc and @sbenthall seemed to like it.

Flattening the dir structure is possible and would be advantageous in some dimensions, but it would go against concerns and principles that were brought up when I was creating the tools like:
- Clutter. E.g., if we eventually add life tables for other countries in different formats, do we want them in the same folder?
- The desire of data being as raw as possible.

I am not sure where the balance of those conveniences and concerns lands.

MridulS:
Thanks for the links, yeah I agree its a tricky balance between future proofing our code and conveniences.

Something like `from HARK.datasets.SCF.WealthIncomeDist.SCFDistTools import income_wealth_dists_from_scf` just seemed a bit too verbose on the first glance :)

PR #980, opened 2021-03-02
Fix Sabelhaus-Song variances

Mv77:
I made a grave error with the calibration tools. Sabelhaus & Song's model returns income shock variances, while HARK expects standard deviations. I did not transform the variances and thus, at the moment, shock variances are being interpreted as standard deviations. This can cause income uncertainty to be understated by a factor of 10.

This PR fixes the error. My sincerest apologies for this bug and its timing.


llorracc:
Oops!  Thanks for catching it -- just in time, we haven't released 0.10.9 yet, but are hoping to do so tonight or tomorrow.

PS.  Maybe you don't need to update the changelog because you had updated it already with your earlier PR?  If that's right, just tick off the "changelog" tick box, and I'll merge.

Mv77:
That's great to hear. Yes, the tools were already in the changelog. I have simply added a link to this PR. It can be merged after tests complete (although they should pass).

Thank you and sorry again!

Mv77:
All tests passed, this can now be safely merged!

PR #981, opened 2021-03-09
Newborns' death probabilities.

Mv77:
The current implementation of `sim_death` finds death probabilities as `DiePrb = DiePrb_by_t_cycle[self.t_cycle - 1]`.

For newborns, `t_cycle==0` so this becomes `DiePrb_by_t_cycle[-1]`. Thus, newborns are getting the last survival probability, which in life-cycle calibrations is usually low. Therefore, many die immediately after being born.

Therefore, I'd say this bug results in the unnecessary deaths of many newborns. If that is not worrying enough, I think it may cause many simulation observations to be wasted, as they are spent on instantly-dying newborns.

I am not the most familiar with `t_cycle` vs `t_age`. Could someone more familiar review? 


mnwhite:
The problem you're describing shouldn't happen. The newborns are put into
the set of agents *after* the death event happens. See
AgentType.get_mortality(), which runs sim_death() and *then*
sim_birth(who_dies). The first opportunity an agent has to die is at the
very beginning of their second period of life, when t_age=1 and t_cycle=1
(in a lifecycle model). Thus it will correctly pull from LivPrb[0] for
these agents. That's what the comment, "Time has already advanced, so look
back one" means in ConsIndShockModel's sim_death() code.

If it *does* happen, we need to fix it.

Mv77:
Thanks!

I was running into the issue yesterday. I was pretty convinced that the problem was happening, at least in the very first period of simulations. Maybe the issue is that `initialize_sim()` spawns newborns and then `sim_one_period -> get_mortality` kills them. I'll double check soon and get more evidence.

mnwhite:
Yeah, it's possible that it happens in the very first simulated period, but
that's harmless-- It just means one batch of newborns is immediately
replaced by another batch before they ever do anything. It's still
random draws of newborns.

sbenthall:
Might be related to this issue
https://github.com/econ-ark/HARK/pull/499

Mv77:
I've double checked.

This is an issue that happens only once. `initialize_sim()` creates newborns and the first thing that `sim_one_period()` does is call `get_mortality()`. That first instance is when this bug can occur.

As you said, it is harmless: a set of newborns is replaced with another. Then the question is whether you want it to happen or not.

The advantages of the proposed fix are:
- Saves some computational cost of re-simulating a part of the first batch of newborns.
- Eliminates this counterintuitive behavior.

The drawback:
- It adds an extra line and its computational cost to something that gets called very often.

If you'd rather not incorporate it, this can be closed.

mnwhite:
My suspicion is that it will be an extremely small fraction of the runtime
of a structural estimation's objective function, on the order of about
1/1,000,000 of the execution time.

llorracc:
For things like this, the best resolution is one that prevents any future wasted HUMAN time, so (a) whatever is less confusing and (b) add a comment so any future @Mv77 will not occupy more than 30 sec grokking it

sbenthall:
Tests are not passing.

llorracc:
@Mv77,

Could you revise the PR to just involve adding a comment or two to the existing code which, if such comment had existed 9 months ago would have prevented you from confusion?  (The comment should include a link to this (revised) PR.

Mv77:
> 
> 
> @Mv77,
> 
> Could you revise the PR to just involve adding a comment or two to the existing code which, if such comment had existed 9 months ago would have prevented you from confusion? (The comment should include a link to this (revised) PR.

This is done in the last few commits. This PR is ready to merge now.

Sorry for throwing everyone into this rabbit hole.

PR #990, opened 2021-03-28
Fix type-related bug in DCEGM

Mv77:
Recent versions of numpy became more strict with respect to data types. Particularly, np.NaN is no longer accepted in integer arrays. This was causing a crash in the DCEGM code whenever it was used with numpy 1.2.0 (which the binder instances do). For instance [the DCEGM DemArk binder link](https://mybinder.org/v2/gh/econ-ark/DemARK/master?filepath=notebooks/DCEGM-Upper-Envelope.ipynb) crashes.

This PR fixes the issue.

### DIFF
--- a/Documentation/CHANGELOG.md
+++ b/Documentation/CHANGELOG.md
@@ -8,6 +8,16 @@ For more information on HARK, see [our Github organization](https://github.com/e
 
 ## Changes
 
+### 0.12.0
+
+Release Data: TBD
+
+#### Major Changes
+
+#### Minor Changes
+
+* Fix bug in DCEGM's primary kink finder due to numpy no longer accepting NaN in integer arrays [#990](https://github.com/econ-ark/HARK/pull/990).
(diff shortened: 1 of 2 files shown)


Mv77:
I have not updated the change-log because I'm not sure what the next release will be and what format you'd like to have (no changes have been recorded since 0.11).

sbenthall:
Let's assume that there will be a 0.12.0 release with data TBD
Then you can fill out the changelog accordingly.
If we wind up doing a 1.0 instead of a 0.12.0 we can update it before the release.

sbenthall:
Is this the same thing as #988 looks like it?

Mv77:
> 
> 
> Is this the same thing as #988 looks like it?

It is, yes. Any of the two solutions should work.

Mv77:
Changelog updated. This is now ready to merge (Alan's solution would also work).

PR #998, opened 2021-04-07
updated IndShockConsumerType cyclical example's CyclicalDict

andyzheng97:
changed 'Rfree' and 'DiscFac' in CyclicalDict to quarterly frequency


llorracc:
@andyzheng97,

Thanks for posting this PR.  

We have automated checks that try to test whether code that people ask to be merged runs successfully, and those tests are failing.  

This is likely due partly to an annoying conflict between two of the greatest technologies going: github and jupyter notebooks.  They don't play nice together.

To resolve the issue, let me ask you to install jupytext on your machine (google it -- you want the "percent" format).  Once you have installed it (and tested that your code works using the CURRENT version of HARK), please repost your PR, which should now have both the jupyter notebook (ipynb) and the simplified python version (py extension).  This will make it much easier to review your PR.

PS.  Looking at the notebook again: Do you understand why LivPrb is multiplied by 4?  I do not.  Seems to me that, if LivPrb at an annual rate is 0.98, it wants to be RAISED to the (1/4) power rather than multiplied by 4.  It will be a good exercise to think this through.  This is a good lesson:  There can be mistakes in online code.  But, also, sometimes things that seem like mistakes, might not be.  I'm not sure which category this falls into, but thinking it through carefully is actually an excellent exercise for you -- partly because it will teach you to be skeptical!

PPS.  I'm cc'ing @sbenthall because he has been thinking about these issues of quarterly adjustment recently.

PR #1006, opened 2021-04-30
core.py updates to review with Seb

llorracc:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -284,3 +284,6 @@ nate-notes/
 # VSCode 
 settings.json
 .idea
+
+# Spyder project
+.spyproject
--- a/HARK/ConsumptionSaving/ConsIndShockModelFast.py
+++ b/HARK/ConsumptionSaving/ConsIndShockModelFast.py
@@ -1193,6 +1193,11 @@ class IndShockConsumerTypeFast(IndShockConsumerType, PerfForesightConsumerTypeFa
             self.solution_terminal.mNrmGrid = np.linspace(0.0, 1.0)
 
     def post_solve(self):
+
+        for i in range(len(self.solution)):
+            if hasattr(self.solution[i], "parameters_solver"):
+                self.solution[i].parameters_solver["solution_next"] = None
+
         self.solution_fast = deepcopy(self.solution)
 
         if self.cycles == 0:
--- a/HARK/ConsumptionSaving/ConsIndShockModelFast.py
+++ b/HARK/ConsumptionSaving/ConsIndShockModelFast.py
@@ -1106,6 +1106,11 @@ class PerfForesightConsumerTypeFast(PerfForesightConsumerType):
         )
 
     def post_solve(self):
+
+        for i in range(len(self.solution)):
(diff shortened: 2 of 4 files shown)


llorracc:
Seb,

Despite the unsuccessful checks, I think this is ready to merge.

Some of the failures were approx_equal failures in which the new solutions were correct to six but not seven decimal places.  I couldn't figure out where to dial back the default degree of precision to six, which is enough for our purposes anyway.  

The other failures were for the "Fast" versions, and were infinite recursion errors.  My guess is that they are related to the two deepcopy commands, make copies of the most recent solution.  However, that solution has a copy of the prior solution, which has a copy of the prior solution, which ... 

Problem with that explanation is that it's hard to see why it wouldn't have caused failures before.  But in any case, when I looked at the "Fast" code there are various ways in which it is out of date and needs to be fixed anyway, so until that can be done it should be removed from the codebase.  (That will be a good place for Alan to start his work for us this summer).

llorracc:
@alanlujan91,

Just saw your other PR for ConsIndShockModelFast.  As long as you're looking at it, could you look into this problem as well?  Explained in my PR notes above.  I suspect that if you take care to copy all elements of the solution EXCEPT the 'solution_next' bit, it may clear up the recursion problem.  

PS.  When I tried running it on my machine, I got errors when it tried

```
from interpolation import interp
```
as we don't seem to have a function called `interp` in interpolation.py.   Something called `interp` does exist in various versions of numpy and scipy, so presumably it is getting it indirectly from there (though why that didn't work on my machine is a mystery).  Anyway, even if it is supposed to work in this indirect way, it would be better to be explicit about where interp is coming from.

alanlujan91:
The line `from interpolation import interp` refers to the [interpolation.py library](https://github.com/EconForge/interpolation.py) so as long as that is installed, it should not be an issue. 

@llorracc @sbenthall @MridulS I have a solution for the issue above, but can't seem to be able to contribute to this PR. Is there a way to do this or do I create my own?

MridulS:
@alanlujan91 you should be able to pull in the changes and push this PR now.

sbenthall:
Two preliminery comments:
 - The use of `stge` to mean "stage" is confusing and it would be easier to use "stage
 - Are there automated test for the new functionality?

MridulS:
There is an unrelated to `assertAlmostEqual` test failure too.
```
_________________________ test_AgentType.test_solve ___________________________
92

93
self = <HARK.tests.test_core.test_AgentType testMethod=test_solve>
94

95
    def test_solve(self):
96
        self.agent.time_vary = ["vary_1"]
97
        self.agent.time_inv = ["inv_1"]
98
        self.agent.vary_1 = [1.1, 1.2, 1.3, 1.4]
99
        self.agent.inv_1 = 1.05
100
        # to test the superclass we create a dummy solve_one_period function
101
        # for our agent, which doesn't do anything, instead of using a NullFunc
102
        self.agent.solve_one_period = lambda vary_1: MetricObject()
103
        self.agent.solve()
104
>       self.assertEqual(len(self.agent.solution), 4)
105
E       AssertionError: 1 != 4
```

PR #1007, opened 2021-05-03
[WIP] updates to ConsIndShockModelFast

alanlujan91:
This PR will document updates to ConsIndShockModelFast.

1. LinearInterpFast is a numba implementation of LinearInterp; see [example notebook](https://github.com/alanlujan91/HARK/blob/updateFast/examples/Interpolation/FastInterpolation.ipynb) for speed comparisons 
2. CubicInterpFast is a numba implementation of CubicInterp; see notebook above


sbenthall:
Is this still WIP @alanlujan91 ?

alanlujan91:
This PR has run into the issue that `jitclasses` are not pickleable. https://github.com/numba/numba/issues/1846

Because HARK makes extensive use of `deepcopy` to copy full solutions and agents, it would not be possible as of now for any solution classes or agent classes to be `jitclasses`. 

@llorracc and @sbenthall, we have seen this before where variable explorer is not able to show some objects that we identified as either `jitclasses` or jitted functions (I can't remember which). It seems `deepcopy` uses a similar process of pickling or serializing objects. 

I'll have to go a different route with updating `ConsIndShockModelFast`.

PR #1019, opened 2021-06-18
use Lognormal.from_mean_std in RiskyAsset forward simulation

### LINKED ISSUE #1003: Better documentation and use of lognormal parameters
The Distribution.LogNormal class takes two parameters, `mu` and `sigma`.

These are the mean and standard deviation for the _underlying normal_ distribution:

https://github.com/econ-ark/HARK/blob/master/HARK/distribution.py#L89

https://numpy.org/doc/stable/reference/random/generated/numpy.random.lognormal.html

This could be made more clear in the documentation.

There is a separate method for creating the lognormal from the values of _its own_ mean and standard deviation.

https://github.com/econ-ark/HARK/blob/master/HARK/distribution.py#L202

This latter method is used in `update_RiskyDstn`:

https://github.com/econ-ark/HARK/blob/master/HARK/ConsumptionSaving/ConsPortfolioModel.py#L209-L249

RiskyDstn is used in the solver.

But the true mean/sigma factory method is not used in `get_Risky`:

https://github.com/econ-ark/HARK/blob/master/HARK/ConsumptionSaving/ConsPortfolioModel.py#L339-L367

which is used for drawing the shocks in forward simulation.

This is a bit confusing.

sbenthall:
fixes #1003

- [x ] Updated documentation of features that add new functionality.

### DIFF
--- a/Documentation/CHANGELOG.md
+++ b/Documentation/CHANGELOG.md
@@ -21,6 +21,7 @@ PR: [#832](https://github.com/econ-ark/HARK/pull/832). See [this forthcoming REM
 
 #### Minor Changes
 
+* Using Lognormal.from_mean_std in the forward simulation of the RiskyAsset model [#1019](https://github.com/econ-ark/HARK/pull/1019)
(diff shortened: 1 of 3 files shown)


Mv77 (approved):
Looks great to me

PR #1011, opened 2021-05-06
implement CubicHermiteInterp

alanlujan91:
CubicHermiteInterp is a HARK compatible wrapper of scipy's CubicHermiteSpline

See notebook for discussion: https://github.com/alanlujan91/HARK/blob/scipyCubic/examples/Interpolation/CubicInterp.ipynb

Please ensure your pull request adheres to the following guidelines:

- [ ] Tests for new functionality/models or Tests to reproduce the bug-fix in code.
- [ ] Updated documentation of features that add new functionality.
- [ ] Update CHANGELOG.md with major/minor changes.

PR #1020, opened 2021-06-18
Additional Features for CubicHermiteInterp

alanlujan91:
This PR adds new methods to `CubicHermiteInterp`. 

Because `CubicHermiteInterp` uses `scipy`'s `CubicHermiteSpline`, we can easily port additional methods that could be useful in HARK. 

These were initially in #1011 but were moved to a standalone PR for further discussion. 

These new methods are:

1. `der_interp(self[, nu])` Construct a new piecewise polynomial representing the derivative.
2. `antider_interp(self[, nu])` Construct a new piecewise polynomial representing the antiderivative.
3. `integrate(self, a, b[, extrapolate])` Compute a definite integral over a piecewise polynomial.
4. `roots(self[, discontinuity, extrapolate])` Find real roots of the the piecewise polynomial.
5. `solve(self[, y, discontinuity, extrapolate])` Find real solutions of the the equation pp(x) == y.

See notebook for more details: https://github.com/alanlujan91/HARK/blob/CHI_new_feats/examples/Interpolation/CubicInterp.ipynb


sbenthall:
@alanlujan91 can you fix the merge conflict? (sorry for duplicating my comment on #1011 )

@mnwhite would you be able to review this?

mnwhite:
I'm working on this now. GitHub flagged a bunch of things as "conflicts" that plainly were not. And it also got upset about notebook execution metadata, which is annoying. It's quite likely that I broke stuff in that commit.

mnwhite:
This has one failing test, and I'm actually not sure whether it *should* pass. It looks like the test is for whether new cubic interpolator can accepts inputs of shape (N,1) rather than just (N,)-- flat 2D arrays. Is that intended?

mnwhite:
The problem was some `self`s that were removed. The init code does a "flat check", but its results weren't be used.

PR #1023, opened 2021-06-22
Update ConsPortfolioModel.py

alanlujan91:
Reading through the solver method for `ConsPortfolioModel` I came across code that was split up by the `black` formatter in an awkward way that made it more difficult to understand. 

This PR reorganizes and reformats the solver code to increase readability.


llorracc:
Merged because the tests that failed were not the result of problems with Alan's code but caused by an update to NumPy which changed fine details of matrix multiplication.

sbenthall:
I would advise against trying to wrestle with the code formatter now that we've agreed on one.

Since we are supposed to be applying the code formatter whenever we do a PR, this will eventually get reverted back to how it was.

alanlujan91:
@sbenthall this reorganization is formatter-consistent. Re-applying `black` on my end makes no changes to it.

Issue #1015, opened 2021-06-03
Allow time-varying UnempPrb


Mv77:
`construct_lognormal_income_process_unemployment` in `ConsIndShockModel.py` currently only accepts constant unemployment probabilities which can be different while working and while retired, `UnempPrb` and `UnempPrbRet`.

This setup does not allow one to implement, for instance:
- _Age-specific unemployment probabilities while working_ (I think it is the case that unemployment is much higher for young people, at least in my native Colombia).
- _Age-varying but risk-free retirement income_: most papers assume that retirement income is constant, but there are exceptions. Cagetti (2003), for instance, has a downward trend in retirement income that is still safe. A way to achieve this in HARK is to ignore the built-in retirement functionality, choose the appropriate `PermGroFac` sequence and make `PermShkStd = TranShkStd = 0` after some age. The issue is that one can not shut down unemployment after some age.

Thus, I think a time-varying `UnempPrb` would be a nice feature to have.

sbenthall:
I completely agree. I was just running into a related issue with the `Adjust` shocks.

I think that rather than having the age-varying nature of these variables be hard coded into the models, it should be pulled out into an age-varying distribution class. I have ticketed this in #1017 

I'm not 100% sure that the 'age-varying' use case is the same as the 'time-varying' one here. Can you confirm? It seems that sometimes this time-varying functionality is being handled in separate ways by the simulation and solution code, though it's really a model element.

Mv77:
Hey,

Yes, I do think they are the same use case. I usually interpret the models annually but what I meant was an unemployment probability that can change from period to period (whatever a time period represents).

A time-varying distribution class might be useful for finite-horizon models. A suggestion if you do implement it is for it to have the capacity to produce the time-specific distributions easily, so that they can still be passed to the solvers without passing the whole age-varying distributions.

Mv77:
I do agree that these case-handling bits of code where one might run through agents checking their `t_cycle` and drawing from different distributions is repeated throughout HARK, so a tool to deal with it would be nice.

sbenthall:
Perfect. Let me see if I can get a good PR for this by Thursday.

sbenthall:
Now that #1018 is in, this might be easier.

mnwhite:
I think HARK already handles this.

Issue #620, opened 2020-04-09
define distributions directly in initial parameters for models


sbenthall:
Related to #227 

Currently, when the parameters of distributions are given to models, they are listed as unstructured fields whose meaning depends on their name, e.g.: `aNrmInitMean`. `aNrmInitStd`. `pLvlInitMean`. `pLvlInitStd`, ...

It can be changed so that these parameters look like: `aNrmInit : LogNormal(0, 1)`.

sbenthall:
This is related to #664 , because these distributions are often '"time-varying".

It looks like the expectation is that `TranShkStd` is a list even when there's a single value and so is "time-varying" even though it is static across time.

https://github.com/econ-ark/HARK/blob/master/HARK/ConsumptionSaving/ConsIndShockModel.py#L2020-L2023

I don't see any reason for this except for haste in the early coding.

sbenthall:
This feature would make HARK closer to Dolo.
Dolo requires users to specify the distribution or process driving values of exogenous shocks in configuration.
E.g.
https://github.com/EconForge/dolo.py/blob/master/examples/models_/rbc.yaml#L72-L73

llorracc:
No, this is deliberate.  

Remember that IndShockConsumerType is used for finite horizon (life cycle) models as well as infinite horizon models.  

There is a lot of variation by age in the magnitude of income shocks, both transitory and permanent.  Actually, that age-related variation can have important implications for things like the profile of stockholding patterns by age.  So it is deliberate that the default is to allow TranShkStd and PermShkStd to vary with time/age.

sbenthall:
I see now I've conflated two things in this ticket.

- The original ticket is specifically for the **Initial** conditions for an agent. It's "birth values". These are not time-varying.
- Yes, of course, allowing time and age-varying income is an important feature and deliberate.
- My comment [here](https://github.com/econ-ark/HARK/issues/620#issuecomment-813072820) was that it looks a little funny to use a list for a singleton, non-time-varying value. I'm afraid I've muddied the water by bringing it into this ticket. This design choice seems to be inconsistent across the [library](https://github.com/econ-ark/HARK/issues/890#issuecomment-749720730).
- My recommendation, based on the design goal to get things closer to Dolo, is to create a Process class for time-varying exogenous values.

So, rather than giving a model:

```
init_cyclical['PermShkStd'] = [0.1, 0.1, 0.1, 0.1]
[...]

PR #1024, opened 2021-06-23
Rework income process using IndexDistribution

Mv77:
@sbenthall recently introduced `IndexDistribution` to deal (among other things) with distributions that time-varying indices.

Seb, this is a first attempt at using that class in the income-process constructor. The current code works and I quite like the change, I'd like your feedback.

The big issue is that, as we discussed before, we need to think a bit more about what is the appropriate thing to do about RNG seeds with this change. I've not taken care of it and that's why tests are failing.


Mv77:
TODOS:
- [ ] Handle seeds appropriately.
- [ ] Replace the code that draws from the income distribution to make use of the new functionality.
- [ ] Docstrings and linting.

HARK/ConsumptionSaving/ConsIndShockModel.py
```diff
@@ -2659,77 +2660,72 @@ def construct_lognormal_income_process_unemployment(self):
...
+
+def TranShk_engine(sigma, UnempPrb, IncUnemp, n_approx, seed = 0):
+
+    TranShkDstn = MeanOneLogNormal(sigma, seed = seed).approx(n_approx, tail_N=0)
+    if UnempPrb > 0:
+        TranShkDstn = add_discrete_outcome_constant_mean(TranShkDstn, p=UnempPrb, x=IncUnemp)
+
+    return TranShkDstn
+
+def InkShk_engine(sigma_Perm, sigma_Tran, n_approx_Perm, n_approx_Tran, UnempPrb, IncUnemp, seed = 0):
```
sbenthall, line 2721: I believe this should be IncShk ... with a 'c'

HARK/ConsumptionSaving/ConsIndShockModel.py
```diff
@@ -2659,77 +2660,72 @@ def construct_lognormal_income_process_unemployment(self):
...
-                seed=self.RNG.randint(0, 2 ** 31 - 1),
-            )
+            normal_length = T_retire
+            retire_length = T_cycle - T_retire
+        else:
+            normal_length = T_cycle
+            retire_length = 0
+
+        UnempPrb_list = [UnempPrb]*normal_length + [UnempPrbRet]*retire_length
+        IncUnemp_list = [IncUnemp]*normal_length + [IncUnempRet]*retire_length
```
sbenthall, line 2682: Do you see a clear path from this to a solution to #1015 ?

HARK/ConsumptionSaving/ConsIndShockModel.py
```diff
@@ -2659,77 +2660,72 @@ def construct_lognormal_income_process_unemployment(self):
...
-                PermShkDstn.append(PermShkDstn_t)
-                TranShkDstn.append(TranShkDstn_t)
-        return IncShkDstn, PermShkDstn, TranShkDstn
+    PermShkDstn = MeanOneLogNormal(sigma, seed = seed).approx(n_approx, tail_N=0)
+
+    return PermShkDstn
+
+def TranShk_engine(sigma, UnempPrb, IncUnemp, n_approx, seed = 0):
+
+    TranShkDstn = MeanOneLogNormal(sigma, seed = seed).approx(n_approx, tail_N=0)
```
sbenthall, line 2715: What I would like to do at some point soon is make the IndexDistribution.approx method return an IndexDistribution.
Maybe that can help streamline these engine processes further, automatically mapping Distribution operations over them.

sbenthall:
Yes, thank you, I like this.

There is an additional failing test:

```
    def setUp(self):
        # Make one agent type and an economy for it to live in
        self.agent = AggShockMarkovConsumerType()
        self.agent.cycles = 0
        self.agent.AgentCount = 1000  # Low number of simulated agents
>       self.agent.IncShkDstn[0] = 2 * [self.agent.IncShkDstn[0]]  ## see #557
E       TypeError: 'IndexDistribution' object does not support item assignment
```

I didn't write a `__setitem__` method for the IndexDistribution class. One could be easily added.

https://www.geeksforgeeks.org/__getitem__-and-__setitem__-in-python/

The seed logic in HARK is quite complex and I'm not sure I fully understand all the motivations for it.

HARK/ConsumptionSaving/ConsIndShockModel.py
Mv77: True, thank you!

HARK/ConsumptionSaving/ConsIndShockModel.py
Mv77: Yes, absolutely. That was my motivation for doing this, and it would amount to simply allowing users to pass these lists (as they do for the standard deviations) instead of floats. I just wanted to get what is currently there updated before introducing that feature.

Mv77:
I am reviving this and will work on it during the next few weeks. It is part of a broad goal of making the income-process construction code more flexible. However, the next step is to update the current code using newer tools that have been added, like `IndexDistribution` and `calc_expectations` (both of which I think are awesome).

A big issue with this kind of update is the RNG. For reasons yet to be determined (maybe the order in which shocks are drawn or some seed that needs to be passed) using `IndexDistribution` changes the shocks that are drawn by tests. Being the income process that pretty much every AgentType in HARK inherits, this change makes almost all of the simulation tests fail.

Before I dive into it, I want to summon the RNG maestros to ask if they have any idea or opinion on how this issue should be resolved. Do I edit the tests to check for the new values? Do I tweak the new implementation until matches the current tests? Do I look for a happy middle?

Have the team's views on simulation tests and how they should be done changed?

sbenthall:
Hooray. Glad you're looking into this.

I don't think this  addresses the whole issue, but one way to tackle the problem is to rewrite tests so that rather than targeting a specific numerical value, they tests a functional relationship:

Here's an example:
https://github.com/econ-ark/HARK/blob/master/HARK/ConsumptionSaving/tests/test_ConsPortfolioFrameModel.py#L81-L84

This is actually a much better test, since it tests the functionality of the code, and not some accidental combination of the code and a random seed.

Mv77:
That test is evaluating a property of the solution. I think we have a fair bit of those.

I still see value in testing simulation results. Imagine for instance that some indexing error in SimOnePeriod made us draw the wrong shocks. Or that, as it happened 2 years ago, some handling of a seed makes agents always draw the same shock. Tests to catch that are great. So I am happy that we have both kinds, but I wonder if in your discussions there has been any development in how we think we should carry out simulation tests going forward.

Mv77:
Something reassuring is that in the current state of the PR, tests that deal with the solution itself (not the simulation history) pass, so we know the current machinery is leaving the distributions that the agents get in the solution process unchanged. We know the miss-matches are generated when we draw from the `IndexDistribution`s as opposed to the lists of discrete distributions currently in `master`.

sbenthall:
These are great points.

This is an example of a test of simulation functionality that does not depend on the random seed:
https://github.com/econ-ark/HARK/blob/master/HARK/ConsumptionSaving/tests/test_ConsPortfolioFrameModel.py#L69-L74

I am not aware of any new consensus for doing simulation tests.
But automated testing wisdom would say that it's always better to test for a specific unit of functionality than to test for a general 'correct answer' from a complicated system. The current way of testing simulations -- just testing for some predetermined output -- is a lazy way of writing tests. I wrote most of them, so I know how lazy they are. This problem with the random seeds exposes the cost of that laziness.

sbenthall:
Is it possible that there is an error in IndexDistribution though?
If so, a new unit test for IndexDistribution might catch it.

Mv77:
> Is it possible that there is an error in IndexDistribution though? If so, a new unit test for IndexDistribution might catch it.

I am 98% sure that:
- There is no bug in terms of representing different distributions from the ones we want to represent. If that were the case we would get errors in tests that deal with model solutions.
- It does seem to change the order or seeds with which shocks are drawn, resulting in different simulation histories even if the distributions we are drawing from are the same as before. That's why we are getting failures in the simulation tests.

Mv77:
@sbenthall I have a clue of what might be happening.

If you run
```
from HARK.distribution import IndexDistribution, Lognormal, MeanOneLogNormal
import numpy as np

n_app = 5
sigmas = [0.1,0.2]

RNG = np.random.RandomState(1)

a = []
for s in sigmas:
    a.append(MeanOneLogNormal(s, seed = RNG.randint(0, 2 ** 31 - 1)))

a_draws = []
for d in a:
    a_draws.append(d.draw(1))

print(a_draws)

b = IndexDistribution(engine = MeanOneLogNormal, conditional  = {'sigma': sigmas}, seed = 1)

b_draws = []
for d in b:
    b_draws.append(d.draw(1))

print(b_draws)
```

You will find that `a_draws` and `b_draws` have the exact same shock draws.

Notice:
- Approach "a" is creating a random-number generator with seed 1, then using it to draw the seeds of the time-specific distributions.
- Approach "b" creates an index distribution with `seed=1` and then draws from the time-specific distributions.

The key of why these two approaches give the same result is that `IndexDistribution` creates its own internal random-number generator with the given seed
https://github.com/econ-ark/HARK/blob/f338c976c813991a095a0d3fea9967232ea59e07/HARK/distribution.py#L67
https://github.com/econ-ark/HARK/blob/f338c976c813991a095a0d3fea9967232ea59e07/HARK/distribution.py#L19-L20
and then uses that random-number generator to draw the seeds of the time-specific distributions
https://github.com/econ-ark/HARK/blob/f338c976c813991a095a0d3fea9967232ea59e07/HARK/distribution.py#L72-L78

Thus, if I represent the time-varying income distribution with an `IndexDistribution`, there will be `t+1` random-number generators:
- A "master" one, that will draw the seed of the time-specific distributions.
- t time-specific ones.

The difference is that under the current approach the seeds for the time-specific distributions are drawn directly from the agent's RNG
https://github.com/econ-ark/HARK/blob/f338c976c813991a095a0d3fea9967232ea59e07/HARK/ConsumptionSaving/ConsIndShockModel.py#L2737-L2742

Thus, without `IndexDstn` we have
`Agent's RNG --(draws seeds for)--> Time-specific RNGs` 
and with `IndexDstn` we have
`Agent's RNG --(draws seed for)--> IndexDstn's RNG --(draws seeds for)--> Time-specific RNGs`

The extra step must make simulations not exactly match (but it might not be the only thing).

Mv77:
A solution might be to let `IndexDstn` (optionally) receive an RNG from which to draw time-specific seeds, instead of creating its own.

Mv77:
@sbenthall, I'd like to ask why do you think `IndexDistribution` must inherit from the `Distribution` class.

I think the above issue is being generated by the fact that the current implementation thinks of the `IndexDistribution` as a random variable in itself. I wonder what you would think of a re-interpretation in which `IndexDistribution` is not itself a distribution but a way to organize time-varying distributions.

Mv77:
> A solution might be to let `IndexDstn` (optionally) receive an RNG from which to draw time-specific seeds, instead of creating its own.

@sbenthall Just confirmed this (mostly) works.

image

Passes all of IndShock's tests except Harmenberg.

Mv77:
This PR finally passes tests! I managed to resolve all the simulation discrepancies.

It does kick the can of simulation tests down the road, but there is already an issue for that [#1105](https://github.com/econ-ark/HARK/issues/1105) and I feel getting to an unified solution and a single fix in a different PR would be better. It was also satisfying to get it to match: I now know there is no bug in `IndexDstn`.

The actual change that is introduced by this pr is not that transcendental: it simply uses `IndexDstn` for the income process. However, this innovation makes it much easier to flexibilize the income process with things like time-varying unemployment, which is not currently supported (and I will need and implement soon).

I am re-requesting review.

sbenthall:
This is looking very good.

There's one change I'd suggest.

According to its documentation, the `engine` parameter of the `IndexDistribution` should be of type `Distribution`. You have made standalone methods (e.g. `IncShk_engine`) for your distribution engines. (I believe I've done this in my own code elsewhere).

I wonder whether it would make sense to make these engine methods into model-specific subclasses of Distribution. Basically, you would just be putting the substance of your engine methods into the `__init__` method of the class.

Mv77:
Thanks Seb,

Just to verify that I follow, you'd like me to move my ***Shk_engine methods to the inside of the Agent type? so that they are called by self.***Shk_engine?

I think that'd make sense, since these are specifically tailored for the given model.

sbenthall:
No, that's not what I mean.

I mean something more like:

```
class IncomeShockDistribution(Distribution):

    def __init__(self,
        sigma_Perm,
        sigma_Tran,
        n_approx_Perm,
        n_approx_Tran,
        UnempPrb,
        IncUnemp,
        seed=0,):
        ...

        IncShkDstn = IndexDistribution(
            engine=IncomeShockDistribution,
            conditional={
                "sigma_Perm": PermShkStd,
                "sigma_Tran": TranShkStd,
                "n_approx_Perm": PermShkCount_list,
                "n_approx_Tran": TranShkCount_list,
                "UnempPrb": UnempPrb_list,
                "IncUnemp": IncUnemp_list,
            },
            RNG=self.RNG,
        )
```

sbenthall:
To elaborate on this @Mv77 ...

The single-period income distributions are used by several models (in the current HARK architecture, via inheritance). But where I think we should move to is that these distributions should be define in model _configuration_, not hard coded into model code. See issue #620 

In the FramedAgent code, this is precisely what happens: the `PermShk` frame takes, as its 'transition function', a Distribution.

https://github.com/econ-ark/HARK/blob/master/HARK/ConsumptionSaving/ConsPortfolioFrameModel.py#L193-L206

That is is then used by the simulation code to know that that variable must be sampled as a shock.

https://github.com/econ-ark/HARK/blob/master/HARK/frame.py#L257-L265

At the very least, if you defined the IncShk engine in terms of classes that are subclasses of Distribution (or, DiscreteDistribution, which may be more appropriate), then I could use that same distribution definition in the PortfolioFrameModel. That's an indication of how this would support code reuse.

Mv77:
Thanks Seb.

I initially thought of this as type puritanism!

But the more I think of it, the more it makes sense. It pushes us in the direction of income distributions being their own object, modular entities easily switched around. I am fully onboard; will revise over the weekend.

Mv77:
Not that type puritanism is bad; it is just one of those things that have a huge social return but are annoying to enforce in one's own life.

Mv77:
@sbenthall 

Seb, I've now made engines their own class. Tests still pass. This is ready for your review again.

Thank you!

llorracc:
A small suggestion:  Many different choices could be made for the `IncomeShockDistribution.`  The choices here are highly specific: Transitory and Permanent shocks, with a transitory unemployment spell.  As long as we are inventing a new name anyway, let's call it something more specific, like `FriedmanBufferStockShockDistribution` or `TranPermWithIIDUnempShockDistribution` or `BufferStockTheoryShockDistribution.`  (Or perhaps substituting `Shk` For `Shock` in accord with guidelines in the NARK.   And maybe `Dstn` for `Distribution`, to prevent the name of the object from being too long.

Mv77:
Agreed. I've now gone with
- Permanent income shocks: `LognormPermIncShk`.
- Transitory income shocks: `MixtureTranIncShk`.
- Joint permanent +  transitory income shock specification: `BufferStockIncShkDstn`.

sbenthall:
Great work, @Mv77 . Thanks for the careful design and implementation. :)

Mv77:
Thank you for the reviews! :)

Issue #1022, opened 2021-06-22
Consistent t-index for time varying shock parameters.


sbenthall:
Referring to #1021 and #326

Parameters for a time-varying income process are entered like this:

```
# Parameters that specify the income distribution over the lifecycle
    "PermShkStd" : [0.1,0.2,0.1,0.2,0.1,0.2,0.1,0,0,0],
    "PermShkCount" : 7,                    # Number of points in discrete approximation to permanent income shocks
    "TranShkStd" : [0.3,0.2,0.1,0.3,0.2,0.1,0.3,0,0,0],
    "TranShkCount" : 7,                    # Number of points in discrete approximation to transitory income shocks
```

But against what one might expect, the 0th element of the list is interpreted as the shock that occurs at the 0th as well as 1th periods. The 1th element of the list corresponds to the 2th shock, etc.

This not a natural way to define a model and is not very well documented; it is done this way because of some logic imported from earlier solution code. That code should be ammended.

sbenthall:
This is the backwards induction code that iterates back through the time steps.

https://github.com/econ-ark/HARK/blob/master/HARK/core.py#L1006-L1034

What it's doing is:

- Computing `T` as the length of the time-varying parameter lists
- Iterating with `t` from 0 to (T-1). Let `k = T - 1 - t`, so `k` is ranging from (T-1) back down to 0. This is a backwards index for each time-varying value.
- Get the one-period solver for `k`.
- Get the properties of the agent at `k` needed for the solver, including the IncShkDstn, etc.
- compute and store the solution.

The resulting list of solutions gets prepended to a list containing the terminal solution, which will be total of (T + 1) in length.

https://github.com/econ-ark/HARK/blob/master/HARK/core.py#L895-L902

The finite model is thus defined over (T + 1) periods, but with the first period omitted from list of T time-varying parameters.

To make the corrections recommended by this issue, we need to:

- For every time-varying default parameter list, prepend a 0th element equal to its first, which will cover the 'newborn' case.
- In the `solve_one_cycle` code, let `k` range from T to 1.

Everything should then work as before and the weird newborn handling questioned in #326 can then be removed.

sbenthall:
Ah, it is not quite so simple.

In the case of a cyclical model, with, say `T = 4`, then this buffering will not work. 

@mnwhite  I'm looking at this case:

```
# Make a dictionary to specify an infinite consumer with a four period cycle
init_cyclical = copy(init_idiosyncratic_shocks)
[...]

PR #1027, opened 2021-06-25
[WIP] Consistent t-indexing for time varying parameters

sbenthall:
This PR works towards addressing #1022, and along with with #326.

The main change to the core logic is that when a model with time-varying parameters of length T is given to `solve_one_cycle`, the solutions in the cycle are solved backwards and given inputs drawn from the following state, which is (t+1).

For truly cyclic models, T solutions are created. The T'th solution (which is solved first) draws is parameters from the 0th period.

For "lifecycle" models, where the "cycle" represents a series of steps executed only once, (T - 1) solutions are created. The Tth solution (which is solved first), which corresponds to the _pen_ultimate period, draws parameters from the last period (whose terminal solution is computed elsewhere).

Further changes to the library entailed by this change include:
 - Adding 0th period values to all of the default parameters lists that are non-cyclical.
    - This includes the Calibration/IncomeTools lists. I ask @Mv77 to review my changes here to see if they make sense.
 - "Rotating" all of the cyclical time-varying parameter lists "forward" one step.
 - Removing the special newborn handling code from ConsIndShockModel (see #1021 and #326 ) `get_shocks()`. Because this newborn code created an extra `draw()` for _every agent_, changing the distribution's random seed accordingly, changing this will change all numerical values from simulations. So this will require a thorough code review. I request that from @mnwhite

What I should not be getting, which I hope to correct for through further work on this PR, are changes to the solution objects.

Some open questions on my end are:
 - How should the ConsMarkovModel be adjusted to handle this new schema? I see a failing test for a structural (non-numeric) reason but I'm not sure how 
 - Why am I getting a failing test for the ConsIndShockModelFast? Maybe @alanlujan91 can tell me

If at all possible, here is what I'm hoping NOT to do in this PR:
 - Resolve the question of how time-varying parameters that are not "look forward" parameters are passed to the solver. I don't think anything I'm doing here is making that _harder_. But it sounds like that requires a lot of model-specific knowledge.
 - Touching any of the handling of the terminal and pseudo-terminal solutions.

While there are models that may require changes along these lines, I'm not sure any of the existing models do. I would recommend not trying to support a feature that does not have a model and test using it in the library.

### DIFF
--- a/Documentation/CHANGELOG.md
+++ b/Documentation/CHANGELOG.md
@@ -14,6 +14,7 @@ Release Data: TBD
 
 #### Major Changes
 
+* A IndexDistribution class for representing time-indexed probability distributions [#1018](https://github.com/econ-ark/pull/1018/).
 * Adds new consumption-savings-portfolio model `RiskyContrib`, which represents an agent who can save in risky and risk-free assets but faces
 frictions to moving funds between them. To circumvent these frictions, he has access to an income-deduction scheme to accumulate risky assets.
 PR: [#832](https://github.com/econ-ark/HARK/pull/832). See [this forthcoming REMARK](https://github.com/Mv77/RiskyContrib) for the model's details.
--- a/Documentation/CHANGELOG.md
+++ b/Documentation/CHANGELOG.md
@@ -21,6 +21,7 @@ PR: [#832](https://github.com/econ-ark/HARK/pull/832). See [this forthcoming REM
 
 #### Minor Changes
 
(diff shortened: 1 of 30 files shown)


sbenthall:
Some of the failing tests were due to problems resolved by #1029 

I'll do a pass through the remaining test failures trying to fix anything which I think isn't due to the changing newborn handling.

sbenthall:
This PR got messy. I'm going to start again with a clean cut off master

sbenthall:
Nevermind I've found where some code that might be responsible for the failing tests is.

Related to https://github.com/econ-ark/HARK/pull/981

Issue #1009, opened 2021-05-06
Reducing pytest approx_equals sensitivity to only 6 decimal places


sbenthall:
From @llorracc in #1006 

> Some of the failures were approx_equal failures in which the new solutions were correct to six but not seven decimal places. I couldn't figure out where to dial back the default degree of precision to six, which is enough for our purposes anyway.

sbenthall:
Any ideas @MridulS ?

llorracc:
@MridulS , 

Renewing this because -- as @alanlujan91 seems to have figured out -- a new version of numpy seems to change some very fine details of matrix multiplication that affect things way down in like the 14th digit.  Could you look into whether there is a global way to turn down the sensitivity of the assertEqual thing?  In the meantime I'll ask @alanlujan91 to see if reducing the length of a few of the asserted numbers fixes the problem.

Mv77:
Seems not to be matrix multiplication but geometric distributions or something like that.

alanlujan91:
I believe it's due to this change although I could be wrong

 https://numpy.org/devdocs/release/1.21.0-notes.html#generator-rayleigh-and-generator-geometric-performance-improved

mnwhite:
Well that will cause *everything* to fail tests just about. What we should
have done is had a trivial RNG test that *only* makes draws from numpy
distributions. If those ever fail, it means the RNG system has changed and
we need to re-compute our test targets (in their own PR).

sbenthall:
We should pin the numpy dependency to the prior version so tests pass and development is not blocked.

Then we can update numpy and correct the tests in one clean PR.

sbenthall:
Please see #1029

Because pinning numpy to v1.19.4 did not clear the tests by itself, that disconfirms the hypothesis that the failure was due to a recent numpy RNG update.

My guess is that the underlying cause is GitHub infrastructure changes.

Changing a few of the tests from `assertEquals` to `assertAlmostEquals`, which bounds the number of significant digits of agreement, solved the immediate issue of test failures.

This does not address the broader problem of this issue, to which #1025 is scoped.

sbenthall:
`assertAlmostEquals` takes an optional `places` argument that can change the number of significant digits the test is sensitive to.

https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertAlmostEqual

[...]

PR #1029, opened 2021-06-25
freezing numpy to 1.19.4

sbenthall:
This PR pins the HARK numpy dependency to 1.19.4

This is a version of numpy before recent changes to the random number generator that are causing test failures.

See: https://github.com/econ-ark/HARK/issues/1009#issuecomment-867049221

If tests pass, I recommend merging this. It will prevent other PRs from blocking based on spurious numerical test failures.

Then we can be more deliberate about updating to a more recent numpy and making changes to the tests in an isolated, disciplined way.


sbenthall:
The three failing tests all used `assertEquals` with a very long decimal number.

As GitHub Actions runs tests on a variety of virtualized operating systems and Python versions, it is possible that the (tiny) discrepancies are due to changes in GitHub's infrastructure.

I've replaced these problematic calls with `assertAlmostEquals`, which has some reasonable cutoff for numerical equivalence.

sbenthall:
This is ready to merge. I think it would be best to do it ASAP so other work is not blocked.

PR #1031, opened 2021-06-28
move 'cycles' from agent constructor argument to parameter

### LINKED ISSUE #1030: no default cycles value
From #1022 , @mnwhite says:

> Yeah, cycles=1 should be removed as a default argument. I think I wrote
> that in the very earliest versions (late summer / early fall 2015) and it's
> never come off.

sbenthall:
Fixes #1030 


sbenthall:
ready for review

mnwhite:
Yes, this is what we want.

PR #1033, opened 2021-06-28
[WIP] Use calc_expectation in ConsPortfolioModel

alanlujan91:
Using `calc_expectation` improves solution speed. `test_ConsPortfolioModel.py` runtime:
- w/o `calc_expectation`: 50.05s
- w `calc_expectation`: 27.89s

### DIFF
--- a/HARK/ConsumptionSaving/ConsPortfolioModel.py
+++ b/HARK/ConsumptionSaving/ConsPortfolioModel.py
@@ -536,7 +536,7 @@ def solveConsPortfolio(
             return b_nrm / (PermGroFac * shocks[0]) + shocks[1]
 
         # Evaluate realizations of marginal value of market resources next period
-        def dvdm_next(shocks, b_nrm):
+        def dvdb_dist(shocks, b_nrm):
             mNrm = m_nrm_next(shocks, b_nrm)
             mNrm_tiled = np.tile(np.reshape(mNrm, (bNrm_N, 1)), (1, Share_N))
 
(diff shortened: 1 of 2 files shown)


alanlujan91:
Well it passes the tests but not the examples in the notebook, which probably means the examples need to be added to the tests. Will keep working on this later.

llorracc:
@alanlujan91,

Thanks for this PR - esp given that calc_expectation is not so easy to work with.  Probably it is the test that needs to be changed, not your PR.  Maybe it's just another example of the excessive stringency of our tests.  @MridulS is working on figuring out how to deal with that, if this is such an example maybe it is a good test case for whether his proposed solution works ...

MridulS:
I haven't looked at the changes in detail yet but the this also breaks the `testPortfolioConsumerTypeSticky.test_sticky` test, not just the notebook examples. Do they pass locally?

Does this pass locally?
``` bash
pytest HARK/ConsumptionSaving/tests/test_ConsPortfolioModel.py
```

alanlujan91:
> I haven't looked at the changes in detail yet but the this also breaks the `testPortfolioConsumerTypeSticky.test_sticky` test, not just the notebook examples. Do they pass locally?
> 
> Does this pass locally?
> 
> ```shell
> pytest HARK/ConsumptionSaving/tests/test_ConsPortfolioModel.py
> ```

@MridulS I just added the test `testPortfolioConsumerTypeSticky.test_sticky` which is what was failing on the notebook. While before it was passing all tests, now that I've added `test_sticky` it doesn't pass it. I added this test to make sure any future changes I make pass this test, since I don't know how to rigorously test all notebooks.

alanlujan91:
@llorracc this PR is failing an actual use case, when `AdjustPrb < 1.0`. I've looked over the code a few times and I can't figure out why it would suddenly fail. This might have to do with how I am using `calc_expectation` with 2D functions `(a,share)`. Are there examples on the codebase where `calc_expectation` is used with 2D functions? That might help me figure out this issue.

alanlujan91:
@Mv77 could you look at this when you have time and provide suggestions?

alanlujan91:
finally figured out my bug, this PR now passes all tests and examples!

PR #1039, opened 2021-07-09
fixing t indexing for inputs for cyclical models

### LINKED ISSUE #1022: Consistent t-index for time varying shock parameters.

sbenthall:
This is PR partially fixes #1022.

It makes the changes for _cyclical_ models only. This involves:
 - rotating the input parameters so that the "last" element, which corresponds to the parameters of the 0th period, goes to the 0th place.
 - changing the solver to iterate backwards through these elements properly
 - using the t'th LivPrb in mortalilty for _this case_.

... it should also involve using t not `t - 1` when drawing shocks but ... we don't seem to have automated tests for a multi-period cyclical case and so they are not failing.

Because this is a pretty complicated change, I'm going to try to fix it for the cyclical cases first and then do the cycles = 1 cases in a separate PR.

### DIFF
--- a/HARK/ConsumptionSaving/ConsIndShockModel.py
+++ b/HARK/ConsumptionSaving/ConsIndShockModel.py
@@ -1729,7 +1729,7 @@ class PerfForesightConsumerType(AgentType):
         # Determine who dies
         DiePrb_by_t_cycle = 1.0 - np.asarray(self.LivPrb)
         DiePrb = DiePrb_by_t_cycle[
-            self.t_cycle - 1
+            self.t_cycle - 1 if self.cycles == 1 else self.t_cycle
         ]  # Time has already advanced, so look back one
 
         # In finite-horizon problems the previous line gives newborns the
@@ -3047,7 +3047,7 @@ init_lifecycle.update({"LivPrb": liv_prb})
 
 # Make a dictionary to specify an infinite consumer with a four period cycle
 init_cyclical = copy(init_idiosyncratic_shocks)
-init_cyclical['PermGroFac'] = [1.082251, 2.8, 0.3, 1.1]
+init_cyclical['PermGroFac'] = [1.1, 1.082251, 2.8, 0.3]
 init_cyclical['PermShkStd'] = [0.1, 0.1, 0.1, 0.1]
 init_cyclical['TranShkStd'] = [0.1, 0.1, 0.1, 0.1]
 init_cyclical['LivPrb'] = 4*[0.98]
--- a/HARK/ConsumptionSaving/ConsIndShockModel.py
+++ b/HARK/ConsumptionSaving/ConsIndShockModel.py
@@ -2173,12 +2173,17 @@ class IndShockConsumerType(PerfForesightConsumerType):
         newborn = self.t_age == 0
         for t in range(self.T_cycle):
             these = t == self.t_cycle
+
+            # temporary, see #1022
+            if self.cycles == 1:
+                t = t - 1
+
             N = np.sum(these)
             if N > 0:
                 IncShkDstnNow = self.IncShkDstn[
-                    t - 1
+                    t
                 ]  # set current income distribution
-                PermGroFacNow = self.PermGroFac[t - 1]  # and permanent growth factor
+                PermGroFacNow = self.PermGroFac[t]  # and permanent growth factor
                 # Get random draws of income shocks from the discrete distribution
                 IncShks = IncShkDstnNow.draw(N)
 
--- a/Documentation/CHANGELOG.md
+++ b/Documentation/CHANGELOG.md
@@ -14,6 +14,7 @@ Release Data: TBD
 
 #### Major Changes
 
+* Input parameters for cyclical models now indexed by t [#1039](https://github.com/econ-ark/HARK/pull/1039)
 * A IndexDistribution class for representing time-indexed probability distributions [#1018](https://github.com/econ-ark/pull/1018/).
 * Adds new consumption-savings-portfolio model `RiskyContrib`, which represents an agent who can save in risky and risk-free assets but faces
 frictions to moving funds between them. To circumvent these frictions, he has access to an income-deduction scheme to accumulate risky assets.
--- a/HARK/ConsumptionSaving/tests/test_IndShockConsumerType.py
+++ b/HARK/ConsumptionSaving/tests/test_IndShockConsumerType.py
@@ -313,7 +313,7 @@ CyclicalDict = {
     "Rfree": 1.03,  # Interest factor on assets
     "DiscFac": 0.96,  # Intertemporal discount factor
     "LivPrb": 4 * [0.98],  # Survival probability
-    "PermGroFac": [1.082251, 2.8, 0.3, 1.1],
+    "PermGroFac": [1.1, 1.082251, 2.8, 0.3],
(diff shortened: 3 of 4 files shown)


sbenthall:
It looks like the multi-period cyclical model input order is not documented anywhere, so no changes are needed to accommodate this.

HARK/ConsumptionSaving/ConsIndShockModel.py
```diff
@@ -3047,7 +3052,7 @@ def construct_assets_grid(parameters):
 
 # Make a dictionary to specify an infinite consumer with a four period cycle
 init_cyclical = copy(init_idiosyncratic_shocks)
-init_cyclical['PermGroFac'] = [1.082251, 2.8, 0.3, 1.1]
+init_cyclical['PermGroFac'] = [1.1, 1.082251, 2.8, 0.3]
```
Mv77, line 3055: My only concern comes from my unfamiliarity with cyclical models in HARK.

For a finite horizon, non-cyclical model with two periods, I think HARK requires `len(PermGroFac) == 1`. The single element in the list is the growth factor of permanent income between the first and second periods.

How does this logic change when I'm considering an infinite-horizon model with a cycle of two periods? I know Chris, Matt and you had some lengthy discussion on how to treat the "terminal" period in cyclical vs non-cyclical models, which I did not follow, but I wanted to make sure this change (which looks good to me) is consistent with whatever you settled on.

Mv77 (commented):
This looks good to me and tests pass, but I have never worked with cyclical models in HARK so I am not to confident in my capacity to catch a bug here.

HARK/ConsumptionSaving/ConsIndShockModel.py
sbenthall: My understanding is that when `cycles == 0`, it's infinite horizon and so there is no terminal period solution.

When `cycles == N`, it's a finite model repeated many times. If a terminal solution is given, the solver returns N * T + 1 solutions.

sbenthall:
@mnwhite want to double check this before I merge?

sbenthall:
@mnwhite unless you have objections to this I'll merge this soon.

llorracc:
@sbenthall 

I haven't had time to look at this, but while it's on my mind let me ask something I've been a bit confused about from the beginning of the descriptions of what you are doing.

What confuses me is that there's nothing "wrong" about the way things have been done since the beginning.  It's perfectly acceptable to have some shocks be at the end of the period (that is, an EOP shock dated t is something whose parameters should properly live on solution[t]).

Problem is, most of our LaTeX and conceptual descriptions of what we are doing correspond to the shocks that matter between t and t+1 being realized at the beginning of t+1, but our code had them being realized at t.  

My pre-alpha-HARK-2.0 PR does it right (in a mathematical and conceptual sense), by allowing both BOP shocks or EOP shocks, and defining an infrastructure for handling either kind (or both kinds).  

So what's confusing me is that much of the discussion of all this seems to describe the shocks as being "on the wrong period" and as "moving" them.  My pre-alpha-HARK-2.0 does not need any change at all in the method of construction of the shocks and I can't quite see how a solution to the discrepancy between our LaTeX and our python description requires "moving" anything.  It just allows a variable 'shockTiming' which tells the code how to interpret the shocks that were already being constructed:  They should be interpreted (by the code) as BOP not EOP.  

I'm guessing that all the stuff you have done amounts to modifying the solution and simulation code to ASSUME that the shocks are at the BOP not the EOP.  But, really, the right solution is not to hard-wire either of them as being "right" or "wrong" -- both kinds of shocks are legit -- but to write code that correctly understands whether shocks of with any given name are BOP or EOP.

If there are going to be BOP and EOP shocks, probably we should have a convention for their names that makes it easy to read the code and figure out what is going on.  We could have IncShkDstn_BOP and IncShkDstn_EOP.

I think what you have done is that you've reworked the code to assume the shocks are BOP.  That's probably the best way to make this fix: If shocks do not have an explicit label, the default assumption should be that they are BOP instead of EOP.  

So, I guess (I hope) that my point is that it's confusing to talk about what is being done as it has been talked about, as though the code before had a "mistake."  The code was producing a perfectly valid solution to a perfectly valid problem.  It just that:

1. The problem being solved did not correspond exactly to the LaTeX description of it;
2. Substantively, what the LaTeX code describes is a more attractive way to do it;
3. So our code should be solving a (slightly) different model than what it was solving before

This requires, for example, the creation of a BeginningOfPeriod_vP object, and the redefinition of EndOfPrdvP to be essentially just the discount factor times BeginningOfPeriod_vP.

llorracc:
@sbenthall, @Mv77 

To clarify the relationship between the previous comment and the discussion of the "cyclical" code:  There's nothing wrong with the original code, which correctly requires the existence of a "solution_next" and constructs the current solution accordingly.  In this case, the "solution_next" for the last stage of the "cycle" would correspond to the solution to the first stage of the next cycle.  That was true before and it's true now, so I'm a bit confused about what needed to change.

sbenthall:
The problem I am trying to fix is that before this PR, there was no way to specify in a model's parameters what the shock values of the 0th period are.

These had to instead be hard-coded into the simulation code for the model. That reduces the generalizability of the software.

The shock value parameters for the 0th period are not used in the current solution methods. That is why they weren't included originally. But that is not a good choice if you're trying to provide a flexible modeling tool.

From my perspective, the question of whether shocks happen at the beginning or end of a period is an utter red herring, and creates more confusion than it settles. I abstain from further comment on that issue. However, I do know that every time we have discussed the changes I've made here (and corresponding changes proposed for the non-cyclical models), there has ultimately been consensus between you, myself, and @mnwhite. That's good enough for me!

llorracc:
I'm in agreement that we needed to do something systematic about the inability to specify shocks that occurred before the agent's first actions in the first simulated period; that is accomplished by allowing beginning-of-period shocks, as in my new code.  

For a variety of reasons, economic, notational, and computational, it makes a great deal of sense to allow shocks within the period either before or after decisions are made.  

But I agree that whenever you and Matt and I have discussed these issues in depth, we seem to have reached agreement conceptually.  So I'm pretty confident the questions in my prior message would evaporate with sufficient discussion.  It's more that I wanted to record a bit of puzzlement about some of the ways you've been talking about things, like whether the last period of this cycle should properly refer to "next" as the first period of the following cycle.  Since I can't see any reason we would change that, and it doesn't seem to have anything to do with the time t issue, I was worried we had somehow misunderstood each other.

llorracc:
Yes, you've got it now -- and as expected, now that we understand each other, we are in agreement.  Thanks.

PR #1047, opened 2021-07-23
ConsPortfolioModel Overhaul

alanlujan91:
This PR covers a number of changes to `ConsPortfolioModel` and supporting files:

- Creates object oriented solver to facilitate using pieces of solution method in other solvers. 
- Updates solution method for joint distribution of income and risky asset shocks (solver now runs, but there are no proper tests for this feature; an easy test is to assume independent distributions are not independent, which should give very similar if not exactly equal results)
- Sync example notebooks (`.py` example was outdated with respect to `.ipynb`)
- Fix example on mathematical limits of model which was outdated

A few comments: 

- `ConsPortfolioModel` does not have enough tests for its different features. 
  - Solver on discrete grid
  - Solver on Joint distribution


sbenthall:
Just to make a note of my answer to a question you asked earlier ...

The best way to design 'unit tests' is to actually test the functionality of individual units of code, rather than test the output of many components put together (which is technically called an 'integration test').

So, if there are gaps in the test coverage, the best thing to do is drill down and write tests covering the various cases for those specific parts.

PR #1048, opened 2021-07-29
These should be innocuous changes which can be merged into HARK master

llorracc:

### DIFF
--- a/HARK/ConsumptionSaving/ConsIndShockModelFast.py
+++ b/HARK/ConsumptionSaving/ConsIndShockModelFast.py
@@ -9,7 +9,8 @@ It currently solves three types of models:
    3) The model described in (2), with an interest rate for debt that differs
       from the interest rate for savings. #todo
 
-See `NARK <https://github.com/econ-ark/HARK/blob/master/Documentation/NARK/NARK.pdf>`_ for information on variable naming conventions.  See `hark.readthedocs.io <https://hark.readthedocs.io>`_ for mathematical descriptions of the models being solved.
+See NARK https://HARK.githhub.io/Documentation/NARK for information on variable naming conventions.
+See HARK documentation for mathematical descriptions of the models being solved.
 """
 
 from copy import deepcopy
@@ -33,7 +34,7 @@ from HARK.interpolation import (
     CubicInterp,
     ValueFuncCRRA,
     MargValueFuncCRRA,
-    MargMargValueFuncCRRA,
+    MargMargValueFuncCRRA
 )
 from HARK.numba import (
     CRRAutility,
@@ -251,8 +252,8 @@ def _solveConsPerfForesightNumba(
     hNrmNow = (PermGroFac / Rfree) * (hNrmNext + 1.0)
 
     # Calculate the lower bound of the marginal propensity to consume
-    PatFac = ((Rfree * DiscFacEff) ** (1.0 / CRRA)) / Rfree
-    MPCmin = 1.0 / (1.0 + PatFac / MPCminNext)
+    APF = ((Rfree * DiscFacEff) ** (1.0 / CRRA)) / Rfree
+    MPCmin = 1.0 / (1.0 + APF / MPCminNext)
 
     # Extract the discrete kink points in next period's consumption function;
     # don't take the last one, as it only defines the extrapolation and is not a kink.
@@ -442,11 +443,11 @@ def _prepare_to_solveConsIndShockNumba(
     )
 
     # Update the bounding MPCs and PDV of human wealth:
-    PatFac = ((Rfree * DiscFacEff) ** (1.0 / CRRA)) / Rfree
-    MPCminNow = 1.0 / (1.0 + PatFac / MPCminNext)
+    APF = ((Rfree * DiscFacEff) ** (1.0 / CRRA)) / Rfree
+    MPCminNow = 1.0 / (1.0 + APF / MPCminNext)
     Ex_IncNext = np.dot(ShkPrbsNext, TranShkValsNext * PermShkValsNext)
     hNrmNow = PermGroFac / Rfree * (Ex_IncNext + hNrmNext)
-    MPCmaxNow = 1.0 / (1.0 + (WorstIncPrb ** (1.0 / CRRA)) * PatFac / MPCmaxNext)
+    MPCmaxNow = 1.0 / (1.0 + (WorstIncPrb ** (1.0 / CRRA)) * APF / MPCmaxNext)
 
     cFuncLimitIntercept = MPCminNow * hNrmNow
     cFuncLimitSlope = MPCminNow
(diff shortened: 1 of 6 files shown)


llorracc:
Fixed one small test failure, where a test had relied upon an internal feature of a function that was renamed.  All tests now passing.

HARK/ConsumptionSaving/ConsIndShockModelFast.py
```diff
@@ -1116,22 +1117,6 @@ def update_solution_terminal(self):
...
-        ``vFuncNvrs`` has value of zero at the lower bound of market resources
-        `mNrmMin`.  See the `PerfForesightConsumerType <https://hark.readthedocs.io/en/latest/example_notebooks/PerfForesightConsumerType.html?highlight=PerfForesightConsumerType#Solution-method-for-PerfForesightConsumerType>`_ documentation notebook
-        for a brief explanation and the links below for a fuller treatment.
-
-        `PerfForesightCRRA/#vFuncAnalytical <https://www.econ2.jhu.edu/people/ccarroll/public/lecturenotes/consumption/PerfForesightCRRA/#vFuncAnalytical>`_
-
-        `SolvingMicroDSOPs/#vFuncPF <https://www.econ2.jhu.edu/people/ccarroll/SolvingMicroDSOPs/#vFuncPF>`_
-
-        """
-
```
sbenthall, line 1134: In the original file, this is the docstring of a method, which would be rendered to the Sphinx docs.

You seem to have moved it elsewhere in the function.

I gather that the description of the math applies better to the later code.

But are there no docstrings for methods in this class?

HARK/distribution.py
```diff
@@ -837,6 +838,36 @@ def draw(self, N, X=None, exact_match=False):
...
+    def __init__(self, pmf, X, seed=0):  # Get the old definition
+        super().__init__(pmf, X, seed)  # execute the old def
+        XYZ = np.column_stack(X)  # stack actually unstacks them
+        if self.dim() == 1:  # numpy 1 dimensional arrays want to be
+            XYZ = XYZ.T  # the wrong shape
+        self.XYZ = XYZ  # [ N dimensional matrix ]
+        # The pmf should be a function, not a vector (as now)
+        # TODO: Replace invocations of pmf with pmv
+        self.pmv = pmf
+
```
sbenthall, line 870: I understand the motivation for this change and it makes sense.

But surely this change should be made to the DiscreteDistribution class directly, rather than creating this very awkward wrapper.

I wonder where pmf is being used in the downstream code.

HARK/distribution.py
```diff
@@ -1196,9 +1227,27 @@ def combine_indep_dstns(*distributions, seed=0):
...
+# handcrafted specialized functions that know that their positional first argument must be
+# a matrix whose structure matches the structure of the discrete income distribution provided
+# as the first argument. For example, if the 2D distribution is of transitory and permanent
+# shocks, the function must take as an input a matrix which is assumed to have a structure
+# that exactly mirrors the internal structure of the arrangement of broadcasted transitory
+# and permanent shocks inside the dstn object, and knows(for example) that the value of the permanent shock can be how to extract
+
+# hard to understand because it requires the user to
+# # provide an array of values of its inputs but then assumes it already knows what those
+# # inputs are. This makes little sense. If it is going to assume it knows what the inputs
```
sbenthall, line 1243: I don't think this comment should be included in the code as is.

It has strange formatting/grammatical errors.

Also, it's not clear what is gained by having this critical opinion included in the source code. It is distracting and, as an opinion, contestable.

It would be more appropriate to raise this as an issue or discussion on GitHub. It is quite possible that there is a better design for the function. The source code comments are not the right place to debate that design.

HARK/distribution.py
```diff
@@ -1249,6 +1298,16 @@ def calc_expectation(dstn,func=lambda x : x,*args):
...
+# "calc_expectation" is not a good name for something that
+# requires an array as an argument, returns an array, and
+# expects a peculiarly shaped distribution object as its first
+# argument.
+# TODO: 20210618: we REALLY need to improve our tools for calculating
+# expectations, both to make them better documented and more flexible.
+
+
+calc_expectation_of_array = calc_expectation
+
```
sbenthall, line 1310: Again, this is not the appropriate place to enjoin a debate about the design of this function, which I agree could use more work.

I also disagree with the opinions in the comment.

The current form of this function does not require an array as an argument, for starters.

sbenthall:
See inline comments.

I have a non-blocking question about what's going on with comments in ConsIndShockModelFast.

I think the proposed changes to `distribution.py` are things that should be discussed on GitHub and implemented correctly, not merged in as opinionated rants within the source code itself, where they are a distraction.

HARK/ConsumptionSaving/ConsIndShockModelFast.py
llorracc: I think the goal here was mainly to fix the urls, and if the docstring was moved so that it no longer applies to the class that is probably a mistake.  I have yet to devote any time to mastering how to use sphinx in docstrings.  

So, please just move it back to being the docstring.

HARK/distribution.py
llorracc: I made the wrapper because pmf is being used in many many places in the code and I didn't want to break them all.  Maybe we could put a "deprecation notice" on the old version of the function and plan eventually to replace it with its wrapper.

HARK/distribution.py
llorracc: I've put comments of this kind pervasively throughout the revisions I've been making, mainly to explain them to the reviewers (you, Mridul, Matt) upon review, and with the expectation that the right workflow will be, if the comment's substance is something we agree on, it should be formulated as an issue or a "TODO" item.  

You are welcome to begin that process right now, with this comment, if you like.  (The odd formatting is just because it seems I commented the inner block earlier and later commented a larger block containing the inner comment at some point, leading to two #)

sbenthall: Ok. So do I understand correctly that you would like Mridul or I to take this PR and make whatever modifications to it are necessary to make it pass review, including:
 - fixing the docstring location 
 - removing these comments in distribution.py but turning them into issues?

llorracc: Yes, that's exactly what I had in mind.  For anything that does not change the functioning of any code, you should propose whatever changes (to docstrings, comments, code formatting or structure, etc) you think suitable here, and I'll either approve them (likely) or follow up with a response.

In this case, just fixing the docstring will not require a response.

sbenthall: Ok. I see that this a branch on the main HARK repository so I can make these changes myself with no permission conflicts.
I'll get to this over the next few days.

llorracc:
@sbenthall,

Since all the tests are passing, if you could either propose any functional changes you think should be made now, or agree to merge now and discuss functional redesigns later, that would be helpful.  I'd like to finalize this by the time of our Thu meeting. 
It would be good to have these innocuous elements that are used in the BST notebook merged, so that the pre-alpha changes that are less innocuous can be considered without clutter.

PS.  As part of your revisions, I would encourage you to remove the commentary that is in the PR but that I never intended to become part of the codebase, per the discussion above.

llorracc:
@sbenthall Thanks!

PR #1057, opened 2021-08-08
Minor reorganization of RiskyContrib

Mv77:
This PR moves a function definition of `RiskyContrib` from the main workspace to inside the solver (which is the only place where it is used). It makes the code more readable and results in less parameters being passed around unnecessarily.

@sbenthall does such a small change merit a changelog entry?


sbenthall:
@Mv77 the change looks good.
It's technically a change to the API, so a changelog entry does not hurt!

Mv77:
Sure, I've added a changelog entry and this is now ready to merge.

PR #1058, opened 2021-08-10
Iterated expectations in RiskyContrib

Mv77:
There is a speed-up trick that can be used in portfolio problems when income is independent of the risky return. This is already incorporated in `ConsPortfolio`, but I had not applied it to `RiskyContrib`.

The trick is to use iterated expectations when computing the expectation of value (and derivatives) next period. The idea is to create a function that interpolates `E[f(income, R)|R]` and then integrate it over `R`, instead of jointly integrating over `income` and `R`. Turns out that---in this model---this change makes the solution 3x-5x faster.

This PR introduces this trick into `RiskyContrib`'s solver. I left the general solver for cases where income and returns are correlated as an option.

### DIFF
--- a/HARK/ConsumptionSaving/ConsRiskyContribModel.py
+++ b/HARK/ConsumptionSaving/ConsRiskyContribModel.py
@@ -1,7 +1,25 @@
 """
 This file contains classes and functions for representing, solving, and simulating
-agents who must allocate their resources among consumption, saving in a risk-free
-asset (with a low return), and saving in a risky asset (with higher average return).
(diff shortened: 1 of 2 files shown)


Mv77:
Since the REMARK has now been published, I took the opportunity to edit the file header and link to it.

Mv77:
Massive thanks to @alanlujan91, as I decided to implement this trick after reviewing his code for the housing project.

sbenthall:
@mnwhite are you able to review this?

PR #1061, opened 2021-08-30
Implementing Harmenberg method to Monte Carlo Methods 

wdu9:


llorracc:
@sbenthall,

@wdu9 showed @Mv77 and me the results from this yesterday and it's pretty awesome.  Eyeballing his results, it looks to me like it will speed up simulations by a minimum of a factor of 3 and maybe as much as a factor of 10 or more.  This should apply directly to the simulations you are doing for the DARPA project.  Since you are burning a lot of cycles to get the simulation results, it might be an efficient diversion for you to figure out how to port this over to what you are doing.

Possibly there is one more element that should be added to the PR before merging, which is a more detailed intuitive explanation of what the Harmenberg method is and why it achieves such a massive speedup, in the form of a Jupyter notebook (probably thinking of it as targeted to eventually become part of the documentation).  I think at present basically we are just citing his paper, which is brief and written at a high level of abstraction.  That sounds like a project that you and Will might tackle together, with input from Mateo, since Mateo will almost certainly find it very useful to adopt the new method for his dissertation research.

llorracc:
@wdu9, as we discussed yesterday, could you take a look to see whether possibly it might be super easy to use the new method for the DistributionOfWealthAndTheMPC notebook?  That seems like a great starting point for incorporating it into applications of the toolkit.

wdu9:
Sure, I will check it out.
________________________________

sbenthall:
I'll review this in the coming week.

sbenthall:
I'm confused by #1065, which seems to be written as if this PR has already been merged.

HARK/ConsumptionSaving/ConsIndShockModel.py
```diff
@@ -1848,6 +1848,47 @@ def get_poststates(self):
         super().get_poststates()
 
         return None
+    
+    def sim_agg_path(self):
+        """
+        Simulates and calculates the paths of aggregate consumption, assets and cash on hand storing each as
+        attrbutes of self. The simulation utilizes (Harmenberg 2021) method involving an altered permanent shock distribution.
```
sbenthall, line 1855: "attributes"

HARK/ConsumptionSaving/ConsIndShockModel.py
```diff
@@ -1848,6 +1848,47 @@ def get_poststates(self):
...
+        
+        """
+        
+        self.track_vars = ['aNrm','mNrm','cNrm'] # #variables to track
+        self.ntrl_msr = True 
+        self.update_income_process()
+        self.initialize_sim()
+        self.simulate()
+        
+        Asset_list = [] # list of aggregate asset values
```
sbenthall, line 1873: PEP8 says these should be in snake_case, or lower_case_with_underscore

HARK/ConsumptionSaving/ConsIndShockModel.py
```diff
@@ -2035,6 +2076,8 @@ def check_conditions(self, verbose=None):
         "T_retire": 0,  # Period of retirement (0 --> no retirement)
         "vFuncBool": False,  # Whether to calculate the value function during solution
         "CubicBool": False,  # Use cubic spline interpolation when True, linear interpolation when False
+        "ntrl_msr": False,      # Use permanent income neutral measure during simulations when True.
```
sbenthall, line 2079: would it be possible to have this be a more legible parameter name?
like `neutral_measure` ?

HARK/ConsumptionSaving/ConsIndShockModel.py
```diff
@@ -2097,10 +2140,18 @@ def update_income_process(self):
             IncShkDstn,
             PermShkDstn,
             TranShkDstn,
+            IncShkDstn_ntrl,
```
sbenthall, line 2143: preference for `_neutral` rather than `_ntrl`

HARK/ConsumptionSaving/ConsIndShockModel.py
```diff
@@ -2178,7 +2229,7 @@ def get_shocks(self):
             these = t == self.t_cycle
 
             # temporary, see #1022
-            if self.cycles == 1:
+            if self.cycles == 1 and self.ntrl_msr == False:
```
sbenthall, line 2232: I'm wondering why this change

HARK/ConsumptionSaving/ConsIndShockModel.py
```diff
@@ -2664,6 +2715,11 @@ def construct_lognormal_income_process_unemployment(self):
         TranShkDstn : [[distribution.Distribution]]
             A list with T_cycle elements, each of which
             a discrete approximation to the transitory income shocks.
+        IncShkDstn_ntrl :  [distribution.Distribution]
+            A list with T_cycle elements, each of which is a
+            discrete approximation to the income process in a period.
+            Permanent shock component of the income process uses the permanent 
+            income neutral measure (see Harmenberg 2021).
```
sbenthall, line 2722: An alternative way to design this would be to have `neutral_measure` to be an input parameter to this `construct_lognormal_income_process_unemployment` and then return the `IncShkDstn`  with the Harmenberg modification.

That would keep the number of distributions passed around more parsimonious and backwards compatible. See lines 2150-2152, which would the be unecessary.

HARK/ConsumptionSaving/tests/test_IndShockConsumerType.py
```diff
@@ -581,4 +581,130 @@ def transition(self):
...
+    "AgentCount" : 15000,                 # Number of agents of this type
+    "T_sim" : 4000,                         # Number of periods to simulate
+    "aNrmInitMean" : np.log(1.3)-(.5**2)/2,# Mean of log initial assets
+    "aNrmInitStd"  : .5,                   # Standard deviation of log initial assets
+    "pLvlInitMean" : 0,                    # Mean of log initial permanent income
+    "pLvlInitStd"  : 0,                    # Standard deviation of log initial permanent income
+    "PermGroFacAgg" : 1.0,                 # Aggregate permanent income growth factor
+    "T_age" : None,                        # Age after which simulated agents are automatically killed
+
+}
```
sbenthall, line 639: I'm unclear as to why a new dictionary is needed here.

Does it have a lot in common with another parameter dictionary?
If so, it would be good to explicitly extend it with any additional parameters that are needed to support this new feature.

sbenthall:
It would be nice to test this for a case where the original permanent income shock has a mean that isn't 1.0.

Issue #1063, opened 2021-09-08
[framed]aggregate variables handled by frame parameter


sbenthall:
Currently in the FramedAgentType, an aggregate variable is indicated through a specialized data structure defined in the class constructor.

It would be better an aggregate variable were indicated to be as such by a parameter handed to its frame upon construction.

PR #1064, opened 2021-09-09
set aggregate property on frame

sbenthall:
This solves #1063 by setting the aggregate property of a variable on its frame in a FramedAgent.
It also pulls all the FramedAgent logic into its own modules `frames`

No new tests -- this is a refactor.
Some docs fixes


sbenthall:
@Mv77  could I ask you to review this? Should be quick.

HARK/ConsumptionSaving/ConsPortfolioFrameModel.py
```diff
@@ -71,19 +73,9 @@ def __init__(self, **kwds):
...
-            'Rport' : None,
-            'aLvl' : None,
-            'aNrm' : None,
-            'bNrm' : None,
-            'mNrm' : None,
-            'pLvl' : None
-            }
+        self.shocks = {}
+        self.controls = {}
+        self.state_now = {}
```
alanlujan91, line 78: where do these get defined now?

alanlujan91: nvm i found it

alanlujan91:
This looks good to me. I wonder if there are frames that could be joined, like PermShk and TranShk. Too many frames might become too complex to understand for new users. It would also be useful to map frames in the simulation with frames (or subperiods) in the solution method.

alanlujan91 (approved):

sbenthall:
Yes, mapping simulation frames to the solution is a desired next step.

There seems to be some difference of opinion about the best way to group frames.
I prefer more granular frames. @mnwhite thinks more grouped frames might be better for performance reasons. I hear you saying you think more grouped frames are easier to understand.

The current implementation is agnostic about the best amount of aggregation in framing. That way performance of the same model, differently written, can be tested.

One benefit of granular frames is that they support the kind of plotting done in #1071

PR #1071, opened 2021-10-12
Plotting graph of equations for FramedAgentType

sbenthall:
This PR introduces into FramedAgentType the mechanics for frames to reference each other as 'parents' and 'children', including 'back' and 'forward' references when these frames are in a different period.

This enables the plotting of a MACID style network diagram representing the problem, as in the following case (taken from the example notebook):

image


alanlujan91 (approved):
This looks good to me. Is there a way to provide a pseudo-simulation history of shocks that can be used by both agent types? In this way, they both experience the same exact shocks, and differences would be fully attributable to bugs and not to changes in random draws.

sbenthall:
That's an excellent idea.

I believe there is support for this in the library, but I've never used it:
https://github.com/econ-ark/HARK/blob/a5634abb233297d7503039dfa2a6d9d34dec42a8/HARK/core.py#L553-L556

Let me try that out before merging. Thanks for the review!

sbenthall:
Ok, I tried copying the shock history over from the first portfolio simulation to the Framed version and using `read_shocks`.

The results were somewhat unexpected:
 - The shock values did _not_ line up exactly.
 - But the results of the simulation (i.e, derived values, such as `mNrm`) _did_ come into closer alignment.

I may well be using `read_shock` incorrectly and so this needs more investigation.

But since that testing is somewhat orthogonal to the plotting work here, I'll merge this PR and address this in a different issue.

PR #1073, opened 2021-10-19
I corrected a mistake in the DCT notebook.

iworld1991:
I discovered a mistake in the DCT.ipynb.  Previously, the example of 1-dimension DCT production seems to suggest that higher precision leads to worse approximated function. It turns out that this was due to an error in code. Now, higher precision leads to better approximation of the function.


iworld1991:
Seems there is still a problem. Let me further debug it.

PR #1074, opened 2021-10-19
Fixed a bug in DCT.ipynb

iworld1991:
Previously, the 1-dimension DCT was not implemented correctly, which leads to the mistaken illustration that the approximated function is not closer to the original function one as the precision level increases.  Now it is fixed.


llorracc:
BayerLuetticke is there -- it's just moved, I forget where, you should be able to find it.

PR #1077, opened 2021-11-04
Update ConsRiskyAssetModel.py

alanlujan91:
I think there is a minor mistake in the following method 

https://github.com/econ-ark/HARK/blob/b36b8f8175568a2bac6afa58cb06198f6cb66c74/HARK/ConsumptionSaving/ConsRiskyAssetModel.py#L167-L190

The parameters for the simulated risky shock should be ~~`self.`~~ `RiskyAvg` and ~~`self.`~~ `RiskyStd` so that the simulated Risky distribution uses the true parameters instead of the age-varying beliefs.


Mv77 (approved):
Indeed, nice catch! Thank you

PR #1079, opened 2021-11-08
Update ConsPortfolioModel.py

alanlujan91:
There's long been a comment on `ConsPortfolioModel` that says: 

> This loop can probably be eliminated, but it's such a small step that it won't speed things up much.

This is my attempt at replacing the loop with vectorized calculations to test just how much faster this could be. Working out how to do this in vectorized form can provide insights into how to do it in higher dimensions, where loops might be much slower than in this instance. 

### DIFF
--- a/HARK/ConsumptionSaving/ConsPortfolioModel.py
+++ b/HARK/ConsumptionSaving/ConsPortfolioModel.py
@@ -716,42 +716,42 @@ class ConsPortfolioSolver(MetricObject):
         Optimization of Share on continuous interval [0,1]
         """
 
-        # For values of aNrm at which the agent wants to put more than 100% into risky asset, constrain them
         FOC_s = self.EndOfPrddvds
-        # Initialize to putting everything in safe asset
-        self.Share_now = np.zeros_like(self.aNrmGrid)
-        self.cNrmAdj_now = np.zeros_like(self.aNrmGrid)
+
+        # For each value of aNrm, find the value of Share such that FOC-Share == 0.
+        crossing = np.logical_and(FOC_s[:, 1:] <= 0.0, FOC_s[:, :-1] >= 0.0)
+        share_idx = np.argmax(crossing, axis=1)
+        a_idx = np.arange(self.aNrm_N)
+        bot_s = self.ShareGrid[share_idx]
+        top_s = self.ShareGrid[share_idx + 1]
+        bot_f = FOC_s[a_idx, share_idx]
+        top_f = FOC_s[a_idx, share_idx + 1]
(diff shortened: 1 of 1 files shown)


HARK/ConsumptionSaving/ConsPortfolioModel.py
```diff
@@ -731,42 +731,42 @@ def optimize_share(self):
...
+        bot_s = self.ShareGrid[share_idx]
+        top_s = self.ShareGrid[share_idx + 1]
+        bot_f = FOC_s[a_idx, share_idx]
+        top_f = FOC_s[a_idx, share_idx + 1]
+        bot_c = self.EndOfPrddvdaNvrs[a_idx, share_idx]
+        top_c = self.EndOfPrddvdaNvrs[a_idx, share_idx + 1]
+        alpha = 1.0 - top_f / (top_f - bot_f)
+
+        self.Share_now = (1.0 - alpha) * bot_s + alpha * top_s
+        self.cNrmAdj_now = (1.0 - alpha) * bot_c + alpha * top_c
```
alanlujan91, line 749: `alpha` and other variables here can be saved and added to `save_points`.

alanlujan91:
Good candidate for benchmark #1131

HARK/ConsumptionSaving/ConsPortfolioModel.py
```diff
@@ -652,42 +652,42 @@ def optimize_share(self):
...
+
         # If agent wants to put more than 100% into risky asset, he is constrained
         constrained_top = FOC_s[:, -1] > 0.0
         # Likewise if he wants to put less than 0% into risky asset
         constrained_bot = FOC_s[:, 0] < 0.0
+
+        # For values of aNrm at which the agent wants to put
+        # more than 100% into risky asset, constrain them
         self.Share_now[constrained_top] = 1.0
+        self.Share_now[constrained_bot] = 0.0
```
alanlujan91, line 680: @Mv77 looks like I thought about fixing Share_now for points constrained at the bottom a few years ago but never got around to merging it

Mv77 (approved):
Looks good.

This is not about this PR but I don't really like how the optimal share and consumption are found.

I think it'd be clearer (and maybe more accurate?) to do the stage thing where the optimal share associated with every a is found and then the a is decided later.

But that's for another day. Not what this PR is about.

Happy to have it merged.

alanlujan91:
I agree. There are some shortcuts here that are meant to be "close enough" and maybe pedagogical. But maybe we should have the "best" portfolio choice solver we can come up with and the more pedagogical one as separate models.

PR #1082, opened 2021-11-11
Keep FOC's with respect to risky share for every point on the grid

alanlujan91:


alanlujan91:
approved by @llorracc via email

PR #1094, opened 2021-12-23
[WIP]: normalize_shocks and normalize_levels allow user to impose true propositions on sims

llorracc:
This implements a simple version of an old idea from CDC Mathematica code:  Simulation efficiency can be substantially improved by imposing on the stochastic draws facts that we know are true in the population, like that the average values of permanent and transitory shocks are 1 in each period.

### DIFF
--- a/HARK/ConsumptionSaving/ConsIndShockModel.py
+++ b/HARK/ConsumptionSaving/ConsIndShockModel.py
@@ -2209,11 +2209,13 @@ class IndShockConsumerType(PerfForesightConsumerType):
                     PermShkMeanNow = np.mean(IncShks[0])
                     TranShkMeanNow = np.mean(IncShks[1])
 
-                PermShkNow[these] = (# permanent "shock" includes expected growth
-                    (IncShks[0, :] / PermShkMeanNow) * PermGroFacNow
-                )  
+                PermShkNow[these] = (
+                    (IncShks[0, :] * PermGroFacNow
+                     / PermShkMeanNow) 
+                )  # permanent "shock" includes expected growth
                 TranShkNow[these] = (
-                    (IncShks[1, :] / TranShkMeanNow) 
+                    (IncShks[1, :]
+                     / TranShkMeanNow) 
                 )
 
         # That procedure used the *last* period in the sequence for newborns, but that's not right
@@ -2221,7 +2223,7 @@ class IndShockConsumerType(PerfForesightConsumerType):
         N = np.sum(newborn)
         if N > 0:
             these = newborn
-            IncShkDstnNow = self.IncShkDstn[0]  # set current shock distribution
+            IncShkDstnNow = self.IncShkDstn[0]  # set current income distribution
             PermGroFacNow = self.PermGroFac[0]  # and permanent growth factor
 
             # Get random draws of income shocks from the discrete distribution
@@ -2234,11 +2236,13 @@ class IndShockConsumerType(PerfForesightConsumerType):
                 PermShkMeanNow = np.mean(IncShkDstnNow.X[0][EventDraws])
                 TranShkMeanNow = np.mean(IncShkDstnNow.X[1][EventDraws])
                     
-            PermShkNow[these] = (# permanent "shock" includes expected growth
-                (IncShkDstnNow.X[0][EventDraws] / PermShkMeanNow)* PermGroFacNow 
+            PermShkNow[these] = (
+                (IncShkDstnNow.X[0][EventDraws] * PermGroFacNow
+                 / PermShkMeanNow)
             )  
             TranShkNow[these] = (
-                (IncShkDstnNow.X[1][EventDraws] / TranShkMeanNow)
+                (IncShkDstnNow.X[1][EventDraws]
+                 / TranShkMeanNow)
             )
         #        PermShkNow[newborn] = 1.0
         TranShkNow[newborn] = 1.0
@@ -2399,7 +2403,7 @@ class IndShockConsumerType(PerfForesightConsumerType):
         #        AgentType.pre_solve(self)
         # Update all income process variables to match any attributes that might
         # have been changed since `__init__` or `solve()` was last called.
-        self.update_income_process()
+        #        self.update_income_process()
         self.update_solution_terminal()
         if not self.quiet:
             self.check_conditions(verbose=self.verbose)
--- a/HARK/ConsumptionSaving/ConsIndShockModel.py
+++ b/HARK/ConsumptionSaving/ConsIndShockModel.py
@@ -1800,7 +1800,7 @@ class PerfForesightConsumerType(AgentType):
         if self.normalize_levels == True:
             pLvlNowMean = np.mean(pLvlNow)
 
-        pLvlNow = pLvlNow / pLvlNowMean # Does nothing if normalize_levels != True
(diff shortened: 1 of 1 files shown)


llorracc:
@Mv77, could you review this?  And, if you have the right permissions, merge it?  

It passes all tests and should not affect any existing results, since if the normalize booleans are not set then all it does is divide some things by 1.0.

PS.  I'm interested to see how much improvement there is from this versus from the Harmenberg thing.  And, I can't see any reason they couldn't be combined, to achieve even more improvement.  What I really want to do, though, is to get the "reshuffling" technology implemented.

@wdu9, you might be interested too.  Thanks for your earlier input -- you put me on the right track

Mv77:
@llorracc I am starting to take a look now. Will get to the code soon, but I have a couple of conceptual questions.

- **On dividing shocks by their means:** I believe that doing things like `TranShkNow[these] = IncShks[1, :] / TranShkMeanNow` introduces the possibility of (normalized) transitory shocks taking on values not included in the discrete approximation `TranShkDstn.X`. How do you feel about that? The agent is drawing shocks that were not allowed in the expectations he formed. I believe this points back to the debate on whether these discretized solutions should be seen as approximations of the continuous problem. I wanted to hear your thoughts on that, since I believe this fact complicates things you have been thinking about (like *shuffling*).

-  **On normalizing levels**: 
a) There should be some guardrails ensuring that `pLvlNow = pLvlNow / pLvlNowMean` happens only for models in which it is in fact true that the average `pLvl` should be 1 every period. This excludes e.g. life-cycle models or models with aggregate growth, right?
b) `pLvlNow = pLvlNow / pLvlNowMean` alters the joint distribution of (m,P). Should we worry about that? Consider an agent who ends period 0 with p=1, a=1. At the start of period 0 he draws PermShk=0.5 and now he has, approximately p_1 = 0.5, m_1 = 2. But then we apply the pLvl normalization and this can shift his p_1 up or down while leaving m_1 fixed. This is picky, the normalization should have a very small effect for large populations. Nevertheless, my point is that p affects m through PermShk before it is normalized, and I wonder if normalizing p after setting m without taking this link into account can have perverse consequences.

llorracc:
> 



> .
> 
> * **On dividing shocks by their means:** I believe that doing things like `TranShkNow[these] = IncShks[1, :] / TranShkMeanNow` introduces the possibility of (normalized) transitory shocks taking on values not included in the discrete approximation `TranShkDstn.X`. How do you feel about that? The agent is drawing shocks that were not allowed in the expectations he formed. I believe this points back to the debate on whether these discretized solutions should be seen as approximations of the continuous problem. I wanted to hear your thoughts on that, since I believe this fact complicates things you have been thinking about (like _shuffling_).

This is exactly right.  It goes back to whether we are thinking of our discretizations as defining an exact model that is being solved on its own terms, or whether we are thinking of them as approximations of a "true" model in which the shock is, say, really continuously lognormally distributed.  My preference is strongly for the latter, because it is the more general formulation.  If two solutions to a model differ because one used a 5 point and the other used a 7 point approximation, then whichever of them is closer to what you get when you have an infinity-point approximation is the one that is defined as being "closer" to the truth.  I'd rather do shuffling than dividing, because shuffling has the virtue that the simulated outcomes match numerically identically the computations that went into the calculation of the expectations, and it is deeply attractive to have identical calculations going into the solution and the simulation phases.  But implementing shuffling would require considerably more work, and it is possible that dividing by the mean gets 95 percent of the benefits -- that is something I want to figure out.



> * **On normalizing levels**:
>   a) There should be some guardrails ensuring that `pLvlNow = pLvlNow / pLvlNowMean` happens only for models in which it is in fact true that the average `pLvl` should be 1 every period. This excludes e.g. life-cycle models or models with aggregate growth, right?

No, PermGroFac is a separate object from PermShk.  Throughout, we have always insisted that E[PermShk]=1, and have handled either life cycle patterns or aggregate growth using PermGroFac.  So, it's not a problem to impose E[PermShk]=1.

> * b) `pLvlNow = pLvlNow / pLvlNowMean` alters the joint distribution of (m,P). Should we worry about that? Consider an agent who ends period 0 with p=1, a=1. At the start of period 0 he draws PermShk=0.5 and now he has, approximately p_1 = 0.5, m_1 = 2. But then we apply the pLvl normalization and this can shift his p_1 up or down while leaving m_1 fixed. This is picky, the normalization should have a very small effect for large populations. Nevertheless, my point is that p affects m through PermShk before it is normalized, and I wonder if normalizing p after setting m without taking this link into account can have perverse consequences.

This may be a very good catch:  I have not examined the simulation code carefully enough to determine whether, as I had assumed, the draw of the permanent shock occurs before the calculation of b or m.  If it does, then we're fine.  That was the case in my original Mathematica code, so I had assumed it is the case in our HARK simulations, but if not then this step may need to be moved to some earlier point.  (Though, if that is the case, maybe some renaming will be in order, since I think of "transitions" as defining how you get from t-1 to t, and if by the time you get to "transitions" some of that has already been done then our nomenclature may not be ideal).

Mv77:
> No, PermGroFac is a separate object from PermShk. Throughout, we have always insisted that E[PermShk]=1, and have handled either life cycle patterns or aggregate growth using PermGroFac. So, it's not a problem to impose E[PermShk]=1.

Fully agree, but you are imposing `E[pLvl] = 1` in some parts https://github.com/econ-ark/HARK/blob/8f0057e0cb51f9acf918222954463e2cd4f7fe7f/HARK/ConsumptionSaving/ConsIndShockModel.py#L1809

Mv77:
> This may be a very good catch: I have not examined the simulation code carefully enough to determine whether, as I had assumed, the draw of the permanent shock occurs before the calculation of b or m. If it does, then we're fine. That was the case in my original Mathematica code, so I had assumed it is the case in our HARK simulations, but if not then this step may need to be moved to some earlier point. (Though, if that is the case, maybe some renaming will be in order, since I think of "transitions" as defining how you get from t-1 to t, and if by the time you get to "transitions" some of that has already been done then our nomenclature may not be ideal).

Here is the relevant code https://github.com/econ-ark/HARK/blob/8f0057e0cb51f9acf918222954463e2cd4f7fe7f/HARK/ConsumptionSaving/ConsIndShockModel.py#L1795-L1816

Notice `pLvlNow` is set in line 1796, using `PermShk`. Then `pLvlNow` is normalized in line 1809. But then `ReffNow`, which is used to compute `bNrmNow` is set using `PermShk` in 1814.

llorracc:
Oh, yes, you're right about that.  When I did that my thought was "the right way to handle this is to have an aggregate PLvl variable that tracks the aggregate movements and an idiosyncratic pLvl whose mean should always be 1 but I have a sneaking suspicion we have not done it that way ... even though it's being done that way in the particular case I'm working with right now (Harmenberg-Aggregation)."

llorracc:
Closing as superseded. The idea here, imposing on the simulated shock draws what is known to hold in the population (a cross-sectional mean of one for the permanent and transitory shocks each period), is now in HARK as the opt-in low-variance draw modes of #1786 (`shuffle`, `draws`, `replicates` in the distributions layer, merged August 2026), which do it at the distribution level for every model rather than inside one consumer type. The other thread of the 2022 discussion, an aggregate permanent income level tracked separately from the idiosyncratic one, is HARK's `PlvlAgg` / `PermGroFacAgg`; see the note on the two economies in #1842.

PR #1099, opened 2022-01-09
Add `aLvl` to `state_vars`

Mv77:
`aLvl` (`aNrm * pLvl`) is not a state that is required for solving or simulating the models in `ConsIndShockModel`. However, it was determined at some point in history that it was useful to keep track of `aLvl` as if it were a state. This is done for convenience and is mostly used in `ConsAggShockModel`.

See
https://github.com/econ-ark/HARK/blob/a11325ddfc1c403fd91a4120487a955ca30cded0/HARK/ConsumptionSaving/ConsIndShockModel.py#L1845-L1846

However, `aLvl` was never added to the list of state variables https://github.com/econ-ark/HARK/blob/a11325ddfc1c403fd91a4120487a955ca30cded0/HARK/ConsumptionSaving/ConsIndShockModel.py#L1583

This PR adds `aLvl` to the list of states so that `state_now` has the same entries as `state_vars`, which seems desirable to me.

### DIFF
--- a/HARK/ConsumptionSaving/ConsIndShockModel.py
+++ b/HARK/ConsumptionSaving/ConsIndShockModel.py
@@ -1580,7 +1580,7 @@ class PerfForesightConsumerType(AgentType):
     )
     time_vary_ = ["LivPrb", "PermGroFac"]
     time_inv_ = ["CRRA", "Rfree", "DiscFac", "MaxKinks", "BoroCnstArt" ]
-    state_vars = ['pLvl', 'PlvlAgg', 'bNrm', 'mNrm', "aNrm"]
+    state_vars = ['pLvl', 'PlvlAgg', 'bNrm', 'mNrm', "aNrm", 'aLvl']
     shock_vars_ = []
 
     def __init__(self, verbose=1, quiet=False, **kwds):


sbenthall:
I believe that originally `aLvl` was in the `poststate_vars`.

What do you see as the meaning of `state_vars`?

Currently this parameter is not documented and I fear that its semantics are actually ambiguous across different models.

mnwhite:
It should mean the same thing in all models. It's just pLvl*aNrm, that's
it. It can't be known until the end of period t, once aNrm has been
determined, so it should be in poststate_vars.

I think I might have originally omitted it from poststate_vars because it
wasn't initialized for newborns, and never needed to be used. That is,
poststate_vars (in my intended design) represents the "interface" variables
between periods t and t+1, and thus the ones that need to be initialized
for a newborn.

llorracc:
Incidentally, in thinking through nomenclature as I was working on my pre-alpha HARK 2.0, I decided we should really have a variable that corresponds to the beginning-of-period-t+1 value that is determined by end-of-period a_{t}.  So, the final version of [BufferStockTheory](https://econ-ark.github.io/BufferStockTheory/BufferStockTheory3.html#DBCparts) defines $k_{t+1}=a_{t}$, where $k_{t+1}$ is known at the very beginning of period $t+1$ even before permanent shocks (or any other shocks) are drawn.

mnwhite:
Yes, I like that as well. And then k_t can be what's drawn (or determined,
in some cases) for newborns, rather than a *very* temporary a_{t-1}. In
models where we redistribute the wealth of the dead to the survivors, the
transition equation from a_t to k_{t+1} would depend on who gets the icy
finger of death in `who_dies`. In a perpetual youth model, that might be
our typical k_t = a_{t-1}/LivPrb. In a lifecycle model, we might give *all*
of the wealth of decedents to the newborns as a stake-at-birth, evenly
distributed. Or something else entirely.

The distinction between a_{t-1} and k_t is also the notation used in the
cAndCwithStickyE model on paper.

llorracc:
Let's agree to settle on this.  Newborns enter life with initial capital k, and there is no "pre-birth" fictive a_{t-1} for them.  It is a state variable, while a_{t} is a "post-state" variable.

mnwhite:
It will take some rejiggering to make those labels appear in the code, but
yes, I agree that's what it should be.

sbenthall:
Am I reading right that `a_t` and `k_t` are both normalized here?

Neither of these is precisely the same as `aLvl`.

I was not concerned that `aLvl` has different meaning in different models. I'm concerned that `state_vars`, a data structure that is used by the AgentType class, can have different meaning in different models.

What do you think a state variable is? In general.

sbenthall:
Yes, @mnwhite I think I understood this to be what you had in mind.

Which explains why `aLvl` has not been in `state_vars`.

This PR #1099 which @Mv77 wrote and @llorracc merged has added `aLvl` to `state_vars`.

Now there is at the very least inconsistent intentions encoded in the library.

llorracc:
I don't think it makes a great deal of sense to try to distinguish between "state" and "post_state" variables, because in practice we probably want to think of the events as occurring in a sequence in which, for example, k_{t} is determined before the permanent shock, which may be determined before or after the rate of return shock, but both of them are determined before b_{t} is determined, while the choice of c_{t} determines a_{t} ... so there are many stages.  The key principle is that any variable should only have one value in a given period.  What has proven to be confusing is the need to refer to, say, a value of a from a prior period, a_{t-1}, as well as a value of a for the current period, a_{t}.  That's why it is useful to define k_{t} = a_{t-1}.  

With this way of doing things, we do not need, for example, to have aNrmPrv and aNrmNow to refer, effectively, to what we would now unambiguously designate as aNrm[t-1] and as kNrm[t], and aNrm[t] would be nan throughout the simulation until the appropriate stage is reached when its value is assigned.

Mv77:
I agree with the theoretical taxonomy.

The issue I and (I think) @sbenthall are trying to raise is that there are inconsistencies around HARK about what the "state" is in the sense that the variables in `state_now` do not always exactly match those in `agent.state_vars`. Also, neither of those matches the theoretical definition of "state" that we just gave. Here's the state declaration for `PerfForesightConsumerType`:
https://github.com/econ-ark/HARK/blob/9562cafef854d9c3d6b4aba2540e3e442ba6ec6c/HARK/ConsumptionSaving/ConsIndShockModel.py#L1583

Issue #1097, opened 2022-01-08
`make_shock_history` and `read_shocks` ignore draws of initial state vector


Mv77:
I believe `make_shock_history` and `agent.read_shocks = True` are meant to be used when one wants to limit the randomness of simulations for reproducibility or when estimating a model and wanting a well defined objective function.

(Are these the use cases you had in mind? Are there others?)

It currently works great for shocks, but it leaves another source of randomness uncontrolled: the initial state of newborns (e.g. `pLvl` and `aNrm` of newborns in the case of `ConsIndShock`).
You can see that regardless of whether `read_shocks == True` or not, `get_mortality()` calls `sim_births()` and draws states for newborns:
https://github.com/econ-ark/HARK/blob/1d379083cdba3ff9bbca494fb8664eb9ae4f29da/HARK/core.py#L631-L635

It would be useful (at least for me at the moment) to have a way to also fix these initial conditions, as we do for shocks.

Mv77:
I'll be working on this.

alanlujan91:
Might be related to @sbenthall's issue #1084.

llorracc:
I'm guessing that this has not come up in the use of these before because the models being simulated started with everyone having identical initial circumstances (and maybe no death).

mnwhite:
This was actually meant as a speedup mechanism. The code should already be
structured so that the same sequence of random shocks is drawn for each
agent on each simulation pass; that's why the resetRNG() or reset() methods
exist (I forget what they're called now) as part of initialize_sim().
Everything about the simulation should be reset to the same computational
state even without read_shocks=True. Objective functions based on
simulations should be stable no matter what.

What read_shocks=True allows is avoiding the *work* of re-drawing the exact
same sequences of shocks (tens of) thousands of times. If the same shocks
would be drawn every time, why not just draw them once, write them down,
and then look them up later. The shock-drawing process can be a non-trivial
part of the computation in models with "nested states", or shock
distributions that depend on (e.g.) a discrete Markov state or another
exogenously evolving state variable.

Mv77:
@mnwhite thanks, I am glad that is the case.

The one case I think might break this is the following.

Suppose that I have two models: A and B. And suppose B nests A in the sense that B has the same shocks and state variables as A but an additional state variable to be initialized. 

If I wanted to run a simulation of agents of type A and B whose common states are initialized in the same (random) values and are hit by exactly the same shocks, I would do something like

```
agent_B = type_B()
[...]

PR #1101, opened 2022-01-10
Include newborns' initial conditions in `make_shock_history` and `read_shocks==True`

Mv77:
Fixes https://github.com/econ-ark/HARK/issues/1097.

`make_shock_history` now creates an additional agent property called `newborn_init_history`, a dictionary of `T_sim x AgentCount` arrays, one for each state variable. The arrays are full of `nan`s except for times, agents and states that are set when an agent is reborn.

For example, if agent 35 is re-born at time 7, then `newborn_init_history['aNrm'][7,35]` will be the `aNrm` value that has to be used to initialize the re-born agent.


Mv77:
An annoying aspect of this solution is that I have to have `nan` arrays for all state variables, not only those that are used to initialize agents. For instance, `IndShockConsumerType` agents require only `[aNrm, pLvl]` to be initialized, but they have many other states (`mNrm`, `PAggLvl`, `bNrm`...).

HARK currently does not declare which subset of the state variables are required to initialize an agent, and that is why my general solution creates arrays for all states. An alternative would be to give `AgentType` a new attribute called something like `newborn_vars`: a list of strings indicating which of the state variables are drawn by `sim_births`. I'd like to get people's thoughts on whether this would be a better solution.

sbenthall:
I see how this is connected to the issue in #1099 

I suppose 'post_states' used to be those states that required newborn initialization.

Assuming that `k_t` is used, then I suppose we can remove post_states entirely.

I believe that the state variables that "must" be drawn/filled for newborns can be identified from from dependency graph implied by the equations.

Mv77:
Yes,

I believe the variables would be those needed by `get_states`:
https://github.com/econ-ark/HARK/blob/9562cafef854d9c3d6b4aba2540e3e442ba6ec6c/HARK/core.py#L541-L559

Once the graph machinery works, that could be automated.

How is this as a temporary solution?

HARK/core.py
```diff
@@ -513,6 +514,17 @@ def initialize_sim(self):
...
         self.sim_birth(all_agents)
+
+        # If we are asked to use existing shocks and a set of initial conditions
+        # exist, use them
+        if self.read_shocks and bool(self.newborn_init_history):
+
+            for var_name in self.state_now:
+                # Copy only array-like idiosyncratic states. Aggregates should
+                # not be set by newborns
+                if hasattr(self.state_now[var_name], "shape"):
```
sbenthall, line 525: I am a little skeptical about using `hasattr(*,"shape")`, which you use a few times in this PR, as the test for being array-like.

Are we being so inconsistent in the code that a proper type check won't work here?

e.g `isinstance(self.state_now[var_name], np.ndarray)`

HARK/core.py
```diff
@@ -594,10 +606,46 @@ def make_shock_history(self):
...
+        for var_name in self.state_vars:
+            # Check whether the state is array-like or an aggregate
+            if hasattr(self.state_now[var_name], "shape"):
+                self.newborn_init_history[var_name][self.t_sim] = self.state_now[
+                    var_name
+                ]
+            else:
+                self.newborn_init_history[var_name][self.t_sim] = self.state_now[
+                    var_name
+                ]
```
sbenthall, line 626: Looks like the same thing is being done in either condition here.

HARK/core.py
```diff
@@ -594,10 +606,46 @@ def make_shock_history(self):
...
+                for var_name in self.state_vars:
+                    # Check whether the state is array-like or an aggregate
+                    if hasattr(self.state_now[var_name], "shape"):
+                        self.newborn_init_history[var_name][
+                            self.t_sim, self.who_dies
+                        ] = self.state_now[var_name][self.who_dies]
+                    else:
+                        self.newborn_init_history[var_name][
+                            self.t_sim, self.who_dies
+                        ] = self.state_now[var_name]
```
sbenthall, line 646: This extra machinery seems to be for the case where a model is keeping an 'aggregate' value is being tracked by the agent.

My understanding is that this is done with some things, like PermAggShock, but that this is questionable design, because really that could be seen as 'outside' the agent.

Can you describe a use case for when a newborn needs an aggregate value assigned to it at birth?

sbenthall:
I've added a few comments inline.

HARK/core.py
llorracc: Here's a use case:  One scenario in which financial market uncertainty can be a big deal is if aggregate financial risk is correlated with individual labor income risk.  Like, unemployment risk is high in a period when there has been a big rate-of-return shock.  That would require agents to know what has happened to the aggregate shock.  If we had a richly and carefully designed structure in which it was always going to be easy and obvious how to query the parent "economy" object in which an agent was embedded in a given period to find out what the shocks were, then maybe we wouldn't need to store them.  But it is definitely easier on the programmer to just be able to retrieve the info directly from the agent directly.

sbenthall: Ok, thank you. That's a very clear use case.

What I'm trying to understand now is how `make_shock_history` is being changed in this PR to support that use case.

`make_shock_history` is called by the user in order to specify this pre-loaded shock behavior. The first thing it does is call `initialize_sim`. `initialize_sim` is being changed to update the state variables with the first line of the newborn initial history:

`self.state_now[var_name] = self.newborn_init_history[var_name][0]`

Then `make_shock_history`, a little above this line, changes the `newborn_init_history` data for time step `self.t_sim`. Since all this is being called just after the simulation is initialized, I think that `self.t_sim = 0` here. So, it's just copying the data back from `state_now` back into `newborn_init_history`.

Then in this last block... it's looping through all the ('future") time steps of the simulation, updating the `newborn_init_history` data, but now drawing from the `state_now` data, which is the timestep 0 data for all agents.

So I think what's going to happen is that after this code is run, `newborn_init_history` will, for agent X at time step 100 (for example), have the same newborn state values as agent X at time step 0. This would be true for both the idiosyncratic agent state and the aggregate state.

So I'm quite confused -- why would agent X, reborn at time 100, have the aggregate state from time 0, when agent Y, who is not reborn then, would have a different aggregate state (that from time 100).

Am I missing something?

HARK/core.py
Mv77: Fair, and better. Thank you!

An issue that I have yet to resolve is that in some models (`ConsIndShock`) an aggregate state like `PLvlAgg` will be a float, but in others (like `PortfolioFrame`) it will be an `np.array` of length 1. So yes we are being inconsistent!

HARK/core.py
Mv77: Good catch, thanks

HARK/core.py
Mv77: > This extra machinery seems to be for the case where a model is keeping an 'aggregate' value is being tracked by the agent.
> 
> My understanding is that this is done with some things, like PermAggShock, but that this is questionable design, because really that could be seen as 'outside' the agent.
> 
> Can you describe a use case for when a newborn needs an aggregate value assigned to it at birth?

I think Chris's example above works.

But for my own simpler example with the current classes that are there:

Imagine an `IndShock` agent type with aggregate growth. Newborns' `pLvl`s are initialized taking the aggregate `pLvl` into account. It is therefore important to keep track of it, because it is needed to initialize the newborns.

Where should it be stored? Not sure. I agree it is not a property of any individual agent. But the actual agent type is in charge of tracking and updating it, even if there is no Economy class. That's why I am storing it there.

Mv77: > So I'm quite confused -- why would agent X, reborn at time 100, have the aggregate state from time 0, when agent Y, who is not reborn then, would have a different aggregate state (that from time 100).
> 
> Am I missing something?

You are right, its a bug. It should be `t` (the variable changing in the for loop), not `self.t_sim`

Mv77: Wait, nevermind. `Agent.t_sim` is being advanced inside the loop
https://github.com/Mv77/HARK/blob/6c615d20a062af41c22a8b14de33f05655c3f26f/HARK/core.py#L653
So I believe `t_sim` must track `t`. Not sure why I am updating `t_sim` and not using `t`. Definitively merits a review on my end.

Thank you!

Mv77: @sbenthall in any case, the (poorly executed) idea is the following.

Before running a full simulation, I want to
a) Draw all the mortality shocks for the simulation.
b) Draw the states that the newborns will get when they die.
c) Store those states.

I achieve that by creating arrays full of `nan`, one for every state. Then running a for through the "future" simulation steps. At every step I:
- Drawing mortality shocks and store them.
- Draw states for agents who die, using `get_mortality`, which puts those states in `state_now`.
- Copying those states *only for agents who died* into my `nan` arrays.

HARK/core.py
llorracc: @sbenthall, I wonder if it is your view that we should never have used floats at all but instead `np.array` of length 1 pervasively throughout, and you are trying to move us in that direction with new code you write?  I might buy this as a reasonable proposition if it were articulated ...

sbenthall: I see. I hadn't realized I had introduced that inconsistency.

@llorracc in terms of my preferred direction...

...in the Frame model code, it's the frame that carries the information that its variables are aggregate. This gets used directly by the simulation code instead of doing a type check on the state value:

https://github.com/econ-ark/HARK/blob/master/HARK/ConsumptionSaving/ConsPortfolioFrameModel.py#L189-L192

https://github.com/econ-ark/HARK/blob/master/HARK/frame.py#L112-L117

Mv77:

Mv77:
This is ready for re-review.

@sbenthall recommended me not to aim for the perfect solution that tried to go into aggregate states and their handling, but to go for whatever works now and revise once the higher-up design questions about states, aggregate, and post-states handling have been resolved. That was a great liberating piece of advise.

I did add a neat test. It creates a shock history, but then manually modifies its shocks and initial conditions to things that make it easy to check whether they are indeed being used. It then uses the shocks to simulate and compares the results with the easy-to-compute predicted behavior.

llorracc:
@sbenthall, I guess you will review this again?  Thx.

sbenthall:
Awesome. Thanks for all this great work @Mv77 . I've reviewed and will merge the PR.

PR #1104, opened 2022-01-19
IndexDstn rework

Mv77:
The current implementation of `IndexDstn` makes its random-number-generation work slightly different to what the majority of the HARK code currently does. Find the details in the discussion in https://github.com/econ-ark/HARK/pull/1024.

There is no right or wrong way to handle the RNG. Some might prefer what IndexDstn does, others might prefer what the legacy HARK code does. In the end both ways of doing things
a) Represent the same distributions when solving the agents' problems.
b) Draw shocks that follow the same (correct) distributions.

The issue is that the exact value of individual shocks is altered when one tries to replace legacy HARK code with the newer `IndexDstn` tool, and this breaks simulation tests.

This PR changes `IndexDstn` so that the user has the option of creating a new RNG (current behavior) or use an existing one (legacy HARK tests) to draw the seeds of the time-specific distributions that compose the `IndexDstn`.

This (I hope) will allow me to use `IndexDstn` and still pass tests. Eventually, if one random-number-generation method is favored over the others, we can go back and alter the tests accordingly (being sure that changes come only from RNG).

### DIFF
--- a/Documentation/CHANGELOG.md
+++ b/Documentation/CHANGELOG.md
@@ -15,7 +15,8 @@ Release Date: TBD
 ### Major Changes
 
 * Updates the DCEGM tools to address the flaws identified in [issue #1062](https://github.com/econ-ark/HARK/issues/1062). PR: [1100](https://github.com/econ-ark/HARK/pull/1100).
-,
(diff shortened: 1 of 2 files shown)


Mv77:
The PR also introduces a slight change in how the surrogate distributions are stored.

Previously, `MyIndexDstn[n]` would call the constructor with the nth set of parameters each time it was called.

With this change, the distributions with all sets of parameters are created at `__init__` and then `MyIndexDstn[n]` returns `self.dstns[n]`

PR #1108, opened 2022-01-25
ConsRiskySolver

alanlujan91:
First pass at `ConsRiskySolver` as requested in #1107

Still a work in progress as I have to figure out if all the additional functionality of `ConsIndShockModel` is appropriate for a model with risky asset returns. For example: 
- Human wealth: Should it be divided by average rate of return? 
- MPCs still apply as before?
- Limiting functions are still as before? 

This PR also showcases the use of a dataclass as a solver class. I like this a lot more than the enumeration of parameters in `__init__` and it doesn't lose the `get_arg_names` functionality. 


HARK/ConsumptionSaving/ConsRiskyAssetModel.py
```diff
@@ -239,6 +244,210 @@ def get_shocks(self):
...
+
+        self.PatFac = (
+            calc_expectation(self.RiskyDstn, pat_fac_func) * self.DiscFacEff
+        ) ** (1.0 / self.CRRA)
+
+        self.MPCminNow = 1.0 / (1.0 + self.PatFac / solution_next.MPCmin)
+
+        self.hNrmNow = (
+            self.PermGroFac / self.Rfree * (self.Ex_IncNext + solution_next.hNrm)
+        )
```
alanlujan91, line 297: @llorracc still thinking about what the appropriate discounting is for human wealth. 
case 1 - divide by `self.RiskyAvg`
case 2 - calculate expectation of `self.PermGroFac * (self.Ex_IncNext + solution_next.hNrm) / risky_shock`

llorracc: I think we decided to impose explicit liquidity constraints, so if I understand things the question is really just about the right interest factor to use to discount to end-of-$t$ the minimum possible income in $t+1$.

If we are going to have only one return factor, the answer is that the minimum notional $a_{t}$ must be such that $a_{t} \times \max{RiskyNrm_{t+1}} = -\min{\theta_{t+1}}$ so we can guarantee with certainty that $a_{t} \times \RiskyNrm_{t+1} + \min{\theta_{t+1}} \geq 0.$

where $\RiskyNrm_{t+1} = \Risky_{t+1}/(\PermGroFac_{t+1} \times \PermShk_{t+1})$ and $\theta$ is the transitory shock.

alanlujan91: I understand this point, but my question is not related to the borrowing constraint. Human Wealth and the MPCs are used when constructing the limit of the consumption function as in https://github.com/econ-ark/HARK/blob/4bc59842cda28aef6f3382cd6ebca7aa7a5c7df3/HARK/ConsumptionSaving/ConsIndShockModel.py#L731-L732

alanlujan91:
This PR now provides a working solver for the `RiskyAssetConsumerType` which faces a risky return on assets and an artificial borrowing constraint of 0.0. 

To do:
- Appropriate discount factor for human wealth
- Verify parameters for limiting consumption function

alanlujan91:
`RiskyAssetConsumerType` now includes `PortfolioBool` option to solve a simple portfolio choice between safe and risky asset. 

I also include an example notebook to compare solution with `PortfolioConsumerType` solution.

alanlujan91:
Getting some failed tests due to some small differences in hardcoded solutions. I am trying a slightly different ShareGrid to find the solution. Since we know the the ShareLimit, it makes sense to me that the Share Grid be between ShareLimit and 1.0 instead of between 0 and 1. @mnwhite I think you wrote the original code, any reason I might not want to do this?

HARK/ConsumptionSaving/ConsPortfolioModel.py
```diff
@@ -173,6 +172,8 @@ def __init__(self, verbose=False, quiet=False, **kwds):
         params.update(kwds)
         kwds = params
 
+        self.PortfolioBool = True
```
sbenthall, line 175: Where is the documentation for this variable? It could go in the docstring.

Why is this variable named in CamelCase rather than snake_case?

examples/ConsPortfolioModel/example_ConsPortfolioModel.py
```diff
@@ -283,8 +302,8 @@ def RiskyShareMertSamLogNormal(RiskPrem, CRRA, RiskyVar):
     agent.solve()
 
     # Compute the analytical Merton-Samuelson limiting portfolio share
-    RiskyVar = agent.RiskyStd ** 2
-    RiskPrem = agent.RiskyAvg - agent.Rfree
+    RiskyVar = agent.RiskyStd**2
+    RiskPrem = agent.RiskyAvg - agent.Rfree[0]
```
sbenthall, line 306: Is the lack of parallelism between RiskyAvg (a scalar) and RFree (a time-varying something) at all a cause for concern?

HARK/ConsumptionSaving/ConsIndShockModel.py
```diff
@@ -470,26 +473,26 @@ def add_mNrmStE(self, solution):
         # All combinations of c and m that yield E[PermGroFac PermShkVal mNext] = mNow
         # https://econ-ark.github.io/BufferStockTheory/#The-Individual-Steady-State
 
-        PF_RNrm = self.Rfree/self.PermGroFac
+        PF_RNrm = self.Rfree / self.PermGroFac
```
sbenthall, line 476: In general ... the reformatting across the library makes it harder to figure out the changes made in this specific PR.

You might consider making a separate PR with just these formatting changes, which could be merged easily.

That would help reviewers focus on the substantive changes in code.

HARK/ConsumptionSaving/ConsRiskyAssetModel.py
```diff
@@ -239,9 +357,919 @@ def get_shocks(self):
         self.get_Adjust()
 
 
-# %% Initial parameter sets
+# This is to preserve compatibility with old name
+RiskyAssetConsumerType = IndShockRiskyAssetConsumerType
+
+
+class FixedPortfolioShareRiskyAssetConsumerType(IndShockRiskyAssetConsumerType):
```
sbenthall, line 364: It strikes me as odd that:
 - IndShockRiskyAssetConsumerType has two 'modes' : strategic share variation, or with a fixed rate at 1
 -  FixedPortfolioShareRiskyAssetConsumerType is subclass of the above class, with a variable fixed rate.

If I'm not mistaken, the fixed-rate-of-1 agent is a special case of the fixed-rate agent, and so logically, the former should be a subclass, if not a specific parameterization, of the latter

Also...these class names are getting to be a little ridiculous. There has got to be a way to shorten them. This may entail a larger conversation about model class naming conventions.

HARK/ConsumptionSaving/ConsRiskyAssetModel.py
```diff
@@ -239,9 +357,919 @@ def get_shocks(self):
...
+            ConsFixedPortfolioIndShkRiskyAssetSolver
+        )
+
+
+####################################################################################################
+####################################################################################################
+
+
+@dataclass
+class ConsIndShkRiskyAssetSolver(ConsIndShockSolver):
```
sbenthall, line 387: I don't have a deep insight into this, but superficially, I'm surprised that the three solver classes are each so long.

I would have expected more of an opportunity for code reuse between these solvers, given their similarity.

I wonder whether there's any functional redundancies in these classes that could be factored out.

examples/ConsPortfolioModel/example_ConsPortfolioModel.py
llorracc: I think RiskyAvg depends on an input of the portfolio share, and thus is not something whose dynamics are predetermined like those of Rfree?  But if that's right, shouldn't it be Rport?

HARK/ConsumptionSaving/ConsPortfolioModel.py
alanlujan91: I tried to name it in a similar pattern to `CubicBool` and `vFuncBool`

examples/ConsPortfolioModel/example_ConsPortfolioModel.py
alanlujan91: `RiskyAvg` is a model parameter, but not a solver parameter, so it is handled differently than `Rfree`. Depending on it's nature, the agent constructs a `Distribution` that will be set to time varying or time invariant. `Rfree` however is now expected to be time invariant, so even when it's the same throughout the lifecycle, as is here, the agent class holds the value as a list.

What @llorracc is talking about is the method `get_Rfree()` which does make a weighted combination of the risky return (distributed ~ lognormal) and the `Rfree` to return `Rport`. 

In this way, `Rfree` behaves like `PermGroFac` or `LivPrb` which are always lists and are solver parameters. 

Perhaps what you are implying is that `RiskyAvg` should also be handled as time-varying, as are the income distributions.

Issue #991, opened 2021-04-01
Terminology and code around time


sbenthall:
In the meeting today, CDC suggested we get a consensus around the use of terms relating to time, with an eye to seeing if we can simplify some of the code around time.

The time-handling in HARK comes up in some other tickets, such as : #890 #983 #798 #872 #718 #571 #95

As a kind of draft on this, I thought I'd jot down my current/preferred understanding of relevant terms.

 - **period** and **time** refers specifically to the _t_ primitive used in the [stochastic dynamic programming/Bellman equation paradigm](https://github.com/sbenthall/sketches/blob/master/economics/Mathematics%20of%20Discrete%20Time%20Stochastic%20Control%20Processes%20in%20Macroeconomics.ipynb)
   - If there is more than one control variable within a period, the problem may be broken down into a sequence of Bellman equations with other primitives besides _t_, such as _s_.
   - These complete Bellman sub-problems are **stages**.
 - **age** refers to an agent-specific variable that increments deterministically with time. In a multi-agent model, there can be multiple agents with different ages.
 -  a **lifecycle**  element is an element of a problem that varies by agae.
    - for example, *lifecycle parameters* are parameters that vary over a problem by age, perhaps according to an age-dependent process.
    -  a problem might also vary with age. For example, the decision of whether or not to go to college can be limited to a specific age range.
 - A model has a set of **model variables** and **transition equations** defining the structural transitions between the variables. These typically recur each period.
 - **frame** refers to a set of transitions that change the state of model variables. At its most granular, a frame can be the update of one model variable, such as a shock.
     - The frames together are a partition of the problem's model variables.
     - The members of each part should be adjacent/contiguous in the graph implied by the models' structural equations.

[...]

Issue #890, opened 2020-12-22
better automation of for loops over 'time varying' values


sbenthall:
Time varying values are currently implemented as lists indexed by `t`.

This is to denote differently _aged_ agents.

In some cases, there is a separate control flow for handling the time-varying values and the non varying values.

It would perhaps improve readability to have a more general, consistent way of handling time-varying values.

For example, if these time-varying values were arrays and not lists, it would be possible to perform vectorized operations on them, rather than use a (slower) `for` loop iteration.

mnwhite:
They're lists because sometimes their elements are not compatible with
being in an array.  E.g. PermGroFunc in ConsGenIncProcessModel has a list
of *functions*, which can't be cast to an array.

It's uncommon that there's a use for vectorized operations on time-varying
model inputs.  When there is, I created an array from the list and worked
with that. Any instance of a time-varying loop in the code is either
because a) it's strictly necessary as there is no vectorized equivalent, or
b) that code would only ever be run once, it takes essentially no time, and
it's significantly easier to read as a loop.

sbenthall:
I thought you were on a break from HARK work @mnwhite ...

sbenthall:
Since vectorization is not an option, it might be possible to use something like the following:

```
def apply_or_map(value, func):
    if timy_varying(value):
       return map(value, func)
    else:
       return func(value)
```

That would allow the code to be refactor in a way that operations could be written just once, rather than repeated over both the time varying and non-time varying cases.

I've notived this while working on this part of the code, but maybe there are other places as well.

https://github.com/econ-ark/HARK/blob/master/HARK/ConsumptionSaving/ConsPortfolioModel.py#L242-L262

sbenthall:
Here is another example where the handling of time-varying and non-time-varying cases could be consolidated.

https://github.com/econ-ark/HARK/blob/master/HARK/ConsumptionSaving/ConsPortfolioModel.py#L277-L286

sbenthall:
Here is another case:
https://github.com/econ-ark/HARK/blob/master/HARK/ConsumptionSaving/ConsPortfolioModel.py#L322-L342

sbenthall:
Note there is a lack of parallelism here.

If `RiskyDstn` is not time varying, it is a scalar (float).

If it's an infinite horizon model with no age variation, then the `solution` is a single element list.

It would probably be best if these were the same structure (i.e., list or not, in the singleton case), for consistency.

mnwhite:
[...]

PR #1112, opened 2022-02-14
Time varying unemployment probability and income

Mv77:
This PR allows `CondIndShock` and its derivatives to use age-varying unemployment probabilities and unemployment incomes.

This can be done passing lists as the arguments for `UnempPrb` and `ÌncUnemp`. 


Mv77:
Addresses https://github.com/econ-ark/HARK/issues/1015

sbenthall:
Adding a variable name to a list called `time_varying` does not, in itself, make the code function in a way such that, say, a list of values is interpreted in all parts of the program as a time-varying parameter.

There are many open tickets that are about make a more unified and coherent approach to time-varying parameters. #95 #664 #890 #991 among others

alanlujan91:
@llorracc I think that's correct. If we want a variable to be time-varying in some cases, even if it's in few cases and the main use case is time-fixed, the variable has to be added to `time_vary` and given as a list `[x]` of one element. Something we have been thinking about with the AgentPopulation in SHARKFin is how to flexibly infer time (and agent) variability depending on the argument passed.

alanlujan91 (approved):
Changes look good to me! 

A quick suggestion is that sometimes the types of parameters might be ints. I saw that you had to change this in some tests to make them floats. I assume this was for the type checks? 

Something that can help with type checks is `isinstance(x, (int, float))` to catch any scalar parameter.

Mv77:
Great, this is ready to merge then.

Thank you all!

PR #1115, opened 2022-02-21
Use 1 point for degenerate income distributions

Mv77:
This PR modifies the engines for transitory and permanent income shocks so that if the distribution is degenerate (`sigma == 0.0`) they use only one point to approximate it. This is useful, for instance, in retirement.

Currently, with `TranShkCount` and `PermShkCount` being constant, multiple points are being used even if `sigma==0.0`. This behavior was probably my own making, in my recent updates.

### DIFF
--- a/HARK/ConsumptionSaving/ConsIndShockModel.py
+++ b/HARK/ConsumptionSaving/ConsIndShockModel.py
@@ -2790,7 +2790,9 @@ class LognormPermIncShk(DiscreteDistribution):
     def __init__(self, sigma, n_approx, neutral_measure=False, seed=0):
 
         # Construct an auxiliary discretized normal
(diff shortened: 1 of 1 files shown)


alanlujan91:
YES! This is a great PR @Mv77! Was just talking to @iworld1991 about how I always have to remember to set PermShkCount = 1 when a random process is shut down. 

This can also help when comparing `RiskyAsset` solutions when we shut down uncertainty.

Mv77:
Ha! Thanks for the endorsement!

sbenthall:
Review discussion on Feb. 23 2022 : It's a little odd how this interacts with the baked-in constant income for retirement, but as it stands this PR is just fine and should be merged.

PR #1117, opened 2022-02-22
Separate FrameModel from FrameAgentType

sbenthall:
This PR makes a number of changes to the "Frame" code in order to improve its flexibility as a way of defining models. A few highlights from the work:

 - A `FrameModel` class, which holds the Frames and parameters, but has no solution or simulation functionality. (See #495)
    - This class keeps track of whether or not it is an `infinite` model. (This is akin to the `cycles` parameter, but a little less confusing).
    - `FrameModel`s can be repeated and combined. This is building towards a more flexible way of composing problems from subproblems -- the subproblems can be built as FrameModels and then glued together.
 - A `FrameSet` data structure, specifically for containing multiple frames. This solves the problem that frames may have multiple variables as targets, but need to be looked up by single variables and index/order sometimes. An extension of `OrderedDictionary`.
 - `FrameAgentType` now takes a `FrameModel` as input and is designed to handle the simulation functionality in a general way, abstracted over `FrameModel`.
 - `control` frames no longer take a transition function. Instead, they get a standard transition function that looks up a `decision_rule` for the frame target.
 - `ConsPortfolioFrameModel` has been adapted so now, when the model is `solve()`ed, the HARK `solution` object is reformed into decision rules that the simulation code can use.
 - The method for visualization a `FrameModel` as an [influence diagram](https://en.wikipedia.org/wiki/Influence_diagram) has been moved to the library as `draw_frame_model()`

This PR has some new tests for the FrameModel functionality.

Requesting a review from @Mv77 .


sbenthall:
The graph layout algorithm that is most useful for rendering influence diagrams is available in `networkx` via the `pydot` package, which is a wrapper around the C package `graphviz`.

Unfortunately, it is not easy to automatically install graphviz in a cross-platform way, and this is causing some of the notebooks to fail.

I'll fix this to get the tests to pass.

sbenthall:
Ok, tests pass now. It's ready for review.

Mv77 (approved):
@sbenthall,

I just went through this.

I must confess that I do not fully understand the specifics (context, scope, references) of the architecture in place, or its advancements, so I am unable to give specific feedback on that for now.

However, I was able to see the general purposes of the PR and I like what I see in terms of making things modular, allowing models to be integrated, visualization tools, infrastructure for solvers ... etc

So I don't get the specifics, but I like the general things that I see. I approve.

sbenthall:
Thank you, @Mv77 . That's good feedback. I surely need to provide better documentation for what's going on here. I will see what I can do about that before merging.

Issue #1105, opened 2022-01-20
functional simulation tests to enable external distribution classes


sbenthall:
Continuing from this discussion:
https://github.com/econ-ark/HARK/pull/1104#issuecomment-1017620775

I believe that @llorracc is keen on replacing as much of the distribution modeling functionality in HARK with an external library as soon as possible.

A major obstacle to doing this is that the tests for model simulations currently mostly check for near equality with specific numerical values, as opposed to testing for a functional relationship between a model's state value and prior relevant state values.

Fixing this is just a matter of labor time.

This would make it much easier to work on #611 #949

sbenthall:
Though it's not entirely clear from the documentation, I've tested it and this `scipy.stats.rv_discrete` class is more full-featured that our current `DiscreteDistribution` class -- including CDF, entropy, moment, and _expected value_ computation -- with an initialization parameter, `values`, that mirrors the way we are currently constructing DiscreteDistributions.

https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.rv_discrete.html

llorracc:
Actually, I've spent a little time in the last couple of days looking into this and am glad to report that I think maybe the path forward is getting clearer, because sympy has recently put some effort into creating some tools that I hope we can build on for our purposes.  [Encouragingly, they started with what Mathematica had done](https://github.com/sympy/sympy/issues/16895), but seem to have ended up building something somewhat simpler, which may be subject to future improvements.  There is also an excellent treatment of the topic on the quantecon site, https://python-advanced.quantecon.org/stationary_densities.html.  

I also came across [this recent paper](https://arxiv.org/pdf/2112.14315.pdf) which claims to have made some fundamental advances in the efficiency of simulating Markov processes.

One thing that is somewhat confusing about this whole area is where exactly the boundaries lie between Markov chains and optimal control problems.  The math-stats people seem mainly to be interested in the question "give me some fixed Markov process, and I'll tell you all sorts of things about its consequences."  Optimal control people (and, mostly, economists) are interested in the question, "tell me the environment someone is in, and I want to figure out the optimal behavior."  Once optimal behavior has been figured out, maybe that gives you some fixed coefficients that can then be represented in a Markov chain, but the Markov chain part of it is in a sense easier than the 'find the optimal behavior' part.  Still, the common element to all of them is that we need to find a robust way to state what the environment is and how it evolves in response to exogenous and endogenous variables; maybe the new sympy objects will be useful for that.

[...]

PR #1121, opened 2022-03-01
[WIP] Arbitrary distributions of newborns' states

Mv77:
It would be useful for us to let users specify whatever distribution they want for drawing the initial states of newborns.

For instance, I would like to draw the initial savings and permanent income levels of an `IndShockConsumerType` agent from some distribution calibrated to the SCF, not necessarily two independent lognormals as the current code requires.

This should be easy. Give me a `Distribution` object with the right dimensions and I will simply draw from it whenever agents are born. The issue, as usual is the current simulation tests (https://github.com/econ-ark/HARK/issues/1105).

This PR will contain my work towards the general-distribution-object purpose. I will try to find an elegant way not to break existing tests.


sbenthall:
Hooray. Great idea. cf #620

PR #1122, opened 2022-03-08
Approximate share in Portfolio Model

alanlujan91:
Linear approximation of risky share given next period's cFunc.


alanlujan91:
image

Blue line is HARK's numeric solver. Orange line is approximation.

HARK/ConsumptionSaving/ConsPortfolioModel.py
```diff
@@ -822,6 +824,74 @@ def make_ShareFuncAdj(self):
...
+
+        def premium(shock):
+            """
+            Used to evaluate mean and variance of equity premium.
+            """
+            r_diff = shock - self.Rfree
+
+            return r_diff, r_diff ** 2
+
+        prem_mean, prem_var = calc_expectation(self.RiskyDstn, premium)
```
Mv77, line 854: @alanlujan91 not sure if this is a bug given your notation, but I think that what your are calling `prem_var` is not the variance of the premium, but rather E[premium^2].

llorracc: It would be interesting to see what happens when you make the maximum amount of assets something much larger, like 2000 or 10000 or something.  Out that far, the consumption function is essentially linear, so the error coming from the approximation's assumption that $c^{''}$ is zero should be inconsequential.  If the orange does converge to the blue, that would suggest that maybe using the 2nd order Taylor approximation would make the results considerably closer for small values of wealth.  If orange does not converge to blue, I'd guess that there's a bug somewhere.

alanlujan91:
0. large value a_max
0. second order approximation (1+\epsilon)^\nu in Chris's notes (MathFacts TaylorTwo)
0. Compare second order to true exp w/ large value of a_max, are they close?
0. implement backsolving using only approximations

alanlujan91:
Did a little bit of work on this. It seems the second order approximation with `TaylorTwo` rule is not particularly better than the first order approximation. Where first order approx is below true function, second order approx is above the function.

image

alanlujan91:
@llorracc this is with aMax = 1000 (instead of 100). 

image

alanlujan91:
> @alanlujan91
> 
> Do you know why the checks are failing?

Yes, the tests for Discrete Solver and Joint Distribution solver are failing, but this is not supposed to be a feature of those models. I have to reorganize where the approximations happen.

HARK/ConsumptionSaving/ConsPortfolioModel.py
alanlujan91: @Mv77 you're right, I have called this prem_sqrd instead.

llorracc:
To debug, I suggest you evaluate the last expression in (15) in the appendix to

https://github.com/ccarrollATjhuecon/EquityPremiumPuzzle/blob/master/EquityPremiumPuzzle.pdf

at $a=1000$ and the first expression in equation (16) at the same point.  

# If the first expression in (16) is not close to zero
 - you can break both 15 and 16 down using the $E[xy] = E[x]E[y]+cov(x,y)$ formula and see whether it is the $E[x]E[y]$ term or the $cov(x,y)$ term or both that are bad.  
 - You could also evaluate 15 using a linear consumption function identical to the one used in (16);
 -  if the discrepancy remains largely the same, that would say that the problem is not coming from the nonlinearity of the consumption function but from the approximation of the shock.  
# If the first expression in (16) IS close to zero
  - That would seem to indicate that the problem is the approximation between the penultimate and the last line in (16).  You could see whether a 2nd (or even a 3rd) order approximation at that point fixes things.

PS.  Whether or not you manage to track down the problem as above, it would be useful if you could edit the GitHub document to incorporate any higher order approximations that you are using so we can see exactly what you are doing.

Issue #1125, opened 2022-03-14
Newborns' transitory shocks to income


Mv77:
@AMonninger noticed a strange behavior during a class exercise. His newborns' transitory shocks to income were always set to 1.0.

He asked me why this was and, looking into it, I found this line of code:
https://github.com/econ-ark/HARK/blob/600372f49203a37b3211f5fca19b424d19ea3144/HARK/ConsumptionSaving/ConsIndShockModel.py#L2224
The line seems to enforce this behavior.

I wonder if this is desired (if so, why?) or a bug. It is definitely not documented and not what an user would expect.

Looking at the line's history, it appears that it existed in a very old version of hark but was commented, and then Edmund uncommented it when he merged a branch related to the `StickyE` paper.

Mv77:
Tagging @llorracc @mnwhite to ask whether they know why this was introduced and ask whether they'd like to keep it.

This would be a pain to fix given our simulation test issues.

mnwhite:
It was partly originally due to the code lacking a concept of the magnitude
of permanent and transitory shocks in the newborn period; at some point,
permanent shock might also have been set to 1.0 in the newborn period.

Edmund likely uncommented it for the StickyE paper for a very particular
detail of the welfare calculation in that paper. Part of the motivation is
that zero income is possible in that paper (or was possible at some point),
so drawing zero income in the first period gets you negative infinite
value. If there's any chance of that happening, the "pre-existence value of
life" is undefined.

Mv77:
@lloracc told me that this behavior should at least be optional.

We'd like to create a fix that makes this line optional through some new parameter. (It might be eventually removed when the simulation tests and newborn handling are revamped).

@AMonninger, Chris said that making a pull request fixing this would be a nice exercise for you. Want to have a go?

AMonninger:
Sure, I'll give it a try.
My approach would be: Default option would be that all newborns get both shocks with the option to have only one, which can be specified as either permanent or transitory.

Mv77:
@llorracc "for now we want the default to stay the way it is, because otherwise it will break all of our tests."

AMonninger:
Just created a pull request:
https://github.com/econ-ark/HARK/pull/1126

PR #1126, opened 2022-03-17
'NewbornTransShk: Option if Newborns get transitory shock'

AMonninger:
Pull request for issue:
Newborns' transitory shocks to income #1125

I added an aditional variable: NewbornTransShk which is by default true. If so, Newborn consumers do NOT get a transitory shock in the first period, only a permanent one.
By setting it 'False' Newborn consumers Do get a transitory shock as well.

### DIFF
--- a/Documentation/CHANGELOG.md
+++ b/Documentation/CHANGELOG.md
@@ -25,6 +25,7 @@ Release Date: TBD
 
 * Updates the lognormal-income-process constructor from `ConsIndShockModel.py` to use `IndexDistribution`. [#1024](https://github.com/econ-ark/HARK/pull/1024), [#1115](https://github.com/econ-ark/HARK/pull/1115)
 * Allows for age-varying unemployment probabilities and replacement incomes with the lognormal income process constructor. [#1112](https://github.com/econ-ark/HARK/pull/1112)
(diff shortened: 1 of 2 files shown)


Mv77:
This is neat and passes tests.

A minor comment:
If I see an option called `NewbornTranShk` and set it to `true`, I'd expect newborns to get shocks. I'd just suggest to make it so that `true`->shocks and `false`->no-shocks, making the default `false`.

llorracc:
I agree with Mateo - true means they get a shock.

Mateo, could you merge this when Adrian revises?

Mv77:
I'd be happy to, but I don't have merge permissions.

AMonninger:
Just pushed the changes. 
Is there a way to put it in the documentation as well?

Mv77:
This now looks good. Thanks a lot @AMonninger .

@llorracc, I do not have the permissions to either approve the remaining tests (workflows?) or merge.

llorracc:
@AMonninger, for the documentation, you should also add something to the docstring listing the variable as an input and briefly explaining what it does.

Once you have done that and it passes all tests, I will merge.

AMonninger:
I added it in the 'get_shocks' function as it is the only function affected by the change. Is this what you meant?

llorracc:
@AMonninger,

Yes, I now see you've added a docstring at the right place.  However, see the syntax for optional boolean variables used elsewhere (I think you need it to say something like ": Boolean, Optional" to mark it in the documentation as optional.  Also, having done that, please rewrite the description accordingly.  (See other examples of optional variables for templates).

Issue #982, opened 2021-03-10
use performance profiler on core use cases of HARK and look for speed improvements


sbenthall:
We talk a lot about computational performance but I don't think we have been very systematic about testing for it.

Especially for computationally expensive actions like computing market equilibria, this could lead to some big performance gains.

https://docs.python.org/3/library/profile.html

sbenthall:
Setting up cProfile for core econ-ark functionality would be a good move @nicksawhney

sbenthall:
cc @alanlujan91

PR #1131, opened 2022-04-12
implement benchmARK

alanlujan91:
Simple profiler for HARK models. See #982. 

Example usage: 

```
from HARK.utilities import benchmark
from HARK.ConsumptionSaving.ConsPortfolioModel import PortfolioConsumerType
agent = PortfolioConsumerType(cycles = 0)
benchmark(agent)
```

Output: 

```
Mon Apr 11 19:58:07 2022    restats
         2447109 function calls (2232506 primitive calls) in 4.810 seconds
   Ordered by: internal time
   List reduced from 213 to 10 due to restriction <10>
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
    12308    1.111    0.000    1.111    0.000 utilities.py:221(CRRAutilityP)
    15510    0.771    0.000    0.771    0.000 {method 'searchsorted' of 'numpy.ndarray' objects}
    10080    0.539    0.000    1.075    0.000 interpolation.py:798(_evalOrDer)
     2715    0.502    0.000    0.786    0.000 interpolation.py:1377(_evaluate)
184035/45795    0.300    0.000    4.380    0.000 {built-in method numpy.core._multiarray_umath.implement_array_function}
      724    0.223    0.000    3.645    0.005 shape_base.py:267(apply_along_axis)
    20272    0.138    0.000    0.138    0.000 ConsPortfolioModel.py:603(m_nrm_next)
    19186    0.109    0.000    0.281    0.000 function_base.py:5143(insert)
      181    0.074    0.000    0.244    0.001 ConsPortfolioModel.py:731(optimize_share)
66564/1091    0.060    0.000    0.147    0.000 copy.py:128(deepcopy)
```

### DIFF
--- a/HARK/utilities.py
+++ b/HARK/utilities.py
@@ -3,7 +3,9 @@ General purpose  / miscellaneous functions.  Includes functions to approximate
 continuous distributions with discrete ones, utility functions (and their
 derivatives), manipulation of discrete distributions, and basic plotting tools.
 """
+import cProfile
 import functools
+import pstats
 
 import numpy as np  # Python's numeric library, abbreviated "np"
 
@@ -132,11 +134,12 @@ def CRRAutility(c, gam):
     >>> utility(c=c, gam=gamma)
     -1.0
     """
-    
+
     if gam == 1:
         return np.log(c)
     else:
         return c ** (1.0 - gam) / (1.0 - gam)
+
 
 def uFunc_CRRA_stone_geary(c, CRRA, stone_geary):
     """
@@ -165,9 +168,10 @@ def uFunc_CRRA_stone_geary(c, CRRA, stone_geary):
     -1.0
     """
     if CRRA == 1:
-        return np.log( stone_geary + c)
+        return np.log(stone_geary + c)
     else:
-        return ( stone_geary + c ) ** (1.0 - CRRA) / (1.0 - CRRA)
+        return (stone_geary + c) ** (1.0 - CRRA) / (1.0 - CRRA)
+
 
 def uPFunc_CRRA_stone_geary(c, CRRA, stone_geary):
     """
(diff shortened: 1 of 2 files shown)


alanlujan91:
Profiling tool for HARK models. Calling `benchmark` on agents calls the solver for
    the agents and provides time to solve as well as the top `max_print` function calls
    in terms of  `sort_by`. Optionally allows for saving a text copy of the profile
    as well as returning the `Stats` object for further inspection.

    Parameters
    ----------
    agent_type: AgentType
            A HARK AgentType with a solve() method.
    sort_by: string
            A string to sort the stats by.
    max_print: int
            Number of lines to print
    filename: string
            Optional filename to save output.
    return_output: bool
            Boolean to determine whether Stats object should be returned.

    Returns
    -------
    stats: Stats (optional)
          Profiling object with call statistics.

alanlujan91:
Note: In Table statistics, `interpolation.py` is not `EconForge`'s library but HARK's internal file `HARK.interpolation.py`.

HARK/utilities.py
```diff
@@ -1101,3 +1105,41 @@ def find_gui():
...
+    Profiling tool for HARK models. Calling `benchmark` on agents calls the solver for
+    the agents and provides time to solve as well as the top `max_print` function calls
+    in terms of  `sort_by`. Optionally allows for saving a text copy of the profile
+    as well as returning the `Stats` object for further inspection.
+
+    Parameters
+    ----------
+    agent_type: AgentType
+            A HARK AgentType with a solve() method.
+    sort_by: string
```
sbenthall, line 1123: As this is passed directly to pstats.Stats, it would make sense to have a link in the docstring to the relevant python documentation. As is, you really have to look at the code to understand how to use the parameters

sbenthall:
This is great! Well done!

The only comment is that since it basically wraps pstats, which is a bit tricky as a package, it would be best if the docstring linked over to the pstats docs:
https://docs.python.org/3/library/profile.html#module-pstats

PR #1136, opened 2022-04-19
added econforge interpolation into LinearInterp and scipy interpolati…

gms158:
…on into CubicInterp

### DIFF
--- a/HARK/fast_interpolation.py
+++ b/HARK/fast_interpolation.py
@@ -1,0 +1,16 @@
+import numpy as np
+from interpolation import interp # econ forge
+from scipy.interpolate import CubicSpline # scipy
+
+# econforge interp and then uses the hark extrapolation for values outside of the range
+def fast_linear_1d_interp(x, obj):
+    result_temp = interp(obj.x_list, obj.y_list, x)
+    above_upper_bound = x > obj.x_list[-1]
+    i = len(obj.x_list) - 1
+    alpha = (x[above_upper_bound] - obj.x_list[i - 1]) / (obj.x_list[i] - obj.x_list[i - 1])
+    result_temp[above_upper_bound] = (1.0 - alpha) * obj.y_list[i - 1] + alpha * obj.y_list[i]
+    return [result_temp]
+
+# gets the scipy cubic spline object
+def get_cubic_spline(obj):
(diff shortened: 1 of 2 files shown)


gms158:
I temporarily removed scipy CubicInterp, since it does have minor tolerance value changes, which causes errors in the unit test. I tested scipy CubicHermiteSpline, which resulted in larger tolerance issues than CubicInterp. I removed this to at least get the base LinearInterp changes implemented. I added a try catch around the econforge linear interpolation test, since this will capture the error and should not increase runtimes. If this approach is not desirable, we can try and think of cleaner ways to do this.

alanlujan91:
@gms158 because EconForge uses numba and numba is for speeding up array arithmetic, I wonder if the error is simply because the test is using ints and not arrays

https://github.com/econ-ark/HARK/runs/6178262613?check_suite_focus=true#step:5:558

Maybe Bilinear and Trilinear should check the type first and if inputs are ints then they should be cast as arrays?

gms158:
Alan-
yes, I was going to look into this more tomorrow. I may or will follow
up with a question or two for you tomorrow.
-Gry

albop:
I'm currently offline but if you want I can help review this PR next week

gms158:
I updated the interpolation code to account for infrequent and outlier cases where the econ forge library fails when evaluating one point consisting of integers, for python 3.8 and higher.

alanlujan91:
@albop @AMonninger did you want to review / have any comments on this before we merge it?

pat749 (approved):

pat749 (approved):
yes I have reviewed it's good to merge.

llorracc:
@alanlujan91, do you know any reason we should not merge this?

sbenthall:
Please add an update to the CHANGELOG describing this feature to the PR

https://github.com/econ-ark/HARK/blob/master/Documentation/CHANGELOG.md

sbenthall:
This PR needs documentation according to the NumPy / SciPy standard:

https://github.com/econ-ark/HARK/blob/master/Documentation/contributing/CONTRIBUTING.md#guidelines

sbenthall:
Superceded by #1151

PR #1144, opened 2022-05-09
Tool to find the distribution of a function of a R.V.

Mv77:
Imagine `x` is a `DiscreteDistribution` object and you want to find the distribution of `f(x)` for some `f`.

This pr introduces a tool called `distr_of_function`. `distr_of_function(x, f)` will return a `DiscreteDistribution` object that represents the distribution of `f(x)`. I wrote tests and made it work for `f: 1->1`, `f: n->1`, `f: 1->n`, and `f: n->m`. You only need `f` to return a `np.array`. 


Mv77:
I see this being useful as we move toward the representation of models using transition equations.

Imagine I'm in period `t` and have:
- A function for `v[t+1]` and its derivatives
- A distribution for t+1 shocks `ShockDstn`.
- A transition equation such that `state[t+1] = transition(shocks, state[t])`.

One could to something like `FutureStateDstn = distr_of_function(ShockDstn, transition, state[t])`. And then with `calc_expectations` compute the things we usually need, say, `calc_expectations(FutureStateDstn, dvdm[t+1])`.

Mv77:
@sbenthall could you review? This is heavily inspired on your `calc_expectations` tool, which I am a huge fan of.

sbenthall:
Yes, I can review this.

Issue #1051, opened 2021-08-03
DiscreteDistribution should have pmv not pmf


sbenthall:
```
# CDC 20210621: We should not use pmf for the point masses of the realizations, because it
# is not a probability mass 'function.' Scipy.discrete_rv, sympy, and Mathematica all
# return pmf (or pdf) as a function. They have different syntaxes for retrieving the
# vector of point masses; here it is called the pmv (probability mass vector)


class DiscreteDistribution(DiscreteDistributionOld):
    __doc__ = DiscreteDistributionOld.__doc__
    __doc__ += """
        XYZ : np.array
            Restructure the dimensions of the np.array so that successive
            columns in XYZ correspond to the vectors of values of each
            discrete random variables across the broadcasted combinations.
            Thus, [instance].XYZ[0] will return the vector of discrete
            realizations of the first variable, [instance].XYZ[-1] will
            get the last RV, etc.
        pmv : The vector of point masses of the broadcasted collection
            of random variables
"""

    def __init__(self, pmf, X, seed=0):  # Get the old definition
        super().__init__(pmf, X, seed)  # execute the old def
        XYZ = np.column_stack(X)  # stack actually unstacks them
        if self.dim() == 1:  # numpy 1 dimensional arrays want to be
            XYZ = XYZ.T  # the wrong shape
        self.XYZ = XYZ  # [ N dimensional matrix ]
        # The pmf should be a function, not a vector (as now)
        # TODO: Replace invocations of pmf with pmv
        self.pmv = pmf

 
```

sbenthall:
Also, "X" should not be "X", but a more descriptive and lowercase term, like "values"

sbenthall:
See discussion here:
https://github.com/econ-ark/HARK/pull/1146/files#r883016319

Because of the ambiguity about the semantics of "X" -- a dimension of X can mean either a dimension of the state values of of random variable, or the 'nature's choice' dimension -- it would be better if X/values were a different kind of data structure that makes these differences more explicit.

For example, if X were a Pandas dataframe, the index could be nature's choice, and the columns would be dimensions of the RV value. However, that might not work for matrix-valued (as opposed to vector-valued) RVs.

alanlujan91:
`xarray.DataArray` has attributes `values` and `data` by default. The `DiscreteDistributionXRA` could easily make this change without significant work. 

@llorracc how about `data` and `prob`?

sbenthall:
[...]

PR #1146, opened 2022-05-16
Array-valued discrete distribution

Mv77:
The current implementation of `DiscreteDistribution` allows us to represent multi-variate distributions in the sense that it can represent random variables with `n` dimensional domains (vector valued).

Something that it does not allow is for us to represent random variables whose realizations are arrays of arbitrary dimensions. Imagine, for instance, a random variable `X` that could be `np.array([[1,2],[3,4]])` with probability 0.5 or `np.array([[0,0],[0,0]])` with probability 0.5. The realizations of X have shape `(2,2)`. We can not represent `X` directly---we can in principle represent the distribution of `X.flatten()` and do the required reshapes later, but this is cumbersome.

This PR works towards allowing random variables whose realizations are arrays of arbitrary dimensions. I will explain why I think that is useful in a comment below.


Mv77:
Here is an example of why I would find this useful.

Consider the portfolio model with sticky "shares". In that model, the end-of-period variables that you need to figure out your state vector next period are `aNrm` and `Share`. Both are, or can be, choices and so what you do is you have big tiled arrays with all combinations of your grids, `aNrm_tiled` and `ShareNext`.
https://github.com/econ-ark/HARK/blob/2155175df60eaff1ee230a55e359bdd8d9ddd008/HARK/ConsumptionSaving/ConsPortfolioModel.py#L679-L682

What you then want to do is compute the expectation of various functions (`v`, `dvdm`, `dvds`) of next period's state (`(m,Sharenext)`) conditional on every combination of this period's `(aNrm,Share)`.  One way to do this that would move us in the direction of using the same transition equations/functions in the solution and simulation of the model would be to compute the distribution of `(m,ShareNext)` conditional on every `(a,Share)`  using something like

`state_next_dstn = dist_of_function(dstn = ShockDstn, func = lambda shocks: transition(shocks, a, Share))`

and then use `calc_expectations` to on `state_next_dstn` to find the expectation of `v`, `dvdm`, `dvds`.

The issue with this approach would be that we would need to have one `DiscreteDistribution` object representing the distribution of `(m,ShareNext)` for *every* combination of `(a,Share)` and would need to take the expectations one by one.

An alternative would be a single `DiscreteDistribution` object whose realizations `X` would be matrices of the same size as the "tiled" arrays. Then, `X[i,j]` would be interpreted as "next period's state conditional on the fact that you chose the i-th `a` and j-th share and whatever shock was realized." This would be a much more practical object to handle and take expectations over. The issue is that the current discrete distributions do not allow matrix-valued random variables, and that is what this PR wants to change.

Mv77:
Thanks @mnwhite. I think @sbenthall has already merged a `calc_expectations` that can be given a function as a input. If not given, it finds the expectation of the distribution itself.

What I'm going for looks something like this.
`state_next_dstn = dist_of_function(shock_dstn, lambda x: transition(x, post_state_now))`
`end_of_period_Ev = calc_expectation(state_dstn_next, vFunc_next)`
`end_of_period_Edvdm = calc_expectation(state_dstn_next, dvdmFunc_next)`
`end_of_period_Edvds = calc_expectation(state_dstn_next, dvdsFunc_next)`

mnwhite:
Yes, I *think* calc_expectations should do that in a combined line, at
least as I had envisioned it. Stepping back from the portfolio choice model
to the basic model, consider computing end of period value and marginal
value of assets:

v_func_temp = lambda psi, theta, a : DiscFac * (PermGroFac*psi)**(1.-CRRA)
* vFunc_next(Rfree/(PermGroFac*psi) * a + theta)
dvda_func_temp = lambda psi, theta, a : DiscFac * Rfree *
(PermGroFac*psi)**(-CRRA) * dvdmFunc_next(Rfree/(PermGroFac*psi) * a +
theta)
end_of_period_v = calc_expectation(v_func_temp, income_shock_dstn,
aNrm_now_grid)
end_of_period_dvda = calc_expectation(dvda_func_temp, income_shock_dstn,
aNrm_now_grid)

I might be getting the order of inputs wrong, but it's roughly that. In
this case (as well as the portfolio choice model), just having the
distribution of future states (m_{t+1}) isn't enough to compute
end-of-period (marginal) value of assets. Because (marginal) value is
scaled by (a power of) permanent income growth, the shock distribution
still has to be known as well when taking those expectations. That's why I
made (my version of) calc_expectation be able to take entire arrays of the
function inputs that *aren't* stochastic.

Mv77:
@mnwhite you are completely right that to take the expectation one would need to work with the permanent shock too. The way I handle that in a proof of concept for the portfolio model was to make the permanent shock part of the "state" of which we are finding the distribution at this stage. It is only a "state" for taking the expectation, I'm not saying that I made it an actual state. Still, it does make things slightly slower.

In terms of speed, the specialized code that we have for each particular model will always win. But I still think the tool is valuable for things like expressing a model with a transition (or a belief about a transition) that can be switched for another with the least pain possible. Having the option won't hurt.

Mv77:
That is not the sole purpose of the PR, however. Even if my transition setup is not what we want to achieve, enforcing discrete distributions' values to be an `np.ndarray` is valuable.
- Knowing the type allows the methods for this class to not have to check whether it is a list, an array, or something else and that makes the code more elegant. See for instance the `.draw()` and `combine_indep_dstns` methods in this PR.
- It brings us closer to being able to use `xarrays` which are a tool that Seb, Alan, and Chris want to use. The big advantage (I hear) is being able to refer to dimensions by name, and that would make for more readable and orderly code.

Mv77:
Base tests are passing. I still need to

- [ ] Fix the examples that are currently failing.
- [ ] Update the class description and documentation.

Mv77:
This is ready for review (or further thoughts).

sbenthall:
@sbenthall I'll review this.

HARK/distribution.py
```diff
@@ -1305,48 +1289,43 @@ def calc_expectation(dstn, func=lambda x: x, *args):
...
     """
-    N = dstn.dim()
-
-    dstn_array = np.column_stack(dstn.X)
-
-    if N > 1:
-        # numpy is weird about 1-D arrays.
-        dstn_array = dstn_array.T
-
-    f_query = np.apply_along_axis(func, 0, dstn_array, *args)
```
Mv77, line 1336: @sbenthall keep the behavior of *args in mind when reviewing. It has changed!

Mv77: A quick search does suggest that `map` with lambda functions is quite slow.

Do you think there is a better, faster way to do it? The issue I was trying to get around is that `np.apply_along_axis` requires the function's inputs to be 1-d

Mv77: Seems like list comprehensions would work better:
https://stackoverflow.com/questions/1247486/list-comprehension-vs-map

Let me run some tests before merging.

HARK/ConsumptionSaving/ConsRiskyContribModel.py
```diff
@@ -1206,7 +1206,7 @@ def post_return_derivs(inc_shocks, b_aux, g_aux, s):
 
         # Define grids
         b_aux_grid = np.concatenate([np.array([0.0]), Rfree * aXtraGrid])
-        g_aux_grid = np.concatenate([np.array([0.0]), max(RiskyDstn.X) * nNrmGrid])
+        g_aux_grid = np.concatenate([np.array([0.0]), max(RiskyDstn.X.flatten()) * nNrmGrid])
```
sbenthall, line 1209: It seems like a lot of extra code is now needed to flip vertical vectors/arrays into horizontal ones.
Would it be possible to have the arrays oriented horizontal in the DiscreteDistribution object?

HARK/distribution.py
```diff
@@ -614,17 +614,15 @@ def approx(self, N, equiprobable=False):
...
         # Apply mean and standard deviation to the Z grid
-        X = np.tile(np.reshape(self.mu, (1, self.M)), (N ** self.M, 1)) + np.matmul(
-            Z, A.T
-        )
+        X = self.mu[None,...] + np.matmul(Z, A.T)
 
         # Construct and return discrete distribution
         return DiscreteDistribution(
-            pmf, X, seed=self.RNG.randint(0, 2 ** 31 - 1, dtype="int32")
+            pmf, X.T, seed=self.RNG.randint(0, 2 ** 31 - 1, dtype="int32")
```
sbenthall, line 625: this also seems like something the user of DiscreteDistribution would rather not have to keep track of.
Is the issue that a 1-dimensional array is now ambiguously a 1d distribution with N buckets, or an N-D distribution with 1 bucket?

HARK/distribution.py
```diff
@@ -821,9 +819,11 @@ class DiscreteDistribution(Distribution):
     ----------
     pmf : np.array
         An array of floats representing a probability mass function.
-    X : np.array or [np.array]
+    X : np.array
         Discrete point values for each probability mass.
-        May be multivariate (list of arrays).
+        For multivariate distributions, the last dimension of X must index
+        "nature" or the random realization. For instance, if X.shape == (2,6,4),
+        the random variable has 4 possible realizations and each of them has shape (2,6).
```
sbenthall, line 826: I believe it would be more consistent with the other downstream code if the 'nature' dimension was the first dimension, not the last.

HARK/distribution.py
sbenthall: Looks like `np.apply_over_axes` is numpy's function to apply over more than one dimension

https://numpy.org/doc/stable/reference/generated/numpy.apply_over_axes.html#numpy.apply_over_axes

Would that fit?

sbenthall:
I've checked it over. See inline comments. Summary:

- `np.apply_over_axes` might be what you're looking for to map a function over multiple dimensions of an array
- Wouldn't it require less alteration to the code if the 'nature' dimension was the first, not last, dimension?

HARK/ConsumptionSaving/ConsRiskyContribModel.py
Mv77: Possible yes, but I still thought this would be the path of least resistance.

I think other parts of HARK that deal with multivariate distributions use things like `dstn.X[n]` to find all the possible draws of the nth dimension. This seems like an intuitive and compact code pattern that I would like to keep. It was easier for me to conceptually think that nature was always the last dimension.

HARK/distribution.py
Mv77: Yes, exactly.

HARK/distribution.py
Mv77: Thanks!

That certainly looks like something I could use. The only issue I see is that it does not accept extra arguments, like `np.apply_along_axis`. I could adapt it with a lambda function, not sure if it will make it lose too much performance.

I'll run some tests and report on Wed. Thanks!

HARK/ConsumptionSaving/ConsRiskyContribModel.py
sbenthall: "X" is truly a terrible name for an object property that is doing so much work. See #1051 

I'm not suggesting you replace "X" with something else in this PR, but I wonder what mathematical object X is now.

It's possible that we could have multiple ways of accessing this data within a distribution, so it requires less manipulation before entering and after pulling it from the object.

PR #1149, opened 2022-06-03
[WIP] Fix "extra dimension" in distribution.calc_expectations

Mv77:
Fixes https://github.com/econ-ark/HARK/issues/1098

### DIFF
--- a/HARK/ConsumptionSaving/ConsPortfolioModel.py
+++ b/HARK/ConsumptionSaving/ConsPortfolioModel.py
@@ -658,9 +658,7 @@ class ConsPortfolioSolver(MetricObject):
         dvdb_intermed = calc_expectation(
             self.IncShkDstn, dvdb_dist, self.bNrmNext, self.ShareNext
         )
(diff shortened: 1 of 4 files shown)


Mv77:
I've fixed the function for now. I'm working on updating the many-many calls that took this issue as given.

Mv77:
This is ready to merge. (Tests will pass, I just updated the changelog).

The one thing that becomes more ugly as a result of this PR is that `calc_expectation(X)` where `X` is a univariate distribution will return a size-1 array, not a scalar. This is the cost of making everything based around arrays and not having if-else statements figuring out the types of multiple possible inputs. It is a cost I'd gladly pay. That is why you will see, in places like `check_conditions` things like `calc_expectations(PermShkDstn)[0]`.

alanlujan91 (approved):
Looks great!

Mv77:
@llorracc @sbenthall I'm merging this in given that it is minor and passed tests and Alan's review.

Thanks @alanlujan91 !

PR #1156, opened 2022-07-23
Faster Expectations and new DiscreteDistributionXRA class 

alanlujan91:
This is a work in progress but I would like some feedback. This PR tackles a few related concepts. 

1. Create new `DiscreteDistribution.calc_expectation(func, args)` method that uses `numpy` functional programming instead of looping. Significantly faster than older method. 
2. Ditto for `DiscreteDistribution.distr_of_function(func, args)`.
3. Creates operator `Expected(func, dist, args)` which uses new `DD.calc_exp` in the back end. 
4. Creates `DiscreteDistributionXRA` which uses underlying `xarray.DataArray` to allow for labeling distributions, dimensions, and coordinates. 
5. `DiscreteDistributionXRA.calc_expectations(func, args, labels=True)` can use labeled and expressive equations that are easier to read as math. 
6. Adds notebooks to demonstrate new functionality. 
7. [Expectations notebook](https://github.com/alanlujan91/HARK/blob/cal_exp/examples/Distributions/Expectation.ipynb)
8. [DiscreteDistributionXRA notebook](https://github.com/alanlujan91/HARK/blob/cal_exp/examples/Distributions/DiscreteDistributionXRA.ipynb)
9. Uses `Expected` operator in `ConsIndShockModel` as proof of concept. In my tests this cuts down runtime by 2/3 (607 ms to 203 ms)

I'm still missing some documentation and tests, but would appreciate feedback and wanted to see if this passes remote tests. 


alanlujan91:
Added `xarray` to requirements.txt but all the errors are related to missing `xarray`. Do i have to change this elsewhere? `setup.cfg`?

HARK/ConsumptionSaving/ConsMarkovModel.py
```diff
@@ -321,29 +321,26 @@ def calc_EndOfPrdvPP(self):
...
             self.DiscFacEff
             * self.Rfree
             * self.Rfree
             * self.PermGroFac ** (-self.CRRA - 1.0)
-            * calc_expectation(
-                self.IncShkDstn,
-                vpp_next,
-                self.aNrmNow
-            )
+            * calc_expectation(self.IncShkDstn, vpp_next, self.aNrmNow, self.Rfree)
```
sbenthall, line 335: This looks like it's using the old calc_expectation construction. A reason for that?

HARK/distribution.py
```diff
@@ -1404,3 +1579,11 @@ def sample(s):
         array_sample = np.frompyfunc(sample, 1, 1)
 
         return array_sample(state)
+
+
+def Expected(func=None, dist=None, args=(), labels=False):
```
sbenthall, line 1584: This is a function with a capitalized name.
That breaks the PyPI naming standard.

Could it be `expected`?

It could potentially be calleable Python class but that would be doing a lot of gymnastics for an idiosyncratic stylizistic win.

HARK/distribution.py
```diff
@@ -930,29 +928,204 @@ def draw(self, N, X=None, exact_match=False):
...
+    def attrs(self):
+        return self._xarray.attrs
+
+    def calc_expectation(self, func=None, *args, labels=False):
+        def func_wrapper(x, *args):
+
+            dim_0 = self.dims[0]
+            idx = self.coords[dim_0].values
+
+            wraped = dict(zip(idx, x))
```
sbenthall, line 1100: wrapped

HARK/distribution.py
```diff
@@ -930,29 +928,204 @@ def draw(self, N, X=None, exact_match=False):
...
+
+    @property
+    def xarray(self):
+        """
+        Returns the underlying xarray.DataArray object.
+        """
+        return self._xarray
+
+    @property
+    def X(self):
```
sbenthall, line 1050: Reminds me of https://github.com/econ-ark/HARK/issues/1051

sbenthall:
Great looking work.
Is the old `calc_expectations` standalone function still in the library?
Is there any reason why it can't be updated to the new, faster numpy code, as you've written for the Distribution method versions of this?
Or do you anticipate converting the whole library over to the method version?

Some minor comments in-line.

HARK/ConsumptionSaving/ConsMarkovModel.py
alanlujan91: I only set up the new method for ConsIndShockModel, this was just some downstream clean up I had to do. We can discuss whether we should implement this across all of HARK at the next meeting.

HARK/distribution.py
alanlujan91: We would *really* like to have the Expectation operator capitalized, as in written math. `sympy` does create a class for it, but you are right that this is a lot for what is just a function in our case

https://github.com/sympy/sympy/blob/77f1d79c705da5e9b3dee456a14b1b4e92dd620c/sympy/stats/symbolic_probability.py#L134-L323

HARK/distribution.py
alanlujan91: Happy to move on from `X`. `xarray`s already use the values and data attribute to return the underlying rvs so this would be easy to do. This was just a patch but we can deprecate the use of `X`

alanlujan91:
> Great looking work. Is the old `calc_expectations` standalone function still in the library? Is there any reason why it can't be updated to the new, faster numpy code, as you've written for the Distribution method versions of this? Or do you anticipate converting the whole library over to the method version?

Yes, the previous `calc_expectations` method is still on the library, which I use to demonstrate the timings in the example notebooks. I would be up for converting the whole library to the new method pending a consensus that the new method is both faster and can be used in the same way that the previous method. Having implemented `ConsIndShockModel` with this new method I would argue that both are true.

HARK/distribution.py
```diff
@@ -930,29 +928,204 @@ def draw(self, N, X=None, exact_match=False):
...
+
+    @property
+    def name(self):
+        return self._xarray.name
+
+    @property
+    def attrs(self):
+        return self._xarray.attrs
+
+    def calc_expectation(self, func=None, *args, labels=False):
```
alanlujan91, line 1094: @sbenthall does this look good to you? Can it be more robust? I still get confused with how python interprets inputs when ordered instead of by key. 

Additionally, wondering if there is a cheap way to remove the boolean `labels`. I found `inspect.getsource(obj)` but I'm wondering if that'd be too much overhead. Haven't tried it.

HARK/distribution.py
alanlujan91: @sbenthall Here I went with args as a tuple because I couldn't figure out how to keep args and label separate.

HARK/distribution.py
llorracc: This looks great!  My only comment is on the ".X" point.  I have a strong preference for giving things names where there's a reasonable chance that one could do a global search-and-replace across the entire codebase and not mess everything up.  Replacing all occurences of the letter X would not satisfy that aim.  Elsewhere I think we have used "vals" and "prbs" for the corresponding objects.  Or, it could be "xvals" and "xprob" (I also like corresponding objects to have the same number of letters).  Maybe do a search through the codebase (and NARK) and see is what is most consistent with prior practice.

llorracc: Or, it could be ".mass" and ".prob"

alanlujan91: Will look into changing `.X` in a different PR.

HARK/distribution.py
```diff
@@ -1404,3 +1664,48 @@ def sample(s):
         array_sample = np.frompyfunc(sample, 1, 1)
 
         return array_sample(state)
+
+
+def ExpectedValue(func=None, dist=None, args=(), labels=False):
```
sbenthall, line 1669: I am -1 on capitalizing this function.
I know that's a minority opinion, and I expect to be overruled.
But I want to note that I think this is short-sighted.
While it may make it marginally easier to _write_ the code, for those who are not accustomed to Python conventions, it will make it more difficult to _read_ the code.
When the project is successful, the code will be read much more often than it is written.
And this choice will ultimately be costly and determined to be unwise.

sbenthall: At the very least, if this remains in CamelCase, then I think the departure from PEP8 and its rationale should be in the documentation.
That way when people look it up in the autodocs there's something to guide them in their confusion about why it isn't a class.

sbenthall: I see you are also still using a tuple for args here.
This is quite confusing given the way 'expected_value' methods on Distribution objects are designed differently.
I expect (no pun intended) you would prefer a parallel construction.

What happens if you use `*args` instead? You mentioned a complication with this on an earlier version of the code.

HARK/distribution.py
```diff
@@ -930,29 +929,288 @@ def draw(self, N, X=None, exact_match=False):
...
+        return self._xarray.name
+
+    @property
+    def attrs(self):
+        """
+        The distribution's attributes.
+        """
+        return self._xarray.attrs
+
+    def expected_value(self, func=None, *args, labels=False):
```
sbenthall, line 1144: The relevant docs on argument ordering are here:
https://docs.python.org/3/tutorial/controlflow.html?highlight=arguments%20style%20positional%20default#special-parameters

HARK/distribution.py
```diff
@@ -930,29 +929,288 @@ def draw(self, N, X=None, exact_match=False):
...
+        def func_wrapper(x, *args):
+            """
+            Wrapper function for `func` that handles labeled indexing.
+            """
+            dim_0 = self.dims[0]
+            idx = self.coords[dim_0].values
+
+            wrapped = dict(zip(idx, x))
+
+            return func(wrapped, *args)
```
sbenthall, line 1187: I would like to talk to you about what you are doing with the 'labels' case here before merging.
You asked if there is a simpler way to do what you're trying to do, and I suspect there is.
But the documentation doesn't make it clear to me what's happening.

HARK/distribution.py
```diff
@@ -930,29 +929,288 @@ def draw(self, N, X=None, exact_match=False):
...
+        Discrete point values for each probability mass.
+        For multivariate distributions, the last dimension of X must index
+        "nature" or the random realization. For instance, if X.shape == (2,6,4),
+        the random variable has 4 possible realizations and each of them has shape (2,6).
+    seed : int
+        Seed for random number generator.
+    coords : dict
+        Coordinate values/names for each dimension of the underlying array.
+    dims : tuple or list
+        Dimension names for each dimension of the underlying array.
```
sbenthall, line 1044: I'm sorry to be getting to a thorough review of this work so late.

What I am noticing here is that the arguments to this DiscreteDistributionXRA class are exposing the configuration of the underlying DataArray data structure: `dims`, `coords`. But the documentation is not providing guidance as to the semantics of these arguments with respect to the class's purpose as _a representation of a probability distribution._

Since you seem to be using these values mainly in the `labels` version of the `expected_value` operation, it leaves me with a lot of questions about the design here. I wonder if there's a clearer way to do what you're trying to do, or a way to describe this capability that doesn't rely so much on a deep understanding of the DataArray format.

It looks like the usage pattern you are intending here is for a multivariate distribution over A and B, you will have 'dims = ['rv', 'x']' where 'x' refers to... the probability mass? And then `coords= {'rv' : ['A','B']}`. The DataArray then stores a values indexed by ((a, b), x) which represent the probability of (A = a, B = b)... but then what is the 'x' dimension?

And if this is the intended usage pattern, would you ever want `dims` to be anything _other_ than `['rv', 'x']`? Would that be a reasonable default?

But I'm actually now quite confused about the data model. Why wouldn't you just have the `dims = ['A', 'B']`? That way each index of (a,b) would map to a probability value. Am I missing something?

HARK/distribution.py
```diff
@@ -930,29 +929,288 @@ def draw(self, N, X=None, exact_match=False):
...
+            The expectation of the function at the queried values.
+            Scalar if only one value.
+        """
+
+        def func_wrapper(x, *args):
+            """
+            Wrapper function for `func` that handles labeled indexing.
+            """
+            dim_0 = self.dims[0]
+            idx = self.coords[dim_0].values
```
sbenthall, line 1183: This code assumes a particular structure of `dims` and `coords`.
That structure should be guaranteed by the constructor of the class, but currently it is wide open to user configuration.
It's easy to make a 'valid' (in terms of what documented) which breaks at this point in the code.

sbenthall:
Looking this over... it looks like currently in the DiscreteDistribution (and the new DiscreteDistributionXRA by extension), the class is initialized with a `pmf` (should really be `pmv`) and a `X/data` argument which for an N-dimensional multivariate is a 2 dimensional array, where each slice `data[:, x]` is a particular point with probability mass `pmf[x]` and values for all N variables.

This takes len(X) + len(X) * N space, where len(X) is determined by the approximation value. In our case, N-VariateNormal().approx(D) expands into D^N different points. So: (1 + N) D ^ N.

I would think that one of the benefits of using xarrays would be to use the dimension labels in a more concise way. I.e., storing all the data of the distribution in a single N-dimensional array, where the value of `data[a,b]` is the point mass of realizing (a,b).

This takes D ^ N space, which is less.

The only downside I can think of is if when approximating a distribution there are some methods that would lead to a very uneven discretization in which many (a,b, ...) values were null. But that could be solved with xarray sparsity, which seems to be supported.

I was wondering if you'd thought about this while working on this @alanlujan91 . It's quite possible I'm forgetting some important rationale for why things are the way they are, but it's probably more likely that I was just trying to extract the functionality from the older HARK code without breaking things, and a new design would be better.

sbenthall:
It's interesting to me that the `draw()` method of the new XRA based class has not been updated -- so the Xarray doesn't seem to be used for operations per se, but rather as a way of storing information about how the multivariate probability data is structured?

sbenthall:
About the 'labels' way of calling a function...

The current syntax for this call is (using the alias):

```
ExpectedValue(
    lambda x: 1 / x["perm_shk"] + x["tran_shk"],
    dist=x_dist,
    labels=True,
)
```

Especially since we want to move away from this rather odd construct of `x` standing for the state of nature (don't we?), I would expect a syntax more like this:

```
ExpectedValue(
    lambda perm_shk, tran_shk: 1 / perm_shk + tran_shk,
    dist=x_dist,
    labels=True,
)
```

... and likewise for when this is called via `dist.expected_value`

llorracc:
@sbenthall,

@alanlujan91 and I talked about this at some length today.  You're right that it is closely connected with the issues around the whole steps/stages/frames direction.  

Also closely connected to some of the things I worked on last year in my pre-2.0-alpha.  

A key goal there was to have a structure from which it would be possible to extract the key equations of a model in a predictable and mathematical-looking way, and where the same exact equations could be applied in the solution and simulation stages.

In the end the only way I found to do that was to encapsulate the relevant python code in a string.  Then the strings associated with transitions could be printed, and the unique string could be evaluated as a python expression both in the solution and the simulation stage.  

If we have a call on Fri, maybe the two of us could briefly discuss what Alan and I discussed today.  

The upshot, though, was a conclusion that maybe we should hold off on merging the xarray and expectation stuff until we are more confident about how we want to handle these things.

sbenthall:
Ok -- it makes sense to hold off on merging the xarray and 'local' function evaluation parts of this PR.

The performance improvements to `calc_expectations` and some of the redesign to the interface of DiscreteDistribution (see #1159 ) could be split off into their own PRs and be most welcome as standalone changes.

There are well known security issues with evaluating strings as Python code. @MridulS has commented on this before.
Let's talk some more soon about this.

alanlujan91:
@Mv77 Not sure why this is causing errors that weren't showing up before, I am guessing they have to do with `interpolation.py`'s new release? Do you have any guess?

Mv77:
Yuup.

The release broke my interpolators. #1157 fixes them but let me push a quick standalone fix.

Incoming!

Mv77:
@alanlujan91 

https://github.com/econ-ark/HARK/pull/1161 is a standalone fix. Its 3 lines of code. If it passes tests you can go ahead and merge it, which should fix the issue.

PR #1157, opened 2022-07-31
Gradients/derivatives in econforge interpolators

Mv77:
This PR adds the capability to compute first-order derivatives of n-dimensional interpolators for my wrapper of `econforge.interpolation.py`.

It works but:
- [x] We should only merge after Pablo has released the new version of `econforge.interpolation.py`. The current version one gets through `pip` has a bug. See https://github.com/EconForge/interpolation.py/issues/94. This PR depends on the version of `econforge.interpolation` currently in their master branch.
- [x] There are various improvements I'd like to make.

*************


Mv77:
Improvements I'd like to make:
- [ ] Add the option to evaluate derivatives w.r.t a single dimension (not the full gradient).
- [ ] Add the option to simultaneously evaluate the function and its derivatives with the same function call. (to save time in figuring indices out).

llorracc:
This is great!  

Though, failing some build tests...

Mv77:
Yes!

The reason is that the current version of `econforge/interpolation.py` that HARK pulls has not fixed the bug! (I think we need to ask Pablo to make a release and require the new version when he does).

llorracc:
Q: If you ask for the derivative exactly at a kink point (where the answer is in principle $\infty$), what does it return?

Or, are piecewise linear splines not allowed?

Mv77:
> Q: If you ask for the derivative exactly at a kink point (where the answer is in principle ∞), what does it return?

I just checked, and it calculates derivatives 'backwards' (using the previous point). Consider the following example
 
```
x = np.array([0.,1.,2.,3.,4.])
y = np.array([0.,0.,1.,3.,6.])

interp = LinearFast(y, [x])
grad = interp.gradient(x)
print(grad)
```
It creates a piecewise linear function. The derivative at 1 is (0 from the left) or (1 from the right). The derivative at 2 is (1 from the left) or (2 from the right). And so on. Here is what I get:

```
grad = array([0., 0., 1., 2., 3.])
```
So it's using left-differences.

--------------------------------

> Or, are piecewise linear splines not allowed?

Yes they are, and they are in fact the only thing I am using at the moment. I have not explored higher order splines.

llorracc:
Hmmm, I guess that's better than $\infty$ or $-\infty$ for preventing the computer blowing up.  But maybe it SHOULD blow up if you ask for a derivative at a point where it is undefined ...

Mv77:
> Hmmm, I guess that's better than ∞ or −∞ for preventing the computer blowing up. But maybe it SHOULD blow up if you ask for a derivative at a point where it is undefined ...

But we would not want to go around blowing up our---or other people's---computers!

Mv77:
@MridulS , @sbenthall , this PR relies on the latest release of Pablo's [`interpolation.py`](https://github.com/EconForge/interpolation.py), which is `[2.2.2](https://pypi.org/project/interpolation/#history)`. There was a bug in previous versions, so if people get the wrong version they will be getting incorrect derivatives.

How do I reflect that in HARK's requirements? Is it a matter only of writing `interpolation >= 2.2.2` in `requirements.txt` or should I do something else?

MridulS:
> How do I reflect that in HARK's requirements? Is it a matter only of writing interpolation >= 2.2.2 in requirements.txt or should I do something else?

Yeah, that will make sure that HARK installed via github is depending on interpolation >= 2.2.2. But we need to make a new release to make sure if someone does `pip install ...` they get the proper version of interpolation.

Mv77:
> > How do I reflect that in HARK's requirements? Is it a matter only of writing interpolation >= 2.2.2 in requirements.txt or should I do something else?
> 
> Yeah, that will make sure that HARK installed via github is depending on interpolation >= 2.2.2. But we need to make a new release to make sure if someone does `pip install ...` they get the proper version of interpolation.

Well the good thing is that none of the content currently in HARK depends on the buggy derivatives, those are only used by elements introduced in this PR. So even without a HARK release people won't be getting bugs from HARK. But when we do roll out these new features we have to enforce `interpolation>=2.2.2`

Mv77:
Bleh. Current tests are failing in windows because `interpolation.py` requieres a package "`curses`" that is [included in base Python for other OSs but not for Windows](https://docs.python.org/3/howto/curses.html).

How on earth do I handle that requirement...

Mv77:
After Pablo's patch this is now ready to review.

Mv77:
@llorracc @sbenthall @alanlujan91, could anyone review this? It's ready to go

HARK/tests/test_econforgeinterp.py
```diff
@@ -1,3 +1,4 @@
+from operator import index
```
alanlujan91, line 1: I don't think this is being used anywhere; and the next line

HARK/tests/test_econforgeinterp.py
```diff
@@ -137,3 +138,125 @@ def test_metric(self):
...
+
+
+class Check1DDerivatives(unittest.TestCase):
+    """ 
+    Checks derivatives in a 1D interpolator
+    """
+
+    def setUp(self):
+
+        pass
```
alanlujan91, line 153: is this necessary?

HARK/tests/test_econforgeinterp.py
```diff
@@ -137,3 +138,125 @@ def test_metric(self):
...
+        y = y0 + slope * x
+
+        interp = LinearFast(y, [x])
+
+        eval_points = np.array([1.5, 2.5, 3.5, 4.5])
+
+        grad = interp.gradient(eval_points)[0]
+        result = np.ones_like(eval_points) * slope
+
+        self.assertTrue(np.allclose(grad, result))
```
alanlujan91, line 170: It doesn't seem like the `derivative` method of `LinearInterp` is tested in the codebase. It might be useful to compare it to the `grad` output here to see if it's correct. Same for Bilinear, etc.

alanlujan91 (commented):
Great PR, left a few comments. Looking forward to implementing this on other models.

HARK/tests/test_econforgeinterp.py
Mv77: You are right, thanks. VScode has gotten into the habit of adding strange, unused imports. Not sure why.

HARK/tests/test_econforgeinterp.py
Mv77: Nope! Thanks!

HARK/tests/test_econforgeinterp.py
Mv77: Good idea, I've just added code to compare with the derivatives of `LinearInterp` and `BilinearInterp`

Mv77:
@alanlujan91 thanks for the great suggestions. The last couple of commits address them all. Don't merge this in yet, I think Chris wanted to chat about it first.

alanlujan91 (approved):

Mv77:
@alanlujan91 approves and I've discussed @llorracc 's questions with him over email, so I'll go ahead and merge. Thanks everyone!

PR #1160, opened 2022-08-04
Fast expectations

alanlujan91:
This PR contains the performance enhancements to `expected` and `dist_of_func` in #1156. 

Also applies `expected` to `ConsIndShockModel`. 


llorracc:
I take it the functionality of this new tool is strictly subsumes what could be done with calc_expectations?  Should we deprecate calc_expectations?

Mv77:
> I take it the functionality of this new tool is strictly subsumes what could be done with calc_expectations? Should we deprecate calc_expectations?

@llorracc the difference is that this new version applies only to operations that can be vectorized. `calc_expectations` is more general (but slower because, since it does not assume that the function that the user provided can be applied to the whole array at once, it has to iterate through shock realizations).

This is amazing work @alanlujan91. It makes me feel validated in the decision to make `DiscreteDistribution.X` a consistent type (`np.arrays` instead of lists of arrays _sometimes_ or arrays _other times_). Yes its cumbersome to deal with the odd length-1 arrays that now happen here and there. But these improvements more than make up for it I think!

PR #1159, opened 2022-08-03
WIP Renames X and pmf

alanlujan91:
See #1156 and #1051 

To be merged after #1156

- [ ] Tests for new functionality/models or Tests to reproduce the bug-fix in code.
- [ ] Updated documentation of features that add new functionality.
- [ ] Update CHANGELOG.md with major/minor changes.

PR #1164, opened 2022-08-16
Rename pmf and X

alanlujan91:
See #1051, #1159


sbenthall:
I like it. Changelog update?

alanlujan91:
PR https://github.com/econ-ark/DemARK/pull/187 fixes DemARK's after this PR is merged.

llorracc:
@alanlujan91, @MridulS 

Do we have a way to check whether any REMARKs might be broken?  

In principle, REMARKs should all have version control on them so should not break.  In practice, I doubt that they all do (or do so correctly).

llorracc:
@MridulS,

To broaden my prior point, if I'm not mistaken, at the moment our procedure is to run DemARK checks nightly by running against the master branch of the development HARK repo.  Would it be easy to set things up so that the DemARKs are also tested against any new PR's?  Or, at least, could we somehow automatically remind the poster of a new PR to check DemARKs (and maybe REMARKs)?

PR #1165, opened 2022-08-19
Interpolators with "Decay" extrapolation

Mv77:
This PR introduces enhanced tools to implement functions (represented by interpolators) that asymptotically approach another function for points that fall outside of their grid.

HARK has some tools for this purpose, specifically in `interpolation.LinearInterp`, which has the option to provide a linear limiting function to which extrapolated values should approach. This PR improves upon those tools in various ways.

- It implements a class `DecayInterp` that receives an interpolator and a limiting function and returns an object that behaves as HARK's other interpolators but approaches the limiting function when queried about off-grid values. This achieves two generalizations with respect to `interpolation.LinearInterp`:
    - It works with n-dimensional (not only 1-dimensional) functions. Also, it works with the faster `econforge` interpolators.
    - The limiting function can be any function, not just a linear function.
 
- It has different methods for calculating the rate at which extrapolated values approach the limiting function.

---------------------------------------------------------

- [x] Remove all the debug lines (prints and plots).


Mv77:
Here is a test that shows how `DecayInterp` compares to the current `LinearInterp` options
https://github.com/Mv77/HARK/blob/259fc6c510a1718e036ff40efae0575bc15ac546/HARK/tests/test_econforgeinterp.py#L388-L402

@llorracc I believe I've succeeded in implementing a decay method (I've called it "decay_hark") that works in multiple dimensions but collapses to the current implementation in 1D. This test compares the two and they produce the same output.

What I still have to wrap my head around is:
- Whether what is currently in there achieves what you wanted, that is, preserving the rate at which the interpolator approaches the limit.
- How to reinterpret the above in multiple dimensions in a way that is tractable.

llorracc:
Great -- thanks for this.   Is the comparison to linearInterp in Decay.ipynb functional already?

Mv77:
Not yet, I was using that notebook to just store some math scribbles on what I intended to do. I will add examples and comparisons soon!

Mv77:
> Great -- thanks for this. Is the comparison to linearInterp in Decay.ipynb functional already?

@llorracc I have quickly moved some plots and tests to [the notebook](https://github.com/Mv77/HARK/blob/plumbing/econforgeinterp/examples/Interpolation/DecayInterp.ipynb) and added minimal comments in case you want to check out/play with the interpolators.

The comparison with base hark is in there.

I must go out now, but will continue polishing this when the Calvo fairy permits.

PR #1166, opened 2022-08-23
Update requirements to include `jupyterlab` and `jupytext`

alanlujan91:


alanlujan91:
Should `pytest` and `nbval` also be part of dev environment?

sbenthall:
Yes, I think they should be part of the dev environment.
Do you agree @MridulS ?

alanlujan91:
It's not currently possible to do `conda install --file requirements.txt`. The output is 

```
>conda install --file requirements.txt
Collecting package metadata (current_repodata.json): done
Solving environment: failed with initial frozen solve. Retrying with flexible solve.
Collecting package metadata (repodata.json): done
Solving environment: failed with initial frozen solve. Retrying with flexible solve.

PackagesNotFoundError: The following packages are not available from current channels:

  - sphinx-rtd-theme
```

Alternatives are `pip install -r requirements.txt` or `pip install -e .`. 

If you want to use packages from conda however you have to install them by hand, and ignore the ones that are only pip packages. An alternative to this could be `conda env create -f environment.yml` where the `environment.yml` file looks something like this

```
name: hark-dev
channels:
  - conda-forge
dependencies:
  - python=3.9
  - numpy
  - matplotlib
  - sphinx
  - future
  - joblib
  - dill
  - scipy
  - flake8
  - nbsphinx
  - seaborn
  - recommonmark
  - interpolation>=2.2.3
  - numba
  - quantecon
  - pandas
  - networkx
  - xarray
  - jupyterlab
  - jupytext
  - pip
  - pip:
    - sphinx-rtd-theme
 ```

I'm not sure if all the dependencies listed are available in conda, but the point is that an `environment.yml` file can enumerate the conda packages and the pip-only packages, and create a full conda environment with 1 command.

PR #1170, opened 2022-08-31
Changes to DiscreteDistributionXRA

alanlujan91:
1. Rename to DiscreteDistributionLabeled.
2. Name dimension of probability "nature". 
3. Update methods. 

- [ ] Tests for new functionality/models or Tests to reproduce the bug-fix in code.
- [ ] Updated documentation of features that add new functionality.
- [ ] Update CHANGELOG.md with major/minor changes.

PR #1168, opened 2022-08-26
UtilityFuncCRRA

alanlujan91:
This PR creates an object oriented CRRA utility function to better organize namespaces and add formal structure to models. 

- [x] Tests for new functionality/models or Tests to reproduce the bug-fix in code.
- [x] Updated documentation of features that add new functionality.
- [x] Update CHANGELOG.md with major/minor changes.

PR #1177, opened 2022-10-05
Labeled models

alanlujan91:


sbenthall:
It looks like this PR includes a lot of changes that work in your new syntax for utility functions.

Can these changes be merged without the additional changes that are about labelled models?

It would be easier to review the material in this PR if it was broken up into multiple PRs that focus an particular functions.

examples/LabeledModels/LabeledModels.py
```diff
@@ -0,0 +1,564 @@
...
+# %% [markdown]
+# Because of the curvature of the value and marginal value functions, it'll be useful to use the inverse value and marginal value functions instead and re-curve them. For this, I create a new class `ValueFunctionCRRA` that returns the appropriate value and marginal value functions.
+#
+
+# %%
+class ValueFunctionCRRA(object):
+    def __init__(self, dataset: xr.Dataset, CRRA: float):
+
+        self.dataset = dataset
+        self.CRRA = CRRA
```
sbenthall, line 229: It seems like what you are doing here is making a callable wrapper around an xr.DataSet.

I wonder if this can be implemented in a more general way.

examples/LabeledModels/LabeledModels.py
```diff
@@ -0,0 +1,564 @@
...
+rand_v = vfunc({"mNrm": rand_states})
+np.max(np.abs(rand_v - rand_ds["v"]))
+
+
+# %% [markdown]
+# ### State transitions
+#
+
+# %%
+def state_transition(s=None, a=None, params=None):
```
sbenthall, line 300: Ah, yes. I've found it useful to write transition functions in similar way.

In my code, I've been using `x` for states and `k` for discretely realized values of shocks.

HARK/distribution.py
```diff
@@ -1069,32 +1067,92 @@ def __init__(
...
+
+    @classmethod
+    def from_unlabeled(
+        cls,
+        dist,
+        name="DiscreteDistributionLabeled",
+        attrs=None,
+        var_names=None,
+        var_attrs=None,
+    ):
```
sbenthall, line 1113: Docstring for this please

HARK/distribution.py
```diff
@@ -1069,32 +1067,92 @@ def __init__(
...
+            name=name,
+            attrs=attrs,
+            var_names=var_names,
+            var_attrs=var_attrs,
+        )
+
+        return ldd
+
+    @classmethod
+    def from_dataset(cls, x_obj, pmf):
```
sbenthall, line 1128: Docstring for this please

HARK/ConsumptionSaving/ConsLabeledModel.py
```diff
@@ -0,0 +1,663 @@
...
+            self.solution_next.attrs["m_nrm_min"] - 1
+        ) / self.params.Rfree
+
+        # if natural borrowing constraint is binding constraint, then we can not
+        # evaluate the value function at that point, so we must fill out the data
+        if self.BoroCnstArt is None or self.BoroCnstArt <= self.BoroCnstNat:
+            self.m_nrm_min = self.BoroCnstNat
+            self.nat_boro_cnst = True
+
+            self.borocnst = xr.Dataset(
```
sbenthall, line 224: I'm having trouble understanding why the borrowing constraint is represented as a DataSet.

HARK/ConsumptionSaving/ConsLabeledModel.py
```diff
@@ -0,0 +1,663 @@
...
+        wfunc = self.create_continuation_function()
+
+        # get optimal actions/decisions from egm
+        acted = self.egm_transition(self.post_state, wfunc, self.params)
+        # get state from actions and post_state
+        state = self.reverse_transition(self.post_state, acted, self.params)
+
+        # egm requires swap dimensions; make actions and state functions of state
+        action = xr.Dataset(acted).swap_dims({"aNrm": "mNrm"})
+        state = xr.Dataset(state).swap_dims({"aNrm": "mNrm"})
```
sbenthall, line 376: You seem to be imputing some semantics on the other order of the dimensions in these DataSets.

What is that semantics?

Documentation/CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ Release Date: TBD
 * Adds keyword argument `labels` to `expected()` when using `DiscreteDistributionXRA` to allow for expressive functions that use labeled xarrays. [#1156](https://github.com/econ-ark/HARK/pull/1156)
 * Adds a wrapper for [`interpolation.py`](https://github.com/EconForge/interpolation.py) for fast multilinear interpolation. [#1151](https://github.com/econ-ark/HARK/pull/1151)
 * Adds support for the calculation of dreivatives in the `interpolation.py` wrappers. [#1157](https://github.com/econ-ark/HARK/pull/1157)
+* Creates `UtilityFuncCRRA` which is an object oriented utility function with a coefficient of constant relative risk aversion and includes derivatives and inverses. [#1168](https://github.com/econ-ark/HARK/pull/1168)
```
sbenthall, line 31: We should merge in this PR #1168 first and then revisit this material around Labeled Models.

alanlujan91:
> It looks like this PR includes a lot of changes that work in your new syntax for utility functions.
> 
> Can these changes be merged without the additional changes that are about labelled models?
> 
> It would be easier to review the material in this PR if it was broken up into multiple PRs that focus an particular functions.

Yes, there are changes on this branch that are part of other PRs, but I should probably remove those. 

The most important files for this PR are `HARK/ConsumptionSaving/ConsLabeledModel.py` and the corresponding notebook `examples/LabeledModels/LabeledModels.ipynb`.

examples/LabeledModels/LabeledModels.py
```diff
@@ -0,0 +1,564 @@
...
+#
+
+# %%
+mNrm = DataArray(
+    mVec,
+    name="mNrm",
+    dims=("mNrm"),
+    attrs={"long_name": "Normalized Market Resources"},
+)
+state = Dataset({"mNrm": mNrm})  # only one state var in this model
```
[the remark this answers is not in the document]
alanlujan91: It seemed appropriate to me that the state variable is a collection of DataArray's each with it's own attributes, names, etc, which would correspond to a Dataset. In this way, the state variable can be generalized to be an n-dimensional Dataset, each dimension being a DataArray.

examples/LabeledModels/LabeledModels.py
alanlujan91: There might be a more general way. This is to use the trick of using the inverse value function to approximate the value function.

examples/LabeledModels/LabeledModels.py
alanlujan91: Here, s=state, a=action(s). Elsewhere I might have used `shocks` explicitly, so I should probably be more consistent with this nomenclature.

HARK/ConsumptionSaving/ConsLabeledModel.py
alanlujan91: This was a suggestion by @llorracc. An action is a consequence of the state, but given the end of period state, the endogenous grid method tells you how the agent must have acted in order to be optimal. This should be made more clear in documentation.

alanlujan91:
I have removed the changes in #1168 but this still depends on the changes in #1170

HARK/ConsumptionSaving/ConsLabeledModel.py
```diff
@@ -0,0 +1,663 @@
...
+
+        if isinstance(state, xr.Dataset):
+            state_dict = state.coords
+        else:
+            state_dict = {}
+            for coords in self.dataset.coords.keys():
+                state_dict[coords] = state[coords]
+
+        result = CRRAutility(
+            self.dataset["v_inv"].interp(
```
sbenthall, line 39: I am confused about what mathematical object "v_inv" refers to.

It does not appear to be the mathematical inverse of the value function, which would be my first guess.

HARK/ConsumptionSaving/ConsLabeledModel.py
```diff
@@ -0,0 +1,663 @@
...
+
+        return result
+
+
+class ConsumerSolutionLabeled(MetricObject):
+    def __init__(
+        self,
+        value: ValueFuncCRRALabeled,
+        policy: xr.Dataset,
+        continuation: ValueFuncCRRALabeled,
```
sbenthall, line 104: To the extent that `value`, `policy`, and `continuation` all are wrappers around an underlying Dataset with shared coordinates, maybe this can be designed more flexibly.

HARK/ConsumptionSaving/ConsLabeledModel.py
```diff
@@ -0,0 +1,663 @@
...
+        v_inv.name = "v_inv"
+        v_inv.attrs = {"long_name": "inverse value function"}
+
+        v_der_inv = cNrm.copy()
+        v_der_inv.name = "v_der_inv"
+        v_der_inv.attrs = {"long_name": "inverse marginal value function"}
+
+        dataset = xr.Dataset(
+            {
+                "cNrm": cNrm,
```
sbenthall, line 177: I think it's an interesting choice that this is being labeled with `cNrm`, which names a variable, when it refers to a policy or decision rule that chooses the value for this variable based on preceding state variables.

I believe in HARK this was called something like `cFunc` in order to denote that it's a decision rule.

I wonder if there's a good reason to pick the more concise notation in this version of the model

HARK/ConsumptionSaving/ConsLabeledModel.py
```diff
@@ -0,0 +1,663 @@
...
+
+        vfunc = ValueFuncCRRALabeled(
+            dataset[["v", "v_der", "v_inv", "v_der_inv"]], self.CRRA
+        )
+
+        self.solution_terminal = ConsumerSolutionLabeled(
+            value=vfunc,
+            policy=dataset[["cNrm"]],
+            continuation=None,
+            attrs={"m_nrm_min": 0.0},
```
sbenthall, line 193: whereas `value`, `policy`, and `continuation` are all objects at the level of abstraction of the Bellman problem, `attrs` is as the level of abstraction of the underlying xarray-powered data structure.

That's somewhat mismatched, and I wonder about the meaning of the `"m_nrm_min" field here. Is this representing a constraint on the policy, or a description of the state grid?

(Also, we should probably use _either_ `cNrm` or `c_nrm` but not switch between them?)

HARK/ConsumptionSaving/ConsLabeledModel.py
sbenthall: Just noting that you explained this to me and it made sense!

HARK/ConsumptionSaving/ConsLabeledModel.py
```diff
@@ -0,0 +1,663 @@
...
+    def post_state_transition(self, ps=None, params=None):
+        """
+        post_state to next_state transition
+        """
+        ns = {}  # pytree
+        ns["mNrm"] = ps["aNrm"] * params.Rfree / params.PermGroFac + 1
+        return ns
+
+    def reverse_transition(self, ps=None, a=None, params=None):
+        """state from post_state and actions"""
```
sbenthall, line 285: 'reverse' or 'inverse' ?

HARK/ConsumptionSaving/ConsLabeledModel.py
sbenthall: This is cleverly implemented!

HARK/ConsumptionSaving/ConsLabeledModel.py
```diff
@@ -0,0 +1,663 @@
...
+    def value_transition(self, action=None, state=None, continuation=None, params=None):
+        """
+        value of action given state and continuation
+        """
+        variables = {}  # pytree
+        ps = self.state_transition(state, action, params)
+        variables.update(ps)
+
+        variables["reward"] = self.u(action["cNrm"])
+        variables["v"] = variables["reward"] + params.Discount * continuation(ps)
```
sbenthall, line 308: It may be my own ignorance, but the question of whether the application of `params.Discount` should be here or in the continuation (and whether other discounting bits like the PermGroFac term should be here or in the continuation) is confusing to me.

alanlujan91:
was able to fix these tests by switching workflow of examples from 3.7-3.8 to 3.8-3.9

@llorracc @MridulS @sbenthall should we still be maintaining compatibility with 3.7? or can we move up?

HARK/ConsumptionSaving/ConsLabeledModel.py
alanlujan91: In the last period, $v(m) = u(c(m)) = u(m)$, so we define the "inverse" as $v^{-1}(m) = u^{-1}(v(m)) \equiv m$. 

In this case $v$ is curved but $v^{-1}$ is linear, so if we want to approximate $v$ off the grid we get better results linearly approximating $v^{-1}$ then recurving, than linearly interpolating $v$, especially at low values of $m$. 

We can do the same at earlier periods, where $v^{-1}_t(m) = u^{-1}(v_t(m))$. In For $t \ne T$, $v^{-1}_t$ is not linear, but it's much less curved than $v_t$ (does not go to neg infty) so a linear approximation approach is still better.

HARK/ConsumptionSaving/ConsLabeledModel.py
alanlujan91: I assumed that was an artifact of differentiating between an array and a callable (function). Since xarray does both, I didn't think i needed to differentiate

HARK/ConsumptionSaving/ConsLabeledModel.py
alanlujan91: I think reverse is more appropriate. State transition is forward (state to post-state, given action), reverse transition is backward (post-state to state, given action).

HARK/ConsumptionSaving/ConsLabeledModel.py
alanlujan91: For this piece, I wanted to consider `Discount` as a coefficient (float) parameter. Stochastic discounts go into the calculation of the continuation function.

sbenthall:
Tests, documentation, and changelog please!

alanlujan91:
Messed up the upstream/master merge and a lot of things broke down. Will have to do a bit of work to fix this one.

alanlujan91:
@llorracc Further changes can be made to these models, but I think they are in good enough shape for release 0.13, pending passing checks

sbenthall:
Hi @alanlujan91 

I know you are stoked to get this feature in before the 0.13 release.
But for a substantial addition to the code base like this, I think we really do need to insist on docs and tests, which are still missing.

We use the numpy standard for docstrings so API documentation is autogenerated into ReadTheDocs:
https://numpydoc.readthedocs.io/en/latest/format.htmls.

For the tests, they can probably be copied and edited from the existing consindshock tests, and used to verify that the new implementation is bug free.

Given that the 0.13 release is already way overdue, I would recommend that we let you give this PR the attention it deserves and hold off on merging it until later. I would like to get on a semiannual release cycle for HARK, which would mean that we would have another release this coming summer.

sbenthall:
Very well. We can hold off on the 0.13 release. But we need tests and docs for this work.

llorracc:
I agree, we need to wait for tests and docs.

sbenthall:
Merged. Hooray! Great work @alanlujan91

PR #1184, opened 2022-10-25
Add seed as property of DiscreteDstnLabeled

Mv77:
I'm starting to use @alanlujan91 's excellent DDL tools.

I noticed a small bug. Some of the discrete distribution features that are extended to DDL expect to be able to access `DDL.seed`. However, the seed is located in `DDL.dataset.seed`; requesting `DDL.seed` returns an error.

This PR fixes the bug by declaring the seed as a property.

Here is an example of code that fails for me without this fix
```
from HARK.distribution import DiscreteDistributionLabeled
import numpy as np

my_dstn = DiscreteDistributionLabeled(
    pmv = np.array([0.5,0.5]),
    data = np.array([-1,1]),
    var_names = ['my_var']
)
my_dstn.dataset

sq_dstn = my_dstn.dist_of_func(lambda x: x['my_var']**2)
sq_dstn.atoms
``` 


alanlujan91 (approved):
Yep! this looks good! glad you found DDL useful!

Issue #1116, opened 2022-02-21
consider explicit representation of Model state/action spaces


sbenthall:
@llorracc has recommended looking into [bellman](https://bellman.dev/docs/latest/index.html), a toolkit for model-based reinforcement learning (MBRL), as inspiration for HARK.

What is `bellman`? It's an implementation of a kind of reinforcement learning algorithm that fits well with OpenAI's [Gym](https://gym.openai.com/) library, a widely used framework for testing RL algorithms.

How might we start using these tools in HARK?

Macro problems are indeed special cases of MDP problems, though the _solutions_ used in HARK are not currently based on RL.

We might ask: does Open AI's Gym contain any tools that can help us _represent problems_, in a way that is agnostic to the kind of solution?

One promising tool might be Gym's implementation of `Space`, which is a superclass for the domain of state and control variables. It includes support for both discrete and continuous space.

https://github.com/openai/gym/tree/master/gym/spaces

llorracc:
@sbenthall is right in guessing that I'm interested in [bellman](https://bellman.dev/docs/latest/index.html) as providing a systematic and well thought out structure for representing and codifying our problems.  It has an explicit structure for each of the elements of a bellman equation (transition equations, states, controls, reward functions, etc), which is something we are sorely lacking.  

The relationship between bellman and Gym is something I have not had time to look into.  I take it that Gym defines a set of particular spaces.  

I'm a bit puzzled by whether Gym allows continuous states and controls; the discussion at the link below suggests that controls, at least, must be discretized:

https://datascience.stackexchange.com/questions/72515/how-to-define-discrete-action-space-with-continuous-values-in-openai-gym

sbenthall:
This is an example of a Gym environment that uses a continuous control variable:

https://gym.openai.com/envs/MountainCarContinuous-v0/

https://github.com/openai/gym/blob/master/gym/envs/classic_control/continuous_mountain_car.py#L99-L101

I have been looking at `bellman` and trying to see if it can do the modeling of bellman equations as is done in macroeconomics. My sense is that the library operates at the MDP level, and that the 'transition equations' used are generally not granular 'frame' transitions, but rather are a trained neural network representing the dynamics of the problem. My sense is that it would be an interesting stretch from the methods/conceptualization of the problem in HARK, but maybe not capable of feature parity with the current methods.

PR #1186, opened 2022-11-08
WIP: BARK: Bellman Abstract Representation toolKit.

sbenthall:
This PR introduces a new module, `stage`, which contains the implementation of a class representing a general Bellman-form stage. This stage can be solved using a generic value iteration algorithm. Stages can be composed to form more complex problems. A notebook demonstrates how the `stage` module can be used to create and solve a Portfolio Consumption problem. The current PR includes some basic tests.

This is the start of a much larger project that addresses some points raised in several issues, such as #1116, #997, #620, among others. It is work in progress. However, the state of this PR is at a point where it would be great to get reviews.


sbenthall:
Looks like my current implementation uses some Python 3.9 features that are available in a different form in 3.8.

https://stackoverflow.com/questions/59955751/abcmeta-object-is-not-subscriptable-when-trying-to-annotate-a-hash-variable

alanlujan91 (commented):
Look at estimagic for alternative optimizers https://github.com/OpenSourceEconomics/estimagic

They also have constraints https://estimagic.readthedocs.io/en/stable/how_to_guides/optimization/how_to_specify_constraints.html

Perhaps the demo notebook should be reversed; start with growth, then allocation, then consumption

sbenthall:
If we use estimagic in one part of the library, we probably should use it throughout.
I just looked it up and found that estimagic is an OpenSourceEconomics project, which is on brand for us!
But the development team seems very small: https://github.com/OpenSourceEconomics/estimagic/graphs/contributors

I'd like to talk more during a meeting about what it would mean to tie HARK to these tools.

sbenthall:
Things seem to be working now.

There is one remaining issue I'd like to discuss, which involves the case where a stage is defined with a pre-given value for pi* for some state value. This is used in the consumption problems for the point where m = 0, so c = 0, but because of a binding constraint. The 'optimal' utility in this case is negative infinity, and this complicates both numerical optimization and interpolation.

The question is: in generality, is it safe to dispense with the role of shocks in these cases? It is straightforward to say, "if (input state) `x` has value `a`, then fix `pi*(a) = b`. But in this Bellman stage framework, the policy function typically takes both input states and realized shocks: `pi : X, K -> A`.

I hope we can discuss this in a meeting.

alanlujan91:
> If we use estimagic in one part of the library, we probably should use it throughout. I just looked it up and found that estimagic is an OpenSourceEconomics project, which is on brand for us! But the development team seems very small: https://github.com/OpenSourceEconomics/estimagic/graphs/contributors
> 
> I'd like to talk more during a meeting about what it would mean to tie HARK to these tools.

I've meet their lead contributor, and they seem to be on their way to being part of NumFocus so I am sure they would be happy to see us use their tools and maybe even future collaboration.

alanlujan91:
> Things seem to be working now.
> 
> There is one remaining issue I'd like to discuss, which involves the case where a stage is defined with a pre-given value for pi* for some state value. This is used in the consumption problems for the point where m = 0, so c = 0, but because of a binding constraint. The 'optimal' utility in this case is negative infinity, and this complicates both numerical optimization and interpolation.
> 
> The question is: in generality, is it safe to dispense with the role of shocks in these cases? It is straightforward to say, "if (input state) `x` has value `a`, then fix `pi*(a) = b`. But in this Bellman stage framework, the policy function typically takes both input states and realized shocks: `pi : X, K -> A`.
> 
> I hope we can discuss this in a meeting.

I think it's safe to dispense with the shocks, as well as anything else in the dynamics. At $m=0, c=0, u(c) = - \infty, v(m) = - \infty, v'(m) = + \infty$, *BUT* $v^{-1}(m) \equiv u^{-1}(v(m)) = 0$ and $v'^{-1}(m) = 0$. This is important not for the $m=0$ case per se, but for the $m = \varepsilon$ small which falls between 0 and the next smallest point $m = \delta$ in our approximation. Interpolation between $- \infty$ and a finite point is undefined, but interpolation in the inverse space is well behaved. 

So for $m =  \delta$ small we use the appropriate dynamics and shocks to evaluate all variables, but for $m= 0$ (or a different borrowing constraint) we need a good anchor for interpolation when an agent is unlucky and happens to fall into low wealth $m = \varepsilon$ between 0 and $\delta$.

sbenthall:
Thanks @alanlujan91 this is all very helpful.

I've been digging a bit deeper into this and have a new take it.

**How it works in this PR**.
It's possible to define a $\pi^*(\vec{x}_p, \cdot) = \vec{a}_p$ in order to set the optimal action at a specific state $\vec{x}_a$.
From this, it is easy to compute $v_x(0) = u(\pi^\*(0)) + \beta E[v_y(m - c)] = -\infty$.
But this only works for the _consumption stage_.

Whatever $\pi^\*(0))$ is in the _allocation_ stage, $v_x(0)$ will just be $v_y(0, \alpha)$.

And during the _growth_ stage, there are no actions, so setting $\pi^\*(0))$ doesn't make much sense.

Does this mean that for generic Bellman stages, it should be possible to add points to the input value function $v_x(0) = -\infty$ directly?

alanlujan91:
> Thanks @alanlujan91 this is all very helpful.
> 
> I've been digging a bit deeper into this and have a new take it.
> 
> **How it works in this PR**. It's possible to define a π∗(x→p,⋅)=a→p in order to set the optimal action at a specific state x→a. From this, it is easy to compute vx(0)=u(π∗(0))+βE[vy(m−c)]=−∞. But this only works for the _consumption stage_.
> 
> Whatever π∗(0)) is in the _allocation_ stage, vx(0) will just be vy(0,α).

Correct. Presumably, $v_y(0, \alpha) = - \infty \  \forall \alpha \in [0, 1]$. This would probably also have to be added manually. 
> 
> And during the _growth_ stage, there are no actions, so setting π∗(0)) doesn't make much sense.
> 
> Does this mean that for generic Bellman stages, it should be possible to add points to the input value function vx(0)=−∞ directly?

I think it would be useful to be able to add points to the input value function. This however could be a generic treatment of the borrowing constraint. Perhaps not true universally, but for a few models if we set `BorroCnstArt = None` then the natural borrowing constraint will generally have this property of tending toward $- \infty$.

sbenthall:
Ok, I am adding in the capability to add arbitrary points to both $\pi^*$ and $v_x$ for any stage.
This will come in the next few commits.

I am now wondering how to handle this for the growth stage, which takes both $a$ and $alpha$ as inputs.
I am tempted to say as you've suggested that  $v_x^{\text{growth}}(0, \alpha) = - \infty \  \forall \alpha \in [0, 1]$.

However... the transition function for the growth stage is of the form $m = f(alpha, R, \eta)a + \theta$.
So while $v_y^{\text{growth}}(m=0) = -\infty$, starting the stage with $a = 0$ ends up with positive $m$ almost 100% of the time.

alanlujan91:
> Ok, I am adding in the capability to add arbitrary points to both π∗ and vx for any stage. This will come in the next few commits.
> 
> I am now wondering how to handle this for the growth stage, which takes both a and alpha as inputs. I am tempted to say as you've suggested that vxgrowth(0,α)=−∞ ∀α∈[0,1].
> 
> However... the transition function for the growth stage is of the form m=(f(α,R,η)a+θ)/(Gψ). So while vygrowth(m=0)=−∞, starting the stage with a=0 ends up with positive m almost 100% of the time.

You are correct. It depends on the exact income process; specifically on $\underline{\theta}$. when we add the possibility of unemployment with no UI, $\underline{\theta} = 0$ so there is some likelihood to end with $m=0$ next period, otherwise as you say the agent will end up with positive $m$. 

This is why the calculation for the natural borrowing constraint is recursive. Assume last period; minimum $\underline{m}_T = 0$. So, in the second to last period $R a + \theta > \underline{m}_T$ which implies $\underline{a} = ( \underline{m}_T - \underline{\theta})/R$. If $\underline{\theta} = 0$ then $\underline{a} = 0$. 

So the treatment at the boundary depends on both the income process and the imposition of a borrowing constraint. In the Portfolio choice model, the artificial borrowing constraint is 0, so it only depends on $\underline{\theta}$.

sbenthall:
OK, thank you.

Since I am trying to build magic solver technology, I think I should be expecting the natural borrowing constraint to show up in the output of the model. But the artificial borrowing constraint must be an input to the model.

I do understand that much of the (different kind of) magic of the SolvingMicroDSOPs methodology is to use analytical model results to improve the computational solution.

What I have not yet figured out is if there is a way to use an analytically derived natural borrowing constraint to improve a magic solver.

sbenthall:
Marking as WIP again as multiple period backwards induction on the Portfolio Choice problem is getting optimization errors, which I expect is due to drift/errors in the value function. I'm working through generalizations of the SolvingMicroDSOPs methodology to see if that irons things out.

The current work is to generalize the use of the marginal value function for the derivation of the consumption function.
The most recent commit implements a solver for the optimal policy function of a general Bellman stage that uses the marginal reward function, marginal output value function, and first order condition.

sbenthall:
`kw_only` argument to dataclasses was introduced in Python 3.10 and may need to be removed from this PR

llorracc:
> When a stage has more than one input variable (such as $a$ and $\alpha$), what is the interpretation of the marginal input value function? Does it need to be computed in a special way?

The "stage" you're describing has two state variables - given the $\alpha$, the value function $v_{\text{after-portfolio-choice}}(a,\alpha)$.  

We are solving the problem on a grid of values for $a$; index them by $i \in \{1, ..., I\}$.  Now suppose that for any specific $a_{i}$ we have a procedure for computing the optimal $\alpha$ (see the SolvingMicroDSOPs code for how this is done).  Call the optimized $\alpha$ function $\check{\alpha}(a)$.  

Effectively, that means there is now a function $v_{\text{with-optimal-portfolio-choice}}(a) = v(a,\alpha^{*}(a))$.

Since we want to assume consumers are always optimizing, this is the function for which we actually need to compute the marginal value:

$$  v_{\text{with-optimal-portfolio-choice}}^{\prime}(a) = \frac{d v(a,\alpha^{*}(a))}{d a}$$

SolvingMicroDSOPs tries to go over all of this carefully, and illustrate it with code.  I'd be grateful for any edits you can suggest for how to clarify or improve the treatment there.

sbenthall:
@llorracc that makes sense when talking about what I'm calling 'the allocation stage' in the work in this PR. The allocation stage is the stage where the agent chooses their allocation $\alpha$.
The 'input' to that stage is $a$, and the input value function to that stage takes into account the optimal allocation choice $\alpha$.

But in the following stage, tentatively called 'the income stage', which looks like this:

* $A = \emptyset$
* $(a, \alpha) \in X = \mathbb{R}^2$
* $m \in Y = \mathbb{R}$
* Shocks:
    * $\psi \sim \text{Lognormal}(0,\sigma_\psi)$
    * $\theta \sim \text{Lognormal}(0,\sigma_\theta)$
    * $\eta \sim \text{Lognormal}(0,\sigma_\eta)$
 * $F(a,\alpha) = 0$
 * $T(a,\alpha) =  \frac{(\alpha \eta + (1 - \alpha) R) a + \theta}{\psi G} $ 
 * $\beta(\psi) = L \psi^{1-\rho}$ where $L$ is chance to remain alive

There is a value function $v_y: Y \rightarrow \mathbb{R}$, and a marginal output value function $v'_y: Y \rightarrow \mathbb{R}$.

With $v_y$, we can solve for the input value function $v_x$ with an optimizer.

With $v'_y$, if the action-value function $q$ is concave and there is a marginal reward function $F'$ (which in this case is trivial), we can (maybe) compute the marginal input value function $v'_x : X \rightarrow \mathbb{R}$.

But in this case, there are two input variables, so $v'_x(a, \alpha)$. So I wonder about the interpretation and computation of this function.

sbenthall:
I would like to revisit this PR, which has been on hold for several months. I thought I would add some context, as well as some ideas for next steps, here, to inform any review. Moving forward, I'm going to the whole approach here, based on Bellman blocks, as BARK, to disambiguate it from the many other HARK redesign proposals floating around. 

This PR is original research supported by [NSF #2105301](https://www.nsf.gov/awardsearch/showAward?AWD_ID=2105301&HistoricalAwards=false) and the Econ-ARK project. The goals of my efforts here have been to:
 - Create a system for defining and solving problems that has feature parity with the scope of @llorracc's [SolvingMicroDSOPs ](https://llorracc.github.io/SolvingMicroDSOPs/) procedure, but not built around a specific problem. Rather, the user can configure the system model definition and solve it using general algorithms.
 - Demonstrate a way to solve a number of outstanding issues in HARK, and resolve many ongoing frustrations with the original design, including https://github.com/econ-ark/HARK/issues/1116, https://github.com/econ-ark/HARK/issues/997, https://github.com/econ-ark/HARK/issues/620,
 - Build a functioning tool that can support future research, including my own research moving forward.

The core ideas represented in this PR are that:
 - dynamic stochastic control problems can be broken up into smaller components that can be solved in sequence. These components are in Bellman form, can be composed like building blocks, and solved using backwards induction.
 - these blocks can be defined as a library of components and reused in different problems, to enable easier comparison between structural models
 - the configuration of the model, the algorithms uses to solve models, and additional information needed by the solver (such as discretization grids) are all logically distinct.

I have been using the term 'stage' for the most basic of these blocks. I now regret this choice. Moving forward, I will take a line out of Pablo's book. There are two structural forms in BARK:
 - *b-blocks*, representing a single Bellman update. I was calling these 'stages' before, but @llorracc seems to want to use `stage` to mean something else he has in mind. So be it.
 - *p-blocks*, which can be composed of *b-blocks* and finish with a *tick*, which is an indication that the model should increment in time. (This is a convention borrowed from NetLogo, where it works quite well.)

Renaming of variables to better fit pre-defined blocks together is accomplished through a simple `twist` operation.

This PR includes an almost complete implementation of BARK. I believe my next step before putting this aside a few months ago was to verify that it can produce results equivalent to HARK for the ConsIndShockModel. 

A few other notes:
 - There has been some notional friction around the variable names used to define a b-block. I'm not wedded to any notation and the friction seems to be due to how b-blocks are defined at an entirely different level of abstraction than the models currently described in HARK.
 - The current implementation supports the use of EGM for each b-block.
 - The current implementation has the limitation that all shocks in b-blocks are IID.
 - I have *not* been trying to optimize the Python code in this version. I have been focusing on the interface and form. Where possible, I've used attractive Python libraries like `xarray`. I know there's a lot to do to improve performance, such as: replacing for-loops with matrix operations; JITing code, and so on. I wanted to wait to work on this until I had automated tests in place for the main functionality.

The main reason I put BARK aside earlier is that @llorracc has been keen on the next HARK version being integrated with Dolang and I was aiming to support that effort. However, in recent conversations, it sounds like @albop is not convinced that Dolang is necessary to support @llorracc 's use case, and proposed building a pure Python API for structural lifecycle models instead. In that case, I propose that BARK might be a solution to @llorracc 's problems. I would be immensely grateful for any feedback from @albop on the BARK design, if only for my own edification. In any case, I don't have any immediate plans to integrate BARK with Dolang; I don't have a need for Dolang, personally -- I'm fine with defining models in software code, as long as it can be done modularly.

I see a few paths forward for BARK and I'm wondering which one is best:
- Prepare BARK as an incremental improvement to HARK, replacing the current `FramedAgentType`, to providing an alternative approach to building and solving models that can be compared with HARK's more hand-crafted models. I think this would take no more than a month of work to wrap up, given its current scope.
- Split off BARK into a separate repository that builds on HARK.

I currently don't see any tools out there that are better than BARK for what I need for my research, and I've recently gotten excited about ways to extend BARK to solve asset pricing problems (this would involve defining new kinds of blocks). I'd be happy to pick up a different tool if somebody pointed me to it.

sbenthall:
I'll add that one reason prefer the 'block' terminology is that it lines up with Bence's work on the[ Sequence Space Jacobian](https://www.bencebardoczy.com/publication/sequence-jacobian/sequence-jacobian.pdf). While the current implementation of BARK uses only a linear sequence of blocks, I could imagine a system where blocks form a DAG for the individual agent's problem. I don't want to add to the scope of this PR at all, which is minimal. I'm just here suggesting where BARK would go, if it evolves where I think it should go.

llorracc:
@sbenthall thanks for this cogent post.

@dedwar65 is right now working on a revision to [SolvingMicroDSOPs](https://llorracc.github.io/SolvingMicroDSOPs), with the particular ambition to remove all references to the antique Mathematica code.  

If BARK is ready to be used to implement all of the elements in the SolvingMicroDSOPs notebook, using it to do that would seem to be a natural next step for demonstrating how BARK can be used in practice.  Having you and Decory work together on that in the near term seems like a good path forward, since he is more familiar with the math and economics and you have more expertise on coding.

sbenthall:
This PR has been superseded by more recent work on 'blocks', which have made it into the repository.

I still draw on this work a great deal in my more recent implementations, but this code won't make it into HARK, so closing the PR.

PR #1189, opened 2022-11-23
Labeled income process

Mv77:
This PR changes the income process constructor in `ConsIndShockModel` to use @alanlujan91's `DiscreteDistributionLabeled` class. This:
- Allows for more expressive code (e.g. `shocks[0]` becomes `shocks['PermShk']`) when taking expectations.
- Might make the code faster, since it uses Alan's vectorized expectation operator.

I have marked it as WIP because I want to extend the change to portfolio models, but tests are currently passing.


HARK/distribution.py
```diff
@@ -1192,6 +1192,10 @@ def RNG(self):
         """
         return self.dataset.RNG
 
+    @RNG.setter
+    def RNG(self,value):
+        self.dataset.attrs['RNG'] = value
```
alanlujan91, line 1197: Should this be `np.random.RandomState(value)`?

Mv77: I am not sure? What does this property do?

I wrote as it is because I thought that the `get` and `set` method should deal with exactly the same object. Since the getter is
https://github.com/econ-ark/HARK/blob/daf2afd9243bf83a9bbd3477c3f14b09b1d52538/HARK/distribution.py#L1188-L1193

I wrote the setter to the RNG property directly. What do you think?

alanlujan91: I think this is how we initialize a random state with a seed. When we initialize a DiscreteDsitribution object, we set RNG as follows 

https://github.com/econ-ark/HARK/blob/daf2afd9243bf83a9bbd3477c3f14b09b1d52538/HARK/distribution.py#L1079

so I guess we would have to do `dist.RNG = np.random.RandomState(seed)`. 

Actually I guess this isn't a problem, you just have to be more explicit that the `value` parameter should be of type `RandomState`. Maybe just a type check?

But also if we have a setter for `dist.seed = value` it should also generate a new RNG.

Mv77: I like the typecheck solution.

You did not add a setter for the seed property (I think?) Should I create one? Would that go in a different PR?

Mv77: Alrighty, added  a typecheck.

I could not remember why I had to add this setter in the first place, but just ran some tests and found out. The reason was that methods like `initialize_sim` call `Distribution.reset()` on things like the income distribution. That method is
https://github.com/econ-ark/HARK/blob/20645e2bae21e3af8bd06d3dcde99ea4b00dc8fe/HARK/distribution.py#L24-L31

So without a setter, it failed

Mv77:
Not yet!

Mv77:
This is ready for review and merge if it passes review.

@llorracc I checked locally that it does not break DemARKs. It's much harder to know for REMARKs, but afaik version control for REMARKs is in the works.

llorracc:
Yes, version control for REMARKs is in the works.

I've asked @sbenthall to work on preparing for a new release (0.13) by the end of this month, and I hope by that time we will have nailed down the status of all the REMARKs: Either pinned to earlier versions of HARK, or added to the tests that are run when new PRs are submitted.  

PS.  I'm a bit puzzled by the combination of the "Needs: Revision" and the "Ready-To-Merge" labels.  Could you clarify what revisions you view as needed?  I guess it's hard to see how to keep track of whether we get those revisions done after the PR is merged.  Maybe the right approach is to make an "issue" containing the revisions that are needed, and link to the issue in this discussion of the merge request?

Mv77:
What I wanted to express was "This is ready from my point of view. It would be good to get someone to look at it. If that someone approves, they can go ahead and merge without further questions."

llorracc:
OK, I guess the right "someone" is @alanlujan91 

Alan, go for it!

sbenthall:
I've removed the "Needs:revision" label.

alanlujan91 (approved):

PR #1195, opened 2022-12-19
Journey 3

AMonninger:

- [x] Adds notebook: Journey_Policymaker.ipynb
- [x] Updates notebook: Journeys_into_HARK.ipynb
- [x] Renames notebook:  Journey_2_Engineering_Background.ipynb such that we can use hyperlinks

To Do:
- [x]  Rename Journey_1_PhD.ipynb to Journey_PhD.ipynb (change in Documentation (symlink, index.rst), files which link to it (examples/Journeys/Journeys_into_HARK.ipynb))


llorracc:
Small comment - nobody should ever give numbers (like "3") to items like this.  The prior two "journeys" were done largely without my intervention and so I did not enforce this; but could you rename this one to something like "Journey-For-Macroeconomists"?  Thanks

When it's done, if you are able to do so, please add a "Ready-To-Merge" tag (if you think it's ready).

llorracc:
Actually, looking in more detail I see that you already are calling it Journey_3_Policymakers.  

1. Don't include underscores in filenames because that can cause problems with referring to content in LaTeX or Markdown.  (Replace with dashes).
2. Remove the "3"
3. Please modify the other "Journeys" to omit numbers from their names and underscores from their filenames, so that you can reference them easily

llorracc:
@AMonninger could you take a look at this and change the name as requested and make whatever other minor adjustments are necessary to resolve the conflicts?  Thx.

llorracc (commented):
Adrian,

Could you replace underscores in the filenames for all Journeys with hyphens?  

Once that's done, they will be ready to merge.



https://www.woorank.com/en/blog/underscores-in-urls-why-are-they-not-recommended

llorracc (commented):
My only other comment is that some of the parameters you are setting are superfluous -- if you are just setting a parameter to the default value, and that parameter is not crucial in understanding a substantive point that the notebook wants to make, then it should not be set.  You only want to set the parameters that either differ from the default, or that you really want to explain the role of in the notebook itself.

PR #1197, opened 2023-01-01
[WIP] `scipy.stats` distributions

alanlujan91:


llorracc:
Only comment is that I'd like it if the invocations of `.discretize` (which replace the `.approx`) were to explicitly specify the method ("equiprobable").

alanlujan91:
> Only comment is that I'd like it if the invocations of `.discretize` (which replace the `.approx`) were to explicitly specify the method ("equiprobable").

Ok, right now the default for every distribution is `equiprobable` except for `Normal` which has a default of `hermite`.

llorracc:
OK, I'm happy with this now.  (Actually, delighted).  

Only two things I can think of:

1. Update the changelog
2. Figure out how to test it against the DemARKs BEFORE we merge into master
    - If all the DemARKs break, we will either need to fix them or to change the code so it works with the old code (maybe issuing a deprecation warning)
    - If only a few DemARKs break, they can be fixed

alanlujan91:
@llorracc the only way I can think of checking this branch against DemARK's is locally. I will try that and see what the results are

alanlujan91:
@llorracc seems like DemARK's already broke with previous merge so I'll have to go and fix them anyway

HARK/ConsumptionSaving/ConsIndShockModel.py
```diff
@@ -934,7 +943,8 @@ def use_points_for_interpolation(self, cNrm, mNrm, interpolator):
 
         # Combine the constrained and unconstrained functions into the true consumption function
         # breakpoint()  # LowerEnvelope should only be used when BoroCnstArt is true
```
sbenthall, line 945: While you're at it, could you remove this commented out breakpoint?

HARK/distribution.py
```diff
@@ -281,58 +305,68 @@ class Lognormal(Distribution):
...
+    def _approx_equiprobable(
+        self, N, endpoints=False, tail_N=0, tail_bound=None, tail_order=np.e
+    ):
         """
         Construct a discrete approximation to a lognormal distribution with underlying
         normal distribution N(mu,sigma).  Makes an equiprobable distribution by
         default, but user can optionally request augmented tails with exponentially
         sized point masses.  This can improve solution accuracy in some models.
 
+        TODO: add endpoints option
```
sbenthall, line 368: Is there an issue for this in the issue tracker?

Are you going to do it in this PR, or leave it to a later release?

HARK/distribution.py
```diff
@@ -411,22 +446,69 @@ def approx(self, N, tail_N=0, tail_bound=None, tail_order=np.e):
...
+
+        limit = {
+            "dist": self,
+            "method": "equiprobable",
+            "N": N,
+            "endpoints": endpoints,
+            "tail_N": tail_N,
+            "tail_bound": tail_bound,
+            "tail_order": tail_order,
+        }
```
sbenthall, line 472: This may be implemented a little more generically using `**kwargs` to capture the arguments given the `discretize` method.

https://www.digitalocean.com/community/tutorials/how-to-use-args-and-kwargs-in-python-3

HARK/distribution.py
```diff
@@ -616,9 +715,9 @@ def approx(self, N, equiprobable=False):
 
         # Now find a discretization for a univariate standard normal.
         if equiprobable:
-            z_approx = Normal().approx_equiprobable(N)
+            z_approx = Normal().discretize(N, method="equiprobable")
```
sbenthall, line 718: I'm a little confused by the structure here...

`_approx` as a private method (?) which takes a boolean argument that corresponds to a string variable to a public method... Seems like we would want `method` instead of equiprobable here and just pass the string through.

HARK/distribution.py
```diff
@@ -627,162 +726,49 @@ def approx(self, N, equiprobable=False):
...
-            )
-        return draws[0] if len(draws) == 1 else draws
-
-    def approx(self, N, endpoint=False):
+    def __init__(
+        self, dist: rv_discrete, *args: Any, seed: int = 0, **kwds: Any
+    ) -> None:
         """
-        Makes a discrete approximation to this uniform distribution.
+        Parametrized discrete distribution from scipy.stats with seed management.
```
sbenthall, line 757: Paramet_e_rized

HARK/distribution.py
```diff
@@ -1019,14 +989,25 @@ def dist_of_func(self, func=lambda x: x, *args):
...
         f_dstn = DiscreteDistribution(list(self.pmv), f_query, seed=self.seed)
 
         return f_dstn
 
+    def discretize(self, N: int, *args: Any, **kwargs: Any) -> "DiscreteDistribution":
+        """
+        `DiscreteDistribution` is already an approximation, so this method
+        returns a copy of the distribution.
+
+        TODO: print warning message?
```
sbenthall, line 1006: Is there an issue for this?

sbenthall:
Looking good! See review comment inline. Please update the CHANGELOG in this PR.

HARK/distribution.py
alanlujan91: I was thinking of leaving it for a different PR since I still need to discuss how to do this with the group. 

I will make an issue for this.

HARK/distribution.py
alanlujan91: Will write an issue for this

HARK/distribution.py
alanlujan91: That makes sense. I'll change this.

HARK/distribution.py
alanlujan91: That's a good point. Will think about this and leave it for a different PR.

alanlujan91:
Changes in this PR:
- Reorganize `Distribution` class as seed manager with abstract `draw` and `discretize` methods. Random number generator `_rng` is now a private attribute. 
- New `discretize` method replaces `approx`, with parameters `N`, `method`, `endpoints`, and additional keywords. The recommended use is to be explicit about the discretization method being used by setting `dist.discretize(N, method = "EXPLICIT")`. The `endpoints` parameter has only been implemented for `Uniform` distribution so far but it should allow user to add points with infinitesimal probability in the Discrete Distribution.
- Create new `ContinuousFrozenDistribution` which allows users to use any `scipy.stats` continuous distribution on HARK even if they have not been implemented in HARK. 
- All HARK distributions now inherit methods from `scipy.stats` such as `rvs`, `pdf`, `cdf`, `sf`, `ppf`, `isf`, `stats`, `moment`.
- New `Lognormal.discretize(N, method = "hermite")` which uses Gauss Hermite method to discretize `Lognormal` and `MeanOneLogNormal`. 
- Create new `DiscreteFrozenDistribution` which allows users to use any `scipy.stats` discrete distribution on HARK even if they have not been implemented in HARK. 
- `DiscreteDistribution` and `DiscreteDistributionLabeled` now take parameter `limit` to indicate the Continuous Distribution it comes from and the method used for discretization. 
- Implement the use of `dist.discretize` with explicit method throughout the codebase.

HARK/distribution.py
alanlujan91: #1204

HARK/distribution.py
alanlujan91: #1205

PR #1203, opened 2023-01-09
[WIP] Interpolation overhaul 

alanlujan91:


alanlujan91:
Requires a big clean up but I would like this to be part of 0.13 as well. Will have to see if I have time before we want to publish new release.

llorracc (commented):
I'd prefer the naming convention to begin with `interp` and end with the source, rather than the other way around.  Also, maybe with an underscore.  So, like, `interp_econforge.py` and `interp_hark` or whatever.

Let's discuss tomorrow.

sbenthall:
If this is modeled on scipy's architecture, where the `interpolation` module imports many smaller 'hidden' modules in its `__init__.py`, it looks like scipy does is drop the intitial `interp` but name these hidden modules with a prefix underscore, which is a python quasi-convention for a hidden variable. See:

https://github.com/scipy/scipy/tree/main/scipy/interpolate

llorracc:
@sbenthall thanks for looking into this.  I like the way scipy does it.  

@alanlujan91, when you return to this topic, please think about whether there are any obstacles to following the scipy approach (or just adopting it outright).

sbenthall:
#1242 is about the EconForge interpolator as an additional backend to multivariate interpolation?

llorracc (commented):
Small comment:  My pref is to be more systematic about naming, and have the general followed by the specific.  Like, it should be `interp_bilinear, interp_decay, interp_linear.`  Also, is `LinearFast` an interpolator?  If so it should be `interp_linear_fast`

PR #1218, opened 2023-01-25
[WIP] Pre-commit

alanlujan91:


MridulS (approved):
LGTM for now, we can always revisit the config.

This PR will probably result in a lot of other merge conflicts in other PRs so before maybe merging that should be taken into account.

llorracc:
@alanlujan91 would merging this make pre-commit the default, or would the result be a situation like the one we discussed earlier in which additional steps would be necessary to activate it?  And, if the latter, would the pre-commit automation be something specific to whatever conda env you were in when you set it up?  So, then, we could each have a conda env in which precommit was activated, and one in which it was not...

alanlujan91:
@llorracc the latter, it would need some steps to activate it
we could have different environments, but the steps to activate it is not to just have some libraries in the environment, you also have to run some commands

alanlujan91:
Superseded by #1225

PR #1225, opened 2023-02-07
Pre-commit pre-0.13

alanlujan91:
Cleaner version of  #1218. 

Should be merged after #1177.


sbenthall:
I'm confused about what's going on in this PR.
Seems like a lot of formatting changes to latex .sty files.

What's the intention here?

alanlujan91:
@sbenthall the intent is to set up more pre-commit tools which will
- [jupytext] sync, clean up, and execute new notebooks in PR's
- [black] format new code in PR
- [pyupgrade] update small python snippets as we drop older versions of python
- [blacken-docs] format documentation
- [isort] sort imports on .py files in PR
- [mirrors-prettier] clean up and format other types of files in codebase
- [pre-commit-hooks] other small clean-up/formatting

After setting up the pre-commits, I also ran `pre-commit run -a` to make sure the codebase is up to the pre-commit standard and not just new changes. 

As we discussed, these hooks should all be modifiers instead of checkers, so they won't just tell you to fix something. 

Also as @llorracc suggested, pre-commit configuration is such that if you know how to use pre-commit and have it installed, you can use this. If you don't, it won't automatically run the pre-commits. This is to avoid new users being turned off by too much overhead.

sbenthall:
Ok, got it.
This is all Very Good.

I wonder if it would be possible to do one more thing in this PR before merging, which is to document this pre-commit option (as we've discussed and you've described it here) in the Contributing to Econ-ARK documentation:

https://hark.readthedocs.io/en/latest/contributing/CONTRIBUTING.html

This will give this work more exposure, so more people are likely to use it!

sbenthall:
Thank you. Merging!

PR #1226, opened 2023-02-08
Requirements

alanlujan91:
Do we need changes to `setup.cfg` as well??


alanlujan91:
Related to https://github.com/econ-ark/DemARK/pull/194

llorracc (approved):
Looks good to me!

sbenthall:
> Do we need changes to `setup.cfg` as well??

Interesting question.

Looking this up:
 - It looks like requirements.txt is mainly used when setting up a _development environment_. I.e., pip install -r requirements.txt is what you do when hacking on the library itself...
 - but when we run `python setup.py sdist` to package the library for PyPI, that doesn't look at `requirements.txt`. Rather, it looks at `setup.py` which points to `setup.cfg`

https://towardsdatascience.com/requirements-vs-setuptools-python-ae3ee66e28af

So, short answer -- adding a minimal amount of version bounds on setup.cfg is a good idea.

The linked article actually recommends tighter pins for `requirements.txt` than we settled on, but this seems to be a matter of community taste.

llorracc:
Ugh.  So now we have 3 different candidates for how we handle the versioning stuff?  requirements.txt, setup.cfg, and environment.yml?  

It would be nice if somehow we could have a "single source of truth" and if the other files are required they could get programmatically generated from the SST.  

Probably should loop in @camriddell on this since he is revamping the REMARKs and their requirements.  And @MridulS

MridulS:
Naaaaaaaa, we don't need to do the conda bit here in HARK itself. We should be doing environment.yml only in REMARKs.

We should be able to get one and only one pyproject.toml file to take care of everything. *All* the configs should go in pyproject.toml

I'll try to get this done before the meeting next week.

llorracc:
@MridulS, is it OK if we make the release without fixing this?  Then, maybe, when fixed, we can make another release?

MridulS:
Yeah no need to wait for this PR. We can always wait for the next release.

llorracc:
OK, then let's go ahead and make the release.

sbenthall:
Ok. But then we should update the requirements list in `setup.cfg`.

Because these changes to `requirements.txt` will not effect the PyPI package.

alanlujan91:
@sbenthall ok just made the change to setup.cfg

sbenthall:
Thanks @alanlujan91 ! I'll merge and do the release today.

PR #1234, opened 2023-03-02
Fast pytest

alanlujan91:


alanlujan91:
attempt to make running checks and tests faster through github actions. Unfortunately the machine we get for testing only seems to have 2 cores at most, which makes the effect of this change limited.

sbenthall:
@MridulS

MridulS (approved):
Looks good @alanlujan91, but it looks like the renovatebot configs have also slipped in this PR. Should they be removed?

alanlujan91:
@MridulS oh yeah, I'll remove those. 

However, I think we should have a discussion about renovatebot or some other dependency manager. Seems pretty useful.

PR #1239, opened 2023-03-09
Update MetricObject

alanlujan91:
Small updates to MetricObject and related functions


llorracc (commented):
What is the gist of this PR?

PS.  What is a TimeVaryingDiscreteDistribution?  One in which there is a transition matrix that connects one period to another?  Or one in which, say, you have a different distribution in the last period of life than in the second-to-last etc?

alanlujan91:
@llorracc this is just a reorganization of MetricObject and related functions, there is no new functionality just thought I would clean up some code. 

A TimeVaryingDiscreteDistribution is kind of a `list` of DiscreteDistributions such that you can call them by index [i].

llorracc:
So, in our Brave New World in which each period has its own environment, we will not need the TimeVaryingDiscreteDistribution?  (Remember that my vision is that you construct your periods backwards by giving each one exactly what it needs).  

Anyway, it seems that this is ready to merge.  Shall I merge it?

sbenthall:
what about breaking the metric object distance functions out into something like `metric.py`

sbenthall:
Maybe overkill implementations for our purposes, but we might want to think about:
 - https://github.com/metric-space-ai/metric
 - https://donsheehy.github.io/metricspaces/

... as ways to do more with 'metric objects'?

alanlujan91:
@sbenthall have moved MetricObject to metric.py

PR #1242, opened 2023-03-17
[WIP] Vector-valued interpolation in `econforgeinterp`

Mv77:
`econforgeinterp` (and all other interpolators in HARK) currently support interpolation of scalar valued functions, $f:\mathbb{R}^n\rightarrow \mathbb{R}$. I would like to work on adding the capability to interpolate vector-valued functions, $f:\mathbb{R}^n\rightarrow \mathbb{R^m}$.

The current version of this PR makes this possible in the barebones `LinearFast`. I still need to adjust the extra-stuff: derivatives and decay-interpolation.

## Why do I think this would be useful?

In a model with, say, a 2-dimensional state space, we often will construct a bunch of interpolators for `vFunc(m,n)`, `dvdmFunc(m,n)`, `dvdnFunc(m,n)`, `cFunc(m,n)`... over the same `(m,n)` grids.

Often, in the expectation step of backward induction, we will need to evaluate, say `dvdmFunc(m,n)`, `dvdnFunc(m,n)` in a bunch of points. We can do this with two different interpolators, one for each function. But this does the slow calculation of which indices and weights to use for interpolation twice for every point (once in each interpolator). If we instead had an
$\mathbb{R}^2\rightarrow \mathbb{R^2}$ interpolator of `(m,n) -> [dvdmFunc(m,n), dvdnFunc(m,n)]` the calculation of indices and weights would be done only once.

### DIFF
--- a/HARK/econforgeinterp.py
+++ b/HARK/econforgeinterp.py
@@ -24,10 +24,12 @@ class LinearFast(MetricObject):
 
     def __init__(self, f_val, grids, extrap_mode="linear"):
         """
-        f_val: numpy.array
-            An array containing the values of the function at the grid points.
-            It's i-th dimension must be of the same lenght as the i-th grid.
+        f_val: numpy.array, or [numpy.array]
+            An array or list of arrays containing the values of the function(s) at the grid points.
+            If it is an array, it's i-th dimension must be of the same lenght as the i-th grid.
             f_val[i,j,k] must be f(grids[0][i], grids[1][j], grids[2][k]).
+            If it is a list of arrays, each array must have the same shape, and
+            f_val[n][i,j,k] must be f[n](grids[0][i], grids[1][j], grids[2][k]).
         grids: [numpy.array]
             One-dimensional list of numpy arrays. It's i-th entry must be the grid
             to be used for the i-th independent variable.
@@ -36,7 +38,18 @@ class LinearFast(MetricObject):
             constant extrapolation. The default is multilinear.
         """
         self.dim = len(grids)
-        self.f_val = f_val
+
+        if isinstance(f_val, list):
+            # First check that all arrays have the same shape
+            if not all([x.shape == f_val[0].shape for x in f_val]):
+                raise ValueError("All arrays must have the same shape.")
+            # Stack arrays in the list across a new dimension
+            self.f_val = np.stack(f_val, axis=-1)
+            self.output_dim = len(f_val)
+        else:
+            self.f_val = f_val
+            self.output_dim = 1
+
         self.grid_list = grids
         self.Grid = CGrid(*grids)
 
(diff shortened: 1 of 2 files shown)


Mv77:
@alanlujan91 not sure how this interferes or interacts with your interpolation rework. Does that change the `econforgeinterp` module?

alanlujan91:
@Mv77 this is a great idea! No worries about my interpolation overhaul. I'll merge your changes and resolve any conflicts. 

The only changes I made to econforgeinterp was moving the file to a HARK.interpolation subdirectory.

Mv77:
Derivatives and decay extrapolators now work :)

Mv77:
> @Mv77, yes, this would be great!
> 
> But I think a good preliminary starting point would be for you do a review of Alan's interpolation notebook to revise it in whatever ways might be appropriate (probably mainly just notation) to make it possible to seamlessly add an example (or examples) using your tool when your tool is in a usable state. That would also get you a bit more into his mindspace which might be useful as you go along. In particular, it is still frustrating to me that we have not figured out how to deal generically (using open source tools) with the "structured but warped" case. The different triangulation methods, and Gaussian Process Regression, I believe are enormously slower than an efficient "structured but warped" tool would be (in which there is a known affine mapping between quadrangles in the warped space and regular rectangles in the unwarped space.

@llorracc I'm not sure this relates to warped interpolation. @alanlujan91 just confirmed that his rework only interacts with `econforgeinterp` to the extent that it changes the location of the file. This is a pretty contained change that is now pretty much ready to go and which does not change the behavior of any of the currently existing code.

I'd be happy to contribute to the warped interpolation discussion but I would not want to tie this PR to that goal.

Mv77:
I've run into some strange potential bug in `econforge/interpolation.py` that affects this PR.
https://github.com/EconForge/interpolation.py/issues/104

Don't merge in yet.

llorracc:
Just to codify a point from our discussion today:  I'd really love to find some framework that is designed to represent a function AND ITS DERIVATIVES at a point, all in a tidy bundle.  

So, for the consumption function, at point $m_{i}$ we would want to measure at least the level and the derivative of consumption with respect to m:

$$[c(m_{i}),c^{\prime}(m_{i})]$$

or in Python

`[cFunc(m[i]),cFunc.deriv('m')(m[i])]`

and using the Envelope result that $v^{\prime}(m)=u'(c(m))$ we would be able to represent the value function consistent with that consumption function by

$$
[v(m_i),v^{m}(m_i) = u^{c}(c(m_i)),v^{mm}(m_i) = u^{cc}(c(m_i))c^{m}(m_{i})]
$$

with corresponding python something like

`[vFunc(m[i]),uFunc.deriv('c')(m[i]),uFunc.deriv('c').deriv('c')(m[i])*cFunc.deriv('m')(m[i])`

Somehow I can't shake the suspicion that surely someone, somewhere, has developed some infrastructure for constructing objects like this, rather than simply representing each item as a completely standalone thing as in Mateo's example.  

(I know that, if you have an interpolating function that has already been constructed, there's a syntax for obtaining its derivatives at a point; the natural syntax for representing the values of these derivatives at given points, to be fed to the interpolator constructor, would be to set the values for each point using exactly the same syntax, and then instruct your interpolation constructor to use the points it has to construct a piecewise function using some specified method.  For example, piecewise Hermite interpolators will generate a function that matches the inputs (in levels and derivatives) at the specified points.)

PR #1249, opened 2023-03-29
Move to pyproject.toml, and redo requirements.txt

MridulS:
With this we can get rid of env.yaml, manifest, setup.py, setup.cfg. Everything goes to pyproject.toml.

To build the wheel we need to use `python -m build`.

TODO:
~Make version dynamic, currently it's hard coded.~ This can taken care of later while releasing 0.14


llorracc:
Pablo/dolo moved to .toml files a year ago, which is enough to convince me that we should do so too.  

I guess this is related to [poetry](https://python-poetry.org/docs/pyproject/)

Anyway, I endorse moving in this direction

alanlujan91:
how do you install this environment? Do you still use `pip install -e .[dev]` ?

MridulS:
Yes, for the base user env just `pip install -e .` and if you want the developer deps `pip install -e .[dev]` should work just fine.

PR #1252, opened 2023-04-04
N-dimensional 'jump_to_grid'

Mv77:
The transition-matrix way of doing things involves spreading the mass of simulated points to an exogenous grid of the state space. We currently can do this for 1D and 2D state spaces with the `jump_to_grid_1D` and `jump_to_grid_2D` methods. This PR adds a method called `mass_to_grid`, which achieves the same goal, but for ND state spaces.

It adds tests that verify that the output of the new function is the same as that of `jump_to_grid_1D` and `jump_to_grid_2D` . In my own tests, `mass_to_grid` is about ~2 times slower than the `jump_to_grid_1D` and `jump_to_grid_2D` methods. Still, I thought it would be good to have the general version around at least for testing higher-dimensional models. If performance ever becomes an issue we can try to make it faster, or code up `jump_to_grid_3D`. But as far as I understand, this part of the matrix-method algorithms is a tiny fraction of their total time so it does not look like the ~2 factor is a concern.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,8 @@
+coverage:
+  status:
+    project:
+      default:
+        target: 70%
+    patch:
+      default:
+        target: 50%
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,8 +1,0 @@
-coverage:
-  status:
-    project:
-      default:
-        target: 70%
-    patch:
-      default:
-        target: 50%
--- a/HARK/mat_methods.py
+++ b/HARK/mat_methods.py
@@ -1,0 +1,256 @@
+import numpy as np
+from numba import njit
+from typing import List
+
+
+@njit
+def ravel_index(ind_mat: np.ndarray, dims: np.ndarray) -> np.ndarray:
+    """
+    This function takes a matrix of indices, and a vector of dimensions, and
+    returns a vector of corresponding flattened indices
+    """
+    # Initialize indices
+    r_ind = np.zeros(ind_mat.shape[1:], dtype=np.int64)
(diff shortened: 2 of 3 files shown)


llorracc:
Looks awesome!  Let's discuss tomorrow.

Mv77:
Codecov is reporting a lot of misses...but I think the tests are pretty exhaustive and I believe it might be because of numba?
@alanlujan91 have you run into this before?

alanlujan91:
We updated codecov to the new version and it seems it is now much more exhaustive... @MridulS might be able to help

sbenthall:
https://stackoverflow.com/questions/26875191/analyzing-coverage-of-numba-wrapped-functions

sbenthall:
"define the NUMBA_DISABLE_JIT environment variable when you run your coverage measurement. "

https://github.com/econ-ark/HARK/blob/master/.github/workflows/coverage.yml

sbenthall:
https://github.com/keurfonluu/fteikpy/blob/master/.github/workflows/ci.yml#L29

Mv77:
The empty commit
Suggested by our friend @MridulS 
Has failed,
And now also empty
Is my soul.

@sbenthall @alanlujan91

Mv77:
I am not sure how that could be, but I can certainly test using the same environment flag when I get back home.

Mv77:
@alanlujan91 nope, still failing coverage

alanlujan91:
it builds on every system/python and runs all the examples, I think we should merge this and try to figure out codecov config later

alanlujan91 (approved):

PR #1257, opened 2023-04-12
move hardcoded params to dictionary

alanlujan91:


HARK/core.py
```diff
@@ -1109,17 +1109,17 @@ def __init__(
         super().__init__()
         self.agents = agents if agents is not None else list()  # NOQA
 
-        reap_vars = reap_vars if reap_vars is not None else list()  # NOQA
-        self.reap_state = {var: [] for var in reap_vars}
+        self.reap_vars = reap_vars if reap_vars is not None else list()  # NOQA
+        self.reap_state = {var: [] for var in self.reap_vars}
```
alanlujan91, line 1113: Or maybe at some point reap_vars was replaced with reap_state, and I should be using reap_state downstream? 

But if reap_vars was replaced with reap_state, why do we still have sow_vars *and* sow_state as attributes? And why do we have track_vars, but not track_state? 

I think there needs to be some consistency, either all vars or all states. or better documentation for what the difference means

alanlujan91: https://github.com/econ-ark/HARK/pull/1257/files#diff-33a0fc9d7783cd0215156e3824a12904f2d025b32245aeeeeeb6b4f5cc509a74R1112-R1113

https://github.com/econ-ark/HARK/commit/a6e2de92eeb4525a96192b58e38a8971be698106#diff-33a0fc9d7783cd0215156e3824a12904f2d025b32245aeeeeeb6b4f5cc509a74R930

alanlujan91: Actually, it appears we need both reap_vars and reap_state, vars has the name of the variables, and state has the values of the variables during the simulation, so they both need to be saved as attributes.

alanlujan91:
if there are no objections to this I will merge it by EOD because it fixes some issues in DistributionOfWealthMPC

PR #1259, opened 2023-04-20
[WIP] Add stable points for PortfolioConsumerType

### LINKED ISSUE #205: solve_distributed_agents solves every agent even when they are identical
Related to #204 

`solve_distributed_agents` calls `solve()` on each AgentType
https://github.com/sbenthall/SHARKFin/blob/cd70a7a18282e348738884c60ee8c378c6b99040/sharkfin/population.py#L258-L262

When we have multiple AgentTypes 'per type' (i.e. `num_per_type`), this introduces unnecessary computation, because the same solution can be used across different AgentTypes.

### LINKED ISSUE #1254: Compute mNrmStE for the PortfolioConsumerType
Compute mNrmStE for the PortfolioConsumerType

See discussion here:
https://github.com/sbenthall/SHARKFin/issues/205#issuecomment-1496453774

I'm curious how this can be done in an extensible way.
Is mNrmStE another kind of 'condition'? I gather not quite, because it is derived from the solution itself.
But we shouldn't need to hard-code the computation of these derived values into each model.

(The architecture fix may be out of scope for this ticket, but I thought I would bring it up.)

alanlujan91:
This is inheriting the way to calculate stable points from `ConsIndShock`, which might not be appropriate for `PortfolioConsumerType` but provides a starting point. 


llorracc:
Alan,

Seb has a PR to the [LucasAssetPrice](https://github.com/llorracc/LucasAssetPrice-Latest) that has a formula that, if it is correct, should be the right impatience condition for the portfolio choice model (when the risky return from the standpoint of the portfolio choice consumer is the portfolio-weighted return). I don’t have time to review Seb’s PR right now — I did the derivations for the approximate formula, which were what was needed for the pedagogical purpose of the handout, and Seb has tried to turn that into the exact formula in levels (that’s what his PR does). If you could review that, then incorporate that formula into the PortfolioConsumerType model (with a comment linking to the LAP lecture notes as the justification), that would be great.

cc’ing Akshay because I’ve mentioned to him that this would be a useful small extension to the derivations in the BufferStockTheory paper that he is reviewing for me. Presumably the condition in my LAP model is identical to impatience formulas in Stachurski, Ma, and Toda (if I am right that their model is one with only rate-of-return risk).

PS.  I've added all three of you as collaborators, so that Alan can merge Seb's PR after review.

PR #1263, opened 2023-05-03
Creating an ndarray from ragged nested sequences is deprecated

MridulS:
This should fix the new test failures popping up in github actions.


alanlujan91:
@MridulS I think this is now looking good, should I merge it so you can test the pyproject PR?

Issue #1276, opened 2023-05-26
pip install does not seem to include some files


alanlujan91:
**Describe the bug**

`pip install econ-ark` does not seem to install HARK.datasets.data

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

- OS: [e.g. Mac/Windows/Linux]
- Econ-ARK and Python version [eg. 0.10.6 and python 3.7]

**Additional context**

alanlujan91:
Leading to issues in DEMARK https://github.com/econ-ark/DemARK/actions/runs/5085827532/jobs/9139693267

alanlujan91:
It also does not seem to install dependencies along with it. Just had to correct some issues in https://github.com/econ-ark/DistributionOfWealthMPC because pip installing econ-ark master did not install all the dependencies. @MridulS @sbenthall can you take a look at this before our next meeting?

MridulS:
This is probably coming from https://github.com/econ-ark/HARK/commit/0ad8bb4f4ab5ad4ccef4f2eda75aa44faee424b0 I will take a look.

MridulS:
@alanlujan91 the datasets part should be fixed now. DemARK build is running now.

> It also does not seem to install dependencies along with it. 

Can you point towards which deps aren't being installed? Current HARK deps are available at https://github.com/econ-ark/HARK/blob/master/requirements/base.txt and these are the ones that should be installed when we do a `pip install`

alanlujan91:
See https://github.com/econ-ark/DistributionOfWealthMPC/blob/master/binder/environment.yml

I had to manually add xarray scipy numpy numba pandas because they were not installed with econ-ark

MridulS:
Hmm, was this to run on binder?

alanlujan91:
yes

PR #1279, opened 2023-06-06
Use setuptools_scm to track data files

MridulS:
This should fix https://github.com/econ-ark/HARK/issues/1276


MridulS:
going for the self merge.

PR #1294, opened 2023-06-28
Update check_conditions()

mnwhite:
This PR incorporates CDC's work on the branch BST-HARK-pre-release-v4, at least with respect to the check_conditions() method for PerfForesightConsumerType and IndShockConsumerType. The new code structures in his branch has not been incorporated, as it represents a large overhaul of HARK. Instead, the condition-checking code has been greatly simplified, but also expanded to be more reader-friendly.

The check_conditions() method is now called automatically as part of pre_solve. This method fills in the attribute conditions_report, which is a string explaining parameter values, various patience factors (and their associated conditions), and an explanation of what the conditions mean *jointly* when verbose is True. A few of those messages need editing or verifying, so this PR is not yet ready to merge.

This PR also moves the calc_stable_points functionality *out* of the solver and into IndShockConsumerType. This change accelerates the solver by a factor of ~3.5x, and there was literally zero value to running calc_stable_points *every single period* during backward solution. The target (and balanced-growth) market resources ratio is not interesting, relevant, or meaningful outside of the infinite horizon, single repeated period case... and it doesn't mean anything until the solution has converged. This functionality is now run automatically as part of of post_solve when appropriate.

The other remaining work item is that all of the various factors that are computed are stored as attributes of the agent, which is not something we want to do anymore. These should be in a dictionary (like history or conditions), but I need feedback on what that dictionary should be called. Maybe factors? It's called Bilt in CDC's dev branch, but I'm not really a fan of that.

No new tests have been added, and I don't think they need to be. The changelog has not been updated, but needs to be.

- [X] Updated documentation of features that add new functionality.


mnwhite:
check_conditions is very, very specific to the PF model and the ConsIndShock model. It could be plausibly extended to KinkedR and *maybe* ConsMarkovModel with more theory work, but it's a heavy lift to go beyond that. I don't think we want to have big code structures or concepts build around check_conditions, because we can only go so far with it-- it's not a general concept, in my opinion.

As for separating the changes to calc_stable_points from the check_conditions stuff... probably, but I'd need to undo work here. It looks like I put all of the "real" work for that into one commit, so maybe it's feasible.

Tests are failing right now because I renamed mNrmStE to mNrmBal. What does StE mean? Steady expectations? It's referred to as the "balanced growth" point in BST, so I changed it to Bal. I suppose I should keep the name for the sake of the tests, and then in a separate PR later change *only* the variable name.

sbenthall:
> check_conditions is very, very specific to the PF model and the ConsIndShock model. It could be plausibly extended to KinkedR and _maybe_ ConsMarkovModel with more theory work, but it's a heavy lift to go beyond that. I don't think we want to have big code structures or concepts build around check_conditions, because we can only go so far with it-- it's not a general concept, in my opinion.

The issue with the current architecture is that while these conditions are specific to the ConsIndShock model, because the ConsIndShock model is the superclass of all other consumption models, we wind up with very heavy, model-specific conditions code in every downstream model, even when it's entirely inappropriate. This is the worst of all worlds.

Making the condition checking code more generic improves that somewhat. Some aspects of the problem, such as wanting printable descriptions for model parameters, may be genuinely general.

But maybe you are right that checking conditions is so model specific that it should not be in the AgentType (or subclasses of it) code at all. In that case, I'd recommend that the conditions-checking functionality live in its own module, and take configuration data and the model as arguments.

mnwhite:
Yes, downstream models will need to overwrite check_conditions() to simply
pass, else they will break or generate unexpected results. The
conditions-checking code isn't "heavy" in the sense of being
computationally intensive, but it is meaningless outside of the context of
that specific model.

But I *don't* think it should live outside of the AgentType subclasses. One
of the improvements that should / will be made is to make details of the
solution method depend on (at least some) of the conditions. As-is, the
consumption function is specified to extrapolate as an exponential decay
toward the limiting linear perfect foresight solution. In some cases, that
limiting linear solution *does not exist*. In the specific case where we
can predict that situation, the solver needs to be told "don't extrapolate
like that, extrapolate like *this* instead". Except no one actually knows
what "*this*" is yet.

sbenthall:
One part in particular that looks like useful general functionality that need not be hard-coded into a model is this part:
https://github.com/econ-ark/HARK/pull/1294/files#diff-7f09d28a3d4136ae35d8835cd1352f060da6e0114fe617c34374c8b57d41a57eR1814-R1822

I understand that you don't have the appetite for generalizing out this sort of functionality. I may get to it once the PR is merged. If you can keep that under consideration as you finish your implementation, that might make the transition towards separating models, solvers, and simulators from each other, which we discussed the other week, go smoother. As is, the `check_conditions` code is a point where the model definition are solver are tightly, and akwardly, coupled.

mnwhite:
I just stripped out the changes to where "stable points" are calculated, so the tests *should* pass now. I'm 95% confident that the only failure mode was the renaming for mNrmStE to mNrmBal.

I still need to go double check some of the messages and make sure I didn't leave any "...I don't know what will happen, this is weird" stuff. The CHANGELOG should also be updated.

Oh, AND: @sbenthall What do you want to call the dictionary that holds auxiliary factors / information that never needs to be used by the solver, but is useful to have around? As-is, I put a lot more crap at the top level of the AgentType, but that should be fixed before merging.

sbenthall:
> Oh, AND: @sbenthall What do you want to call the dictionary that holds auxiliary factors / information that never needs to be used by the solver, but is useful to have around? As-is, I put a lot more crap at the top level of the AgentType, but that should be fixed before merging.

That's a good question. 'auxiliary' is not an obvious label to me. 'definitions'? Maybe @llorracc has an idea?

HARK/ConsumptionSaving/ConsIndShockModel.py
```diff
@@ -1903,169 +1905,296 @@ def get_poststates(self):
...
+        '''
+        Make a string describing this instance's parameter values, including their
+        representation in code and symbolically.
+
+        Returns
+        -------
+        param_desc : str
+            Description of parameters as a unicode string.
+        '''
+        params_to_describe = [
```
sbenthall, line 2016: this is nice

HARK/ConsumptionSaving/ConsIndShockModel.py
```diff
@@ -1903,169 +1905,296 @@ def get_poststates(self):
...
+            #[name, description, symbol, time varying]
+            ['DiscFac', 'intertemporal discount factor', 'β',False],
+            ['Rfree', 'risk free interest factor', 'R',False],
+            ['PermGroFac', 'permanent income growth factor', 'G',True],
+            ['CRRA', 'coefficient of relative risk aversion','ρ',False],
+            ['LivPrb', 'survival probability','ℒ',True],
+            ['APFac', 'absolute patience factor', 'Þ=(βℒR)^(1/ρ)',False]
```
sbenthall, line 2023: conceptually:
 - these are probably more like tuples than lists
 - you have an implicit type here: (string, string, string, bool)
 - this is really data about how the model is configured. Rather than putting it _inside_ the method, it would be easier to (a) modify it and (b) change it in subclasses if this was stored in a data structure outside of the method.

HARK/ConsumptionSaving/ConsIndShockModel.py
```diff
@@ -1609,7 +1611,7 @@ def __init__(self, verbose=1, quiet=False, **kwds):
         self.quiet = quiet
         self.solve_one_period = make_one_period_oo_solver(ConsPerfForesightSolver)
         set_verbosity_level((4 - verbose) * 10)
-
+        self.bilt = {}
```
sbenthall, line 1614: I believe that in the "HARK 2.0 alpha" PR, CDC's use of mispelt English words was to make it easier to find-and-replace them all with a more readable, less opaque name.

Is this being used for anything besides conditions? I would be happy to name this `conditions`. Or could be `built`. Or `derived`.

HARK/ConsumptionSaving/ConsIndShockModel.py
```diff
@@ -1903,169 +1905,296 @@ def get_poststates(self):
...
+            ['CRRA', 'coefficient of relative risk aversion','ρ',False],
+            ['LivPrb', 'survival probability','ℒ',True],
+            ['APFac', 'absolute patience factor', 'Þ=(βℒR)^(1/ρ)',False]
+        ]
+        
+        param_desc = ''
+        for j in range(len(params_to_describe)):
+            this_entry = params_to_describe[j]
+            if this_entry[3]:
+                val = getattr(self,this_entry[0])[0]
```
sbenthall, line 2030: I believe that we now have an alternative way to get at a model's parameters.

I think we are currently storing the parameters dictionary as a dictionary, so that we can pull params from it, rather than relying on `getattr` to pull them from the object. The latter is the "whiteboard problem", and while this still works, I think we've implicitly deprecated this style.

So... consider `self.parameters[this_entry[0]]`

HARK/ConsumptionSaving/ConsIndShockModel.py
```diff
@@ -3071,120 +3163,239 @@ def pre_solve(self):
         self.update_solution_terminal()
         if not self.quiet:
             self.check_conditions(verbose=self.verbose)
+            
+            
+    def describe_parameters(self):
```
sbenthall, line 3168: Confused about what this empty method is doing here.

HARK/ConsumptionSaving/ConsIndShockModel.py
```diff
@@ -3071,120 +3163,239 @@ def pre_solve(self):
...
+        
+        # Make a new entry for weierstrass-p (the weird formatting here is to
+        # make it easier to adapt into the style of the superclass if we add more
+        # parameter reports later)
+        this_entry = ['WorstPrb', 'probability of worst income shock realization', '℘', False]
+        try:
+            val = getattr(self,this_entry[0])
+        except:
+            val = self.bilt[this_entry[0]]    
+        this_line = this_entry[2] + f'={val:.5f} : ' + this_entry[1] + ' (' + this_entry[0] + ')\n'
```
sbenthall, line 3193: This way of constructing a string from these 'entries' is repeated often.
This can be functionalized out.

sbenthall:
Review is inline. Also see the merge conflict on the CHANGELOG.

mnwhite:
I just fixed the last (very tiny) outstanding issue with this, so it's ready to merge, assuming checks pass. Exactly one check failed (for one version of Python?) on the previous commit, which only merged in recent changes from master. Not sure what's going on, but I've seen timeout issues before.

mnwhite:
This is ready to merge @alanlujan91

mnwhite:
Ugh, I just realized that I missed / never saw all of @sbenthall 's review comments. Literally *none* of them, until just now when I got the notification that Alan merged it. To address them:

1) Yes, I think that was mostly a whitespace change. I absolutely and utterly hate the way that black (or whatever we're using) formats code. It makes it much, much harder to read and massively inflates the linecount. If I added the commented out line with # add this at some point, it's because it looks like the simulator is grabbing from the wrong PermShk distribution when t_age = 0. This is because our current format doesn't define the income distribution in the very first period of life.

2) No other classes will be getting similar treatment, so there's no point in making a general code structure for this. I just did that for convenience.

3) CDC has dictated that it shall be called bilt for now.

4) I don't see any empty method there; pre_solve just calls update_solution_terminal, like usual, and it tries to check_conditions if relevant.

5) I think I used that string combining functionality twice.

PR #1306, opened 2023-07-18
[WIP] Labeled dist-of-fun

Mv77:
Aims to fix: https://github.com/econ-ark/HARK/issues/1290

Still need to reach consensus with @alanlujan91 on whether the different behavior of `dist_of_func` and `espected` depending on the class of their input is a feature, a bug, or a temporary solution.


alanlujan91 (approved):

HARK/distribution.py
```diff
@@ -1181,7 +1181,7 @@ def func_wrapper(x: np.ndarray, *args: Any) -> np.ndarray:
 
         if len(kwargs):
             f_query = func(self.dataset, **kwargs)
-            ldd = DiscreteDistributionLabeled.from_dataset(f_query, self.pmv)
+            ldd = DiscreteDistributionLabeled.from_dataset(f_query, self.probability)
```
alanlujan91, line 1184: should probably standardize this with `DiscreteDistribution` ... that's on me

Mv77: Yeah this was the only kind of "bug" this addresses; it makes some cases work that didn't before.

The other stuff is a design question. Happy to merge this in without resolving that, but also happy to have a conversation, agree, and implement it.

PR #1308, opened 2023-07-19
[WIP] Abstract Variables/Symbols

alanlujan91:

### DIFF
--- a/HARK/variables.py
+++ b/HARK/variables.py
@@ -33,7 +33,7 @@ class State(Variable):
     yaml_tag: str = "!State"
 
     def assign_values(self, values):
-        return make_state(values, self.name, self.attrs)
+        return make_state_array(values, self.name, self.attrs)
 
     def discretize(self, min, max, N, method):
         # linear for now
@@ -59,9 +59,8 @@ class Action(Variable):
         Variable (_type_): _description_
     """
 
-    yaml_tag: str = "!Action"
-
     is_optimal: bool = True
+    yaml_tag: str = "!Action"
 
     def discretize(self, *args, **kwargs):
         warn("Actions cannot be discretized.")
@@ -104,7 +103,7 @@ class Auxilliary(Variable):
     yaml_tag: str = "!Auxilliary"
 
 
-def make_state(
+def make_state_array(
     values: np.ndarray,
     name: Optional[str] = None,
     attrs: Optional[dict] = None,
@@ -136,7 +135,7 @@ def make_state(
     )
 
 
-def make_states(
+def make_states_array(
     values: Union[np.ndarray, list],
     names: Optional[list[str]] = None,
     attrs: Optional[list[dict]] = None,
@@ -162,7 +161,8 @@ def make_states(
     attrs = attrs or [{}] * values_len
 
     states = [
-        make_state(value, name, attr) for value, name, attr in zip(values, names, attrs)
+        make_state_array(value, name, attr)
+        for value, name, attr in zip(values, names, attrs)
     ]
 
     return xr.merge(states)
--- a/HARK/variables.py
+++ b/HARK/variables.py
@@ -152,10 +152,14 @@ def make_states(
     Returns:
     States: An arrayray Dataset representing the states.
     """
+    if isinstance(values, list):
+        values_len = len(values)
+    elif isinstance(values, np.ndarray):
+        values_len = values.shape[0]
 
     # Use default names and attrs only when they are not provided
-    names = names or [f"state{rng.integers(0, 100)}" for _ in range(len(values))]
-    attrs = attrs or [{}] * len(values)
+    names = names or [f"state{rng.integers(0, 100)}" for _ in range(values_len)]
+    attrs = attrs or [{}] * values_len
 
     states = [
         make_state(value, name, attr) for value, name, attr in zip(values, names, attrs)
--- a/HARK/variables.py
+++ b/HARK/variables.py
@@ -1,0 +1,164 @@
+from dataclasses import dataclass, field
+from typing import Optional, Union
+from warnings import warn
+
+import numpy as np
+import xarray as xr
+from yaml import YAMLObject
+
+rng = np.random.default_rng()
+
+
+@dataclass
+class Variable(YAMLObject):
+    name: str  # The name of the variable, required
+    domain: Union[list, tuple]  # The domain of the state, required
+    is_discrete: bool = False  # Whether the state is continuous or discrete, optional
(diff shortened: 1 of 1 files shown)


alanlujan91:
@ingydotnet here I am starting the work of building the abstract objects

sbenthall:
Test failures may be due to Python versions (3.8, 3.9, etc.). A lot of these more advanced language features are relatively recent.

HARK/abstract/tests/consindshk.yml
```diff
@@ -0,0 +1,39 @@
+---
+states: !StateSpace
+  variables:
+    - !State
+      name: m
+      short_name: money
+      long_name: market resources
+      latex_repr: \mNrm
```
sbenthall, line 8: I know how you're imagining using the Latex representation.
But my preference would be to not include it in the PR unless you have some demonstration of how it works ready.
4 different ways to name something for a quick demo seems like a lot....

HARK/abstract/tests/consindshk.yml
```diff
@@ -0,0 +1,39 @@
...
+      short_name: &short_name risky share
+      long_name: &long_name risky share of portfolio
+      latex_repr: &latex_repr \stigma
+
+actions: !ActionSpace
+  variables:
+    - !Action
+      name: c
+      short_name: consumption
+      long_name: consumption
```
sbenthall, line 20: Are these fields optional? Can one spill over to the others as a default?
Basically, how can we make these config files lighter weight?

HARK/abstract/tests/consindshk_full.yml
```diff
@@ -0,0 +1,52 @@
+---
+states: !StateSpace
+  variables:
+    - !State
+      name: m
+      short_name: money
+      long_name: market resources
+      latex_repr: \mNrm
+    - !State
+      name: &name stigma
```
sbenthall, line 10: I don't understand what the ampersands are doing here.
Maybe document that?

HARK/abstract/tests/consindshk.yml
alanlujan91: For now these are filler, I want to throw any non-required keys into an attributes dictionary.

HARK/abstract/tests/consindshk.yml
alanlujan91: These are optional, only required is name.

HARK/abstract/tests/consindshk.yml
```diff
@@ -0,0 +1,39 @@
...
+    - !Action
+      name: *name
+      short_name: *short_name
+      long_name: *long_name
+      latex_repr: *latex_repr
+
+post_states: !PostStateSpace
+  variables:
+    - !PostState
+      name: a
```
sbenthall, line 31: Looks like you've repeated the post_states block twice in this file?

Not sure how @mnwhite feels, but maybe we don't need to draw a firm distinction between states and post states like this.

Or the labels could be inside the variable, not part of the document structure.

Compare:
```
var_type_1:
   variables:
       - !VarTypeClass1
           details

var_type_2:
   variables:
       - !VarTypeClass2
           details
```

with

```
variables:
    - !VarTypeClass1
       details
    - !VarTypeClass2   
       details
```

The latter is same information, but fewer lines.

HARK/abstract/tests/consindshk_full.yml
alanlujan91: ampersands are aliases, since one of the states is also a control and a post-state. I will document this more as I make more progress.

HARK/abstract/variables.py
```diff
@@ -0,0 +1,277 @@
...
+class ShockSpace(VariableSpace):
+    shocks: list[Shock]
+    yaml_tag: str = "!ShockSpace"
+
+    def __post_init__(self):
+        super().__post_init__()
+        self.shocks = self.variables
+
+
+def make_state_array(
```
sbenthall, line 215: Is this method duplicated?

HARK/abstract/tests/test_variables.py
```diff
@@ -0,0 +1,19 @@
+import unittest
+
+import numpy as np
+import yaml
+
+import HARK.abstract.variables
+
```
sbenthall, line 7: A test case showing how this data structure could be used in practice would go a long way.

sbenthall:
I think this is a cool demonstration of how PyYAML can be leveraged to make model configuration files in YAML without a custom parser.

The tricky part, as we know, is function definitions.

alanlujan91:
> Test failures may be due to Python versions (3.8, 3.9, etc.). A lot of these more advanced language features are relatively recent.

yep, this might have to be a feature for the future, just getting some initial work done on it

sbenthall:
Mind blown emoji.

One more thing: I'd be keen to see how you would initialize a distribution for a shock directly from YAML.

HARK/abstract/variables.py
```diff
@@ -0,0 +1,277 @@
...
+class Parameters(VariableSpace):
+    """
+    A `Parameters` is a collection of parameters.
+    """
+
+    yaml_tag: str = "!Parameters"
+
+
+@dataclass
+class Auxiliary(Variable):
```
sbenthall, line 101: I agree with Chris that 'auxiliary' is more like a macro.
I wouldn't use 'auxiliary' here, though it makes sense to have this sort of object.

alanlujan91:
@MridulS thank you for fixing the checks! 

Might you be able to help me with the failing test? I'm not sure why `!Parameters` is not being recognized.

HARK/abstract/variables.py
```diff
@@ -0,0 +1,277 @@
...
+
+
+@dataclass
+class Parameter(Variable):
+    """
+    A `Parameter` is a variable that has a fixed value.
+    """
+
+    value: Union[int, float] = 0
+    yaml_tag: str = field(default="!Parameter", kw_only=True)
```
MridulS, line 78:
```suggestion
@dataclass(kw_only=True)
class Parameter(Variable):
    """
    A `Parameter` is a variable that has a fixed value.
    """

    value: Union[int, float] = 0
    yaml_tag: str = "!Parameter"
```

This should fix it @alanlujan91

alanlujan91: This makes it work, but doesn't quite do what I want, I think. 

I would like parameters to be instantiated as `Parameter("name")` or `Parameter("name", value)`, and not `Parameter("name", value = value)`, which I think this does. 

You might be right that is probably better to do this more simply, removing the requirement to use py>=3.10 and mixing in @dataclass with `YamlObject`.

PR #1310, opened 2023-07-21
Point links at ``docs.econ-ark.org``

AA-Turner:
cc: @MridulS, @DominicWC

This PR updates links within notebooks and some documents to point to the new documentation homepage, [docs.econ-ark.org](https://docs.econ-ark.org/).

A


MridulS:
looks good thanks!

The `latest` tag in readthedocs reminds me that we should also turn on versioned docs. It's not turned on by default right? 

The docs branch will always point to the latest HEAD.

PR #1312, opened 2023-07-21
Remove ``.readthedocs.yml``

AA-Turner:
cc: @MridulS 

This removes a no-longer-needed configuration file.

A


MridulS:
Should we wait till we have got all the desired changes in?

AA-Turner:
RTD builds [are currently failing](https://readthedocs.org/projects/hark/builds/) -- this PR wouldn't remove the RTD site, but only configuration for already-failing builds.

If desired, we could try and fix RTD?

A

MridulS:
> If desired, we could try and fix RTD?

Ah no, let's not invest too much time there. Let's move to docs.econ-ark.org as soon as possible :)

PR #1314, opened 2023-07-24
Add a grid of links to the landing page

AA-Turner:
cc: @MridulS @DominicWC

This PR is on top of #1313, and adds a landing-page grid of links, similar to SciPy, NumPy, MatPlotLib, et al.

Open to suggestions about the order and content of the four boxes, I've chosen what I think is a sensible starting point.

A


MridulS:
I like it, we can reiterate on this later :)

PR #1319, opened 2023-07-24
Update README.md

DominicWC:
Changes the docs link so that it goes to the home page

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,7 +183,7 @@ Looking for help? Please open a [GitHub issue](https://github.com/econ-ark/HARK/
 
 Current releases follow [Semantic Versioning](https://semver.org/). For more information please see the [Release documentation](https://github.com/econ-ark/OverARK/wiki/Release-Management).
 
-## API Documentation
+## Documentation
 
 Documentation for the latest release is at [docs.econ-ark.org](https://docs.econ-ark.org/index.html).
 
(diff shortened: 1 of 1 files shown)


DominicWC:
@MridulS, @AA-Turner 

It's a small change, but now that the wiki home page is spruced up, we should link to it.

AA-Turner:
Ahh, that was an intentional choice as the header says "API Reference" so I thought we should link directly to the relevant section. Perhaps there should be a different link to the documentation in general? The first paragraph links only to the website, and a docs link could be added there.

A

DominicWC:
Maybe we could change the header so it just says "documentation"?

README.md
```diff
@@ -183,9 +183,9 @@ Looking for help? Please open a [GitHub issue](https://github.com/econ-ark/HARK/
 
 Current releases follow [Semantic Versioning](https://semver.org/). For more information please see the [Release documentation](https://github.com/econ-ark/OverARK/wiki/Release-Management).
 
-## API Documentation
+## Documentation
 
-Documentation for the latest release is at [docs.econ-ark.org](https://docs.econ-ark.org/reference/).
+Documentation for the latest release is at [docs.econ-ark.org](https://docs.econ-ark.org/index.html).
```
AA-Turner, line 188:
```suggestion
Documentation for the latest release is at [docs.econ-ark.org](https://docs.econ-ark.org/).
```

AA-Turner (approved):

PR #1320, opened 2023-07-24
Use Python 3.10 in the documentation workflow

AA-Turner:
cc: @MridulS

Given ``numba`` now supports Python 3.11 I tried to increase all the way, but ``interpolation`` has an upper-limit on Python 3.10 -- see https://github.com/EconForge/interpolation.py/issues/103 and https://github.com/EconForge/interpolation.py/pull/105.

A

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -28,10 +28,10 @@ jobs:
     steps:
     - uses: actions/checkout@v3
 
-    - name: Set up Python 3.9
+    - name: Set up Python 3.10
       uses: actions/setup-python@v4
       with:
-        python-version: "3.9"  # Numba doesn't support Python 3.11 [2023-05]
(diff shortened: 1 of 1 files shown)


MridulS:
Yeah, this needs to be fixed upstream first.

alanlujan91:
I've had issues with econ-ark=0.13 and python=3.10 because of interpolation, so does it even work with 3.10?

MridulS:
It's not updated with conda-forge, it works on pip btw.

Mv77:
Hey, Pablo just released a 3.11 compatible version of interpolation. https://github.com/EconForge/interpolation.py/releases/tag/v2.2.5

Wonder if this lets us advance to python 3.11.

MridulS:
It's a github release, not released on https://pypi.org/project/interpolation/#history or https://anaconda.org/conda-forge/interpolation/files

PR #1321, opened 2023-07-24
Use a gallery for the example notebooks

AA-Turner:
cc: @MridulS @DominicWC

This converts the list of example notebooks to a 'gallery' preview that looks slightly nicer. The titles of some of the notebooks were very long, so I took liberties to rename them.

A


DominicWC:
Would it be possible to add every single Notebook from the "Consumption Savings" folder in the examples section of hark into the documentation?

DominicWC:
If we want users to look at the API reference then we'd want to structure it more like NumPy's with a written description of all the aspects of a function, and examples of it being used, rather than just using AutoDocs.

MridulS:
> If we want users to look at the API reference then we'd want to structure it more like NumPy's with a written description of all the aspects of a function, and examples of it being used, rather than just using AutoDocs.

I agree, numpy is also just autodocs. It's just their inline docs are much better :D

PR #1322, opened 2023-07-24
Execute notebooks on rendering documentation [DO-NOT-MERGE]

AA-Turner:
cc: @MridulS

Opened for discussion, as currently the PR doesn't work for all notebooks (``import EstimationParameters as Params`` fails in the life cycle model).

Would this be useful? It has the benefit of ensuring that the content in the notebook renders and is accurate, but significantly slows down the rendering of the documentation.

A


AA-Turner:
I'm working on an alternate idea which hopefully will mitigate these downfalls -- will update shortly (I'm away Tue & Wed so likely after that)

A

PR #1326, opened 2023-07-28
Add documentation section to the contributing guide

AA-Turner:
cc: @MridulS @DominicWC 

I also went through and broke very long lines at reasonable points, and moved some links out-of-line to markdown link targets.

A


Documentation/guides/contributing.md
```diff
@@ -270,14 +364,54 @@ multiple entries.  If no output, this is just "None".
...
+2. Run `sphinx-build`:
+
+   ```bash
+   $ sphinx-build -M html Documentation HARK-docs -T -W
+   ```
+3. View the HTML files
 
-Finally, if you write a new model module, the HARK team asks that you also provide a short mathematical writeup of the model as a PDF. This document does not need to go into great detail about the solution method for the model or the functions and classes included in the module, merely specify the economic model and provide a summary of how it is solved. See [ConsumptionSavingModels.pdf](https://github.com/econ-ark/HARK/blob/master/Documentation/ConsumptionSavingModels.pdf) for an example of this.
+   ```bash
+   $ firefox HARK-docs/html/index.html
```
MridulS, line 407: Maybe we should just ask contributors to open the file `HARK-docs/html/index.html` instead?

PR #1328, opened 2023-07-29
Add process to automatically update example notebooks

AA-Turner:
cc: @MridulS

This is the promised alternative to #1322.

This workflow strips all output from example notebooks and then re-executes them all, before opening a PR with the changes. I chose to have the workflow open a PR rather than committing directly itself to allow reviewer/human intervention if need be.

The workflow will only run on changes to the ``examples`` directory -- I thought this was a better tradeoff than having it run on every change to the core HARK code, though of course changes to core HARK code will change outputs in the example notebooks.

A


MridulS:
> This workflow strips all output from example notebooks and then re-executes them all, before opening a PR with the changes. I chose to have the workflow open a PR rather than committing directly itself to allow reviewer/human intervention if need be.

I like this, thanks!

Let's try running this :)

PR #1330, opened 2023-07-31
[bot] Execute example notebooks

MridulS:
This PR was automatically generated to re-execute
the example notebooks for use in the documentation.


MridulS:
Hmm, @AA-Turner it seems like we need to use a personal access token to be able to generate these PRs.

I have added mine for now, we can keep with that but I would prefer if we can use github actions to create the PR instead.

PR #1343, opened 2023-08-19
Let `ConsRiskyAssetModel` handle time-varying Rfree

Mv77:
I was getting errors trying to get `ConsPortfolio` to use time varying interest rates `Rfree`.

The culprit was that the share-limit calculation assumed `Rfree` is constant. This PR fixes the issue.

### DIFF
--- a/Documentation/CHANGELOG.md
+++ b/Documentation/CHANGELOG.md
@@ -22,6 +22,7 @@ Release Date: TBD
 - Addresses [#1255](https://github.com/econ-ark/HARK/issues/1255). Makes age-varying stochastic returns possible and draws from their discretized version. [#1262](https://github.com/econ-ark/HARK/pull/1262)
 - Fixes bug in the metric that compares dictionaries with the same keys. [#1260](https://github.com/econ-ark/HARK/pull/1260)
 - Fixes bug in the calc_jacobian method. [#1342](https://github.com/econ-ark/HARK/pull/1342)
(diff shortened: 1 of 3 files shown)


Mv77:
This is a pretty trivial PR and most of you are in Lausanne. Also, all tests pass and test coverage increases.

I'll merge without review.

alanlujan91:
Can you confirm that the share limit applies to the 1 period solution and not just to the infinite horizon problem? I thought the share limit only existed in infinite horizon problem

Mv77:
Hey, I think so. See below.

In any case we were computing that object even for LC calibrations (whether we're using it or not, which I'm not sure). And so time varying `Rfree`'s resulted in errors.

Here's why I think the result applies:

As long as an agent has CRRA utility, no income, and decides on the share of wealth in stocks continuously, it is the case that $$v_t(m_t) = \Omega_t u(m_t)$$ for some possibly time-varying constant $\Omega$. This is true in infinite horizon and in finite horizon as long as every period satisfies that description (no income, etc) and the CRRA is constant.

So the portfolio problem is
$$\max_{s_t} E[\Omega_{t+1}u( a_t (R + s_t(\tilde{R} - R)))] )] = \Omega_{t+1}a_t^{1-\rho}\max_{s_t} E[u( R + s_t(\tilde{R} - R) )] )],$$
and I think we're using the FOC to the second problem to get the limiting share. If we interpret the limiting share as the share that would be optimal as assets become so large that income is irrelevant, it applies.

Does that sound right @llorracc ?

Issue #1346, opened 2023-08-28
library of  ConsumptionSaving model definitions


sbenthall:
We now have several PRs that address parts of the HARK 1.0 refactoring.
It's gotten to the point where further progress is blocked because we don't have the model information for our current ConsumptionSaving models in a compact Python format. (See e.g. #1296 #1295, as well as #1292, and #1283)

I think the best thing at this point would be to start writing model configuration objects in 'the new style' for the existing ConsumptionSaving models. We could put these in a directory like `HARK/models`. These could then be used for testing generic algorithms.

The hard part is capturing all the nuances that are embedded in the current ConsumptionSaving, and I expect @mnwhite is going to have to ultimately review these.

Naturally, because we don't have functionality that processes these model definitions yet, it will be hard to validate that they are 'correct'; but I think the solution to this chicken-and-egg problem is to irradiate the dinosaur. If we do this, we will learn a lot about what we will need from a future YAML file format.

But it's also something that anybody could work on, and it's even a good entry task for somebody with a structural microeconomics background. (In my opinion, there's no reason why we should add model definitions for other subfields of Economics besides ConsumptionSaving). cc @JohnRGreen

mnwhite:
This is kind of ongoing on several fronts. My expanded simulation framework has model definitions (at least as far as needed for simulation) for a bunch of models, and the rest can be added very easily. Work on the new stage-based representation also continues separately.

PR #1347, opened 2023-08-29
perfect foresight model data

sbenthall:
This PR aims to demonstrate how we could proceed with #1346 by writing lightweight model definition files in pure Python.

Models defined in this way could then be the source of equations that are integrated into AgentType (see #1292 ) or used as an input to generic Monte Carlo simulation (#1296) or solution algorithms (#1286).

I'm not committed to the form in this PR, but it reflects the kind of information that's useful for downstream algorithms (as in (#1296) and is otherwise quite minimal. I'd argue that any problem representation we develop in YAML will be parsed into something quite like this, so this is a fine start which we could then expand upon as needed.


sbenthall:
add fischer two period:
https://github.com/econ-ark/DemARK/blob/master/notebooks/FisherTwoPeriod.ipynb

But @mnwhite should this be a 'different model'?

sbenthall:
Fisher 2-period
http://www.econ2.jhu.edu/people/ccarroll/public/LectureNotes/Consumption/2PeriodLCModel.pdf

sbenthall:
I've added CHANGELOG and automated tests are meaningless in this case.

This PR has no new functionality but is needed for further testing of #1296 (Generic monte carlo simulation).
I would like to defer the documentation step on this until the #1296 PR, which demonstrates the functionality.

@mnwhite @alanlujan91 could you please review and, if passing, merge?

alanlujan91 (approved):

alanlujan91 (approved):
Looks good to me!

Issue #1354, opened 2023-10-26
Modifications to A Gentle Introduction to HARK


sidd3888:
### Your First HARK Model: Perfect Foresight

- Better to address survival probabilities as $`\aleph_t`$ to begin with, or express both survival probabilities and growth factors as constants from the outset, rather than jumping back and forth
- Introduction of notation $`D_t`$ for probability of death is unnecessary and unused, as this is just $`1 - \aleph_t`$

### Solving an Agent's Problem

- Change "paramterization" to "parameterization" (pargraph after `PFexample.solve()`)
- Change "associated the lecture notes" to "associated lecture notes" (after next code block)

### Changing Agent Parameters

- Upon creating the new instance `NewExample` with the original discount factor and growth factor $`\Gamma = 0.01`$, the diagram shows the consumption to be much higher than in the case with $`\Gamma=1.01`$, though the commentary indicates that the graph should emanate from near 0

### Your Second HARK Model: Adding Income Shocks

- The notation used for the second model with a borrowing constraints and income shocks is a little confusing. For instance, whereas variables are normalized, the value function is also written as $`v_t`$ instead of $`V_t`$ as before, though $`U(\cdot)`$ is maintained as is. This makes it confusing to the unversed reader. Maybe explaining some of the new notation clearly would also help
- Remove large whitespace between the term $`\max_{c_t} U(c_t)`$ and the expected continuation value in the statement of the value function
- Some of the variables, such as standard deviation of the permanent and transitory income shocks are stated inside lists. Might help to explain why these are in lists as opposed to being scalars

@alanlujan91: If I can help in making some of the necessary changes, happy to do so

PR #1355, opened 2023-11-05
Changes to Gentle Introduction to HARK

sidd3888:
Changes made according to points mentioned in #1354

Primarily made changes to the description of the models or fixed typos. Only thing I changed in the code was to generate the plot of the model with updated parameters and then removing the pertinent code, so that the image matches the description (also detailed in the issue).

@alanlujan91 sent you an email about one of details. Ready to merge from my end once that is addressed


alanlujan91 (approved):
Great job! I made some notational changes but agree with all your changes.

sidd3888:
A couple of things I wanted to clarify:
- I don't know if this is an accident, but the first statement of survival probability is as a constant, not as $\aleph_t$, though setting it to a constant is clarified only two paragraphs down. Switching between $\aleph$ and $\aleph_t$ may appear notationally inconsistent
- The use of $\mathsf{D}_t$ for probability of death: this notation is not used in the notebook after this. Should we be retaining it here? Just for this notebook, might suffice to say the problem ends with probability $1 - \aleph_t$
- Everytime someone runs the `NewExample` model with $\Gamma = 0.01$, the graph would revert to one without the updated parameters, as it is left for the reader to change the parameters. I can slightly change the wording here to indicate that the graph would obey the characteristics described **upon** changing the parameters

Once these things are addressed, this is good to go (but you'll have to merge them, I can't)

alanlujan91:
@sidd3888 were you going to redo this one?

mnwhite:
I adjusted a few words, including the age-varying notation in the first paragraph. I moved the sentence about death probability to be semicolon clause in the LivPrb sentence; it's less prominent now, as Siddarth is correct that it serves little purpose.

The existence of the reference to death probability might be a legacy of older notation. I think we used to use D for death probability and \cancel{D} for survival probability, making it more useful to actually mention death probability.

Some of the checks are now failing after adjusting text, which is surprising. I'll look into it.

mnwhite:
It looks like it's because I cleared notebook output, but Sphinx relied on its existence. Will fix.

mnwhite:
Ok, I'm baffled @sbenthall . I'm now getting an error from ruff-format, which is apparently unhappy with a coding style change I made-- I think I moved a comment or two. I can't get ruff to run on my computer, as it says that the notebook isn't a JSON file at all.

alanlujan91:
I think this should pass now, is it ready to merge @mnwhite ?

mnwhite:
It should be. Any guesses for getting ruff to work for me?

alanlujan91:
first thing i did was merge all the changes in master branch
then update ruff 
then `ruff format .\examples\Gentle-Intro\Gentle-Intro-To-HARK.ipynb`

@MridulS is there a way to let ruff make and push the changes needed instead of not passing?

PR #1358, opened 2023-11-10
Update PerfForesightConsumerType

sidd3888:
Made some fixes in the language and description of the model and the code. One mistake left to be clarified.

Paragraph in the subsection **Example parameter values to construct an instance of PerfForesightConsumerType** that elaborates, "In contrast, we could specify a life-cycle model...", which states that the parameter `T_cycle` could be set to $1$. However, it is already set to $1$ and this would not allow room for age-varying parameters of income growth and survival probability.

Ready to merge in all other aspects.

CC: @alanlujan91


mnwhite:
Ah, that should say "cycles", not "T_cycle". I'll look at that comment; we
don't want misleading or false information in docstrings.

sidd3888:
@mnwhite The thing I pointed out, and you were looking into, is still pending to be resolved. Will open a new PR to resolve that once these stylistic issues have been decided upon. Thanks.

PR #1362, opened 2023-11-22
Update KinkedRconsumerType.ipynb

sidd3888:
Made typo and LaTeX environment changes as usual. Link to the paper detailing the underlying theory was broken, so I linked the AER and NBER websites hosting the published version and its expanded edition instead. If there is some alternate link on the JHU website, which you would want to use instead, I can make the change.

CC: @llorracc @alanlujan91


alanlujan91:
@sidd3888 the text "\verb" seems to show up in the notebook preview, is this a bug?

sidd3888:
@alanlujan91 Yes, I have changed that on my local repo... I'll push the changes here in just a bit. By the way, where can I see the preview, as I am currently just relying on compatibility across multiple IDEs

sidd3888:
@alanlujan91 is the \verb still visible on the preview? I looked at the notebook on all possible platforms, barring the preview you mentioned. If it is still the case, please let me know where I can see this preview? Thanks.

Issue #1375, opened 2024-01-29
utilities.py: metadata bug in make_figs()


kennetheva:
When calling make_figs(), an unrecognized argument error occurs at:

```
plt.savefig(
            os.path.join(target_dir, f"{figure_name}.jpg"),
            metadata={"CreationDate": None},
) 
```

The [matplotlib.pyplot.savefig documentation](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.savefig.html) says metadata is currently not supported for .jpeg files. Commenting out the metadata argument stops the error and the figure is produced as expected.

Should `metadata={"CreationDate": None}` be removed from the JPEG savefig call until support is added in a later version of the package?

llorracc:
The reason for suppressing the CreationDate was that when the code gets run, even if the image part of a regenerated figure is bit-by-bit identical to the corresponding part of the former figure, if each of them is stamped with a different CreationDate git will declare that the files have changed.

Omitting the CreationDate from the metadata therefore enormously reduced the number of files that git identified as different.

If matplotlib has been modified to exclude the creation date and other metadata by default, this workaround to prevent it from including CreationDate should be no longer needed.


From: kennetheva ***@***.***>
Date: Monday, January 29, 2024 at 15:50
To: econ-ark/HARK ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [econ-ark/HARK] utilities.py: metadata bug in make_figs() (Issue #1375)

When calling make_figs(), an unrecognized argument error occurs at:

plt.savefig(

            os.path.join(target_dir, f"{figure_name}.jpg"),

            metadata={"CreationDate": None},

)

The matplotlib.pyplot.savefig documentation<https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.savefig.html> says metadata is currently not supported for .jpeg files. Commenting out the metadata argument stops the error and the figure is produced as expected.

Should metadata={"CreationDate": None} be removed from the JPEG savefig call until support is added in a later version of the package?

sbenthall:
It might be better to use `.gitignore` to suppress version control over the jpegs entirely.

llorracc:
Matt,

As the original author of savefig, can you look into this?  I'm puzzled that a lot more examples have not cropped up.  

Also, my preference would be that whenever a figure is saved, it gets saved in jpg, pdf, svg, and png forms.

mnwhite:
[...]

PR #1386, opened 2024-02-20
Remove metadata argument for jpg in make_figs

mnwhite:
Metadata argument was not supported for this filetype, was throwing error. CDC wants a minor release for this. I'm going to put in one more annoying behavior fix before I do that.

### DIFF
--- a/HARK/tests/test_HARKutilities.py
+++ b/HARK/tests/test_HARKutilities.py
@@ -3,9 +3,9 @@ This file implements unit tests to check HARK/utilities.py
 """
 # Bring in modules we need
 import unittest
-import matplotlib.pyplot as plt
 from types import SimpleNamespace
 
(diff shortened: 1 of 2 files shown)


mnwhite:
@alanlujan91 This is a one line bugfix that Chris wanted for #1375 , turning off an illegal argument for jpg in make_figs. He wants a minor release for this (for HAFiscal, I think), so I'm going to do one more little thing (__repr__ behavior for IndShockConsumerType) and then release 0.14.1. Review/merge when you have time.

alanlujan91 (approved):
This looks good to me. 

Ideally we should have a quick test that used to fail but now passes, but this is such a minor fix I'll leave that up to your discretion.

mnwhite:
Let me see if I can add a quick test after I grab lunch.

PR #1388, opened 2024-02-23
Allow for a pre-built grid in `LinearFast`

Mv77:
The `CGrid` constructor of the econforge interpolation library can take a substantial time if you build many interpolators.

This PR allows you to pass a pre-constructed `CGrid` object to `LinearFast`.

This can save a lot of time in instances where you are building many interpolators over the same grid. For example, if you have consumption functions for many Markov states that are defined over the same asset grid. (cc: @wdu9 ?)

### DIFF
--- a/Documentation/CHANGELOG.md
+++ b/Documentation/CHANGELOG.md
@@ -8,6 +8,16 @@ For more information on HARK, see [our Github organization](https://github.com/e
 
 ## Changes
 
+### 0.15.0
+
+Release Date: TBA
+
+### Major Changes
+
+### Minor Changes
+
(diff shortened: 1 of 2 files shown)


alanlujan91:
This is a useful change. 

However, currently `interpolation.py` is preventing us from moving up to `python=3.12`. Do you think it will be maintained and supported going forward?

alanlujan91 (approved):

Mv77:
> This is a useful change.
> 
> However, currently `interpolation.py` is preventing us from moving up to `python=3.12`. Do you think it will be maintained and supported going forward?

¯\_(ツ)_/¯

PR #1392, opened 2024-03-02
Fix cons bequest

alanlujan91:
@mnwhite  I think I fixed what we talked about over email

@dedwar65 also I think this fixes the bug you were dealing with

I don't like the way I did this, with some repetitive code, but hopefully it gets it done. 

@dedwar65 I can't add you as a reviewer, but please review this


mnwhite:
Yes, this is the solver fix I mentioned. Let me look at the failing checks to see what's up.

mnwhite:
It looks like a tiny discrepancy got introduced into the KinkedR model with this PR, possibly including an issue with the mNrm grid for its value function. I'll take a look.

mnwhite:
Well that was easy to find. This PR swaps CubicHermiteInterp for CubicInterp in ConsIndShockModel, and I don't know what the new interpolator does / how it works. It looks like it has a check for x_list to be strictly increasing, which will cause problems when constructing EndOfPrd_vFunc in the KinkedR model: a=0 appears twice consecutively. That's easily fixable in one line.

The discrepancy in the cFunc of 0.0001 at the query/test point is due to the change in the cubic interpolator. I expected more consistent accuracy between cubic interpolator versions, so this might be a real error. Digging deeper...

alanlujan91:
@mnwhite you are right, the fix i just made removes CubicHermite

There might be a true error somewhere else, but I think for the sake of this PR the change i made is sufficient, and we can address the remaining issue in a separate PR

alanlujan91:
pre-commits are still failing, which are addressed in #1396, but hopefully everything else should pass

mnwhite:
Let's see if the checks pass here, then merge. I agree we can do the other fix in another PR-- it's a pre-existing "harmless" error that was exposed by CubicHermite's additional check. It doesn't cause problems in the current tests because the *only* way the double a=0 would cause problems is if EndOfPrd_vFunc were evaluated at *exactly* zero (and even then it might not actually throw a NaN, it depends on how searchsorted works). It should be fixed, but it's not critical here.

But it looks like there might be a formatting "error", ugh.

alanlujan91:
@mnwhite I propose we merge this if it passes everything else, then we merge #1396 which should fix the formatting error issue

mnwhite:
Sounds good.

mnwhite:
The only failing tests here are the formatting, which is being fixed in another PR, and codecov. But the only untested new lines are additional parameter typing checks, which would throw errors if we wrote tests for. I'm merging this, then will look at and merge the pre-commit PR.

PR #1393, opened 2024-03-05
Redo modern python packaging

alanlujan91:
based on https://learn.scientific-python.org/development/

bunch of checks will probably fail... this is just a test


MridulS:
This is a nice idea but we really need to be disciplined in the way PRs are merged and changelogs are managed to use the recommendations properly.

PR #1400, opened 2024-03-22
Add more consumption-saving tests

mnwhite:
This PR *only* adds new tests (and makes tiny adjustments to old ones) for consumption-saving models and should be merged *before* my solvers overhaul branch. But not yet because I'm actively adding tests.


mnwhite:
I think these are the tests I wanted to add. Two are commented out because those capabilities are new in my other PR.

PR #1395, opened 2024-03-08
DO NOT MERGE: Functional Solvers

alanlujan91:
This PR contains some suggestions for @mnwhite's ongoing solver overhaul. 

Simple functions that can be re-used within a "family" of models, but should not be expected to "jump" out of their own file.

PR #1404, opened 2024-04-05
Update Wealth portfolio solver

alanlujan91:


mnwhite:
It looks like this PR makes changes to ConsIndShock as well, moving some functions out of the solver. Looking through that, there's at least one actual code change, to `MPCmaxNow` vs `MPCmaxEff`. These are not (necessarily) the same value. The solver constructs the *unconstrained* consumption function and combines it with the constrained consumption function via `LowerEnvelope`. `MPCmaxNow` is the MPC at the lower bound of the unconstrained function, whereas `MPCmaxEff` is the *effective* maximum MPC for the actual consumption function. `MPCmaxNow` shouldn't be overwritten by the "should this just be 1?" line, because its value is still used later on.

llorracc:
Am in agreement with Matt.  But maybe to prevent confusion for future users, the name of MPCmaxNow should be changed to something like `MPCmaxIfNoLiqConstr` and perhaps a comment could be added to the code to explain why it is computed and what it means.

alanlujan91:
This PR builds on #1395, so it's more appropriate to discuss MPC constructions there and leave this for later

HARK/ConsumptionSaving/ConsWealthPortfolioModel.py
```diff
@@ -1,624 +1,438 @@
...
+            euler,
+            method="toms748",
+            bracket=[lower_bound, upper_bound],
+            args=(a_nrm, CRRA, WealthShare, WealthShift, end_dvda_now[a_idx]),
+        )
 
-        # Construct a basic solution for this period
-        self.prepare_to_optimize_share()
-        self.optimize_risky_share()
+        cNrm_now[a_idx] = result.root
```
alanlujan91, line 355: @mnwhite could you give me your thoughts on this re-worked solver? 

This problem does not have an invertible euler equation, so I instead try to do root-finding

everything else should be the same as the simple portfolio problem, and I have confirmed that when WealthShare=0.0, it is indeed the same result

alanlujan91: might be easier to just see the solver rather than try to see the diffs 

https://github.com/econ-ark/HARK/blob/51e1d357617406d0b482fcb8f2613c9d2bd94af2/HARK/ConsumptionSaving/ConsWealthPortfolioModel.py

mnwhite:
This is on my immediate list.

alanlujan91:
@mnwhite can you take a look now?

mnwhite:
I haven't gone through every single line of math in the auxiliary functions, but I have three pieces of feedback on the solution method:

1) When I did the math/code for the "chi from omega" trick, I didn't know there was such a thing as the WealthShift parameter. The existence of that parameter changes things significantly for two reasons. First, I would want to go back and redo the math with that shifter in place to make sure everything goes through without complication; I put about 80% confidence on it, but that's not very high. Second, it means the functional representation isn't great: we'll never actually approach one of the limits of the function, and there might be a better representation due to that. I strongly recommend writing out the math by hand from scratch with WealthShift to be super duper extra sure that all of the aNrm in the math without it can just be replaced with aNrm+WealthShift.

2) I need to double check the envelope condition math. When I first looked at the code, I thought "this isn't right", but after writing out my reasoning, I'm revising that to "this isn't how I would have done it", and I need to do the math myself. I now think what you've done is probably correct.

3) We/you should do some robustness checking on the logic of whether/when to add an extra gridpoint for aNrm=0. I think the proper logic is more complicated than you have, and should involve a check of whether WealthShift > 0. My concern is that there are situations where two gridpoints are generated at exactly mNrm = 0.

PR #1407, opened 2024-04-06
remove references to `parent_class_name` and some redundant constructions

alanlujan91:


HARK/ConsumptionSaving/ConsIndShockModel.py
```diff
@@ -3366,15 +3363,10 @@ def __init__(self, **kwds):
...
-        PerfForesightConsumerType.__init__(self, **params)
+        super().__init__(**params)
 
         # Add consumer-type specific objects, copying to create independent versions
         self.solve_one_period = solve_one_period_ConsKinkedR
-        self.update()  # Make assets grid, income process, terminal solution
-
-    def pre_solve(self):
-        #        AgentType.pre_solve(self)
-        self.update_solution_terminal()
```
alanlujan91, line 3377: Not needed, parent already updates and updates solution terminal in that update

HARK/ConsumptionSaving/ConsIndShockModel.py
```diff
@@ -2663,7 +2661,6 @@ def pre_solve(self):
         # Update all income process variables to match any attributes that might
         # have been changed since `__init__` or `solve()` was last called.
         #        self.update_income_process()
-        self.update_solution_terminal()
```
alanlujan91, line 2666: Not needed, update method already updates solution terminal

mnwhite:
I can't work out why, but one of the *seemingly* extraneous calls to `update_solution_terminal` is actually critical for exactly one test: the Jacobian methods on `IndShockConsumerType`. The answer isn't just a little off, it's off by orders of magnitudes. I didn't write those methods and don't really understand them, but their test is really simple. Very strangely, putting an additional call to `update_solution_terminal` into the test itself doesn't fix the issue.

Anyway, all tests pass after resurrecting that one line, so...?

alanlujan91:
@wdu9 I believe the jacobian work was all done by you. Do you have any guesses as to why this is critical?

For a summary of what this PR does: We found there are numerous calls to `update`, `update_income_process`, or `update_solution_terminal` for a a few models, when obviously you should only have to do this once. To correct for this I removed the superfluous calls to ensure we only update everything once. Every other test in HARK works, except for `test_Jacobian_methods`. https://github.com/econ-ark/HARK/actions/runs/8584487477/job/23524783772

mnwhite:
Per meeting today, I'm merging this, then will try to get @wdu9 to look into the mysterious maybe-bug.

PR #1410, opened 2024-04-15
Add constructor architecture

mnwhite:
Beginning work on the `construct` method of `AgentType`. As of this moment, a basic prototype works for income distributions. Still a lot more to develop and test.


mnwhite:
As noted in the issue where I proposed this, there are two significant design points to work out:

1) How to handle situations where it's natural to construct *multiple* objects in the same function. The most obvious case is the first example I implemented: `IncShkDstn`, `PermShkDstn`, and `TranShkDstn`. I put in an ugly workaround, but something else should be done before this is merged.

2) Whether and how we want constructors to indicate whether they've made a time-varying or time-invariant object. I'm beginning to come around to the idea that these concepts (to the extent they're in HARK now and will be until several more overhaul steps are completed) should *not* be intended to be changed at the instance level. Rather, they're a fixed fact of the `AgentType`. There's a very specific reason they were defined at the instance level, but that reason no longer exists.

mnwhite:
Can someone tell me how to run the pre-commit routine locally? I can run black, but I don't know how to make lint work.

sbenthall:
`$ pre-commit run --all-files`

sbenthall:
more info in the developer guide here: https://docs.econ-ark.org/guides/contributing.html#pre-commit-hooks

alanlujan91:
> `$ pre-commit run --all-files`

you probably don't want to run all files, only your changes

mnwhite:
A bunch of tests needed syntax changes, but everything passes now. The other things that I'm moving to constructors will be much easier from here on out, and should be completed on Monday.

For this PR, I'm going to set the goal at getting everything in constructor functions that should be (plus some alternates that can be swapped in). After this is merged, in a separate branch/PR, I'm going to tackle undoing/simplifying the custom init methods for each AgentType subclass, and get them all just using AgentType.__init__ ... That also means that the default dictionary should be a class attribute.

alanlujan91:
We should be doing super().init everywhere

mnwhite:
Even where `super().__init__` is used, there can be problems when `update()` is called by a parent class as part of init before the child class is "ready" for it, leading to unexpected results. In the not-too-distant future, I hope all our AgentType subclasses inherit directly from AgentType, and don't have any special init stuff.

mnwhite:
If tests pass on this commit, the next commit will change *exactly one line* of code, and revise tests to fit. As is, the "construct pLvlGrid by simulation" method draws on AgentCount, which is fine *as long as AgentCount is actually large*. The number of eventual simulated agents shouldn't be used in the *solution* of the problem, even extremely indirectly. So I'm just going to have it use a *large number* of agents no matter what. This is also why the hardcoded seeds don't matter in this function/method.

mnwhite:
The remaining tasks on this PR are pretty small, almost optional:

- make some example MrkvArray constructors
- make terminal solution constructors for types that are missing it
- consider moving one method in ConsAggShockType to a constructor (but maybe not)
- make a method for AgentType that reports what the user *needs* to make a valid type
- a few "alternate example" constructor functions, to demonstrate swappability

llorracc (commented):
This also allows for transitory shocks, I assume?

Someday I want a tool that can handle the case that I think captures almost all of what can be said about annual income processes: It's hard to do better than a process with a fully permanent shock and a transitory shock with an MA(1) term (as time aggregation would imply).  I take it that your setup here allows for purely transitory terms but probably not for MA(1) transitory terms?

llorracc:
Also: Is this ready for use?  Ryan Edwards is working on making a new Aiyagari REMARK using HARK and I want him to use the latest and greatest tools ...

mnwhite:
> This also allows for transitory shocks, I assume?
> 
> Someday I want a tool that can handle the case that I think captures almost all of what can be said about annual income processes: It's hard to do better than a process with a fully permanent shock and a transitory shock with an MA(1) term (as time aggregation would imply). I take it that your setup here allows for purely transitory terms but probably not for MA(1) transitory terms?

This PR is just a restructuring. I've made *very* small changes to the constructor methods when converting them to functions, but they're almost trivial adjustments. This isn't about economics or new capabilities at all. So all the same income processes are in there that we had before.

It *might* be possible to do an MA(1) transitory process using `MarkovConsumerType`, but I'd have to work out whether there's a better way than having N^2 discrete states, where N is the number of transitory shock values.

mnwhite:
Meant to say before posting: If it's not, we can add something later that can handle this income process. Let me think about it.

mnwhite:
The simplest way to do it in HARK (read: the way that requires the least new code) is to make a subclass of `MarkovConsumerType` with an adjusted simulation procedure. When solving the model, the Markov state can be "today's transitory shock", which simply gives the forward-looking agent a little bit of information about next period's *transitory income* (weighted average of today's shock and tomorrow's shock).

When *simulating* the model, you just need a mechanism that enforces that (the index of) today's transitory shock is the same as (the index of) today's discrete state.

mnwhite:
Still behind schedule, as there were more terminal solvers than I remembered. Still to do the following small tasks:

- make some example MrkvArray constructors
- a few "alternate example" constructor functions, to demonstrate swappability

HARK/core.py
```diff
@@ -327,6 +329,164 @@ def __str__(self):
...
+                has_no_default = {
+                    k: v.default is inspect.Parameter.empty
+                    for k, v in inspect.signature(constructor).parameters.items()
+                }
+                temp_dict = {}
+                any_missing = False
+                for j in range(len(args_needed)):
+                    this_arg = args_needed[j]
+                    if hasattr(self, this_arg):
+                        temp_dict[this_arg] = getattr(self, this_arg)
```
sbenthall, line 393: Is there a way to do this without hasattr/getattr/setattr, using manipulations of a data structure rather than an object's attributes?

HARK/core.py
```diff
@@ -327,6 +329,164 @@ def __str__(self):
...
+                                # Record missing key-data pair
+                                any_missing = True
+                                missing_key_data.append((key, this_arg))
+
+                # If all of the required data was found, run the constructor and
+                # store the result in parameters (and on self)
+                if not any_missing:
+                    temp = constructor(**temp_dict)
+                    setattr(self, key, temp)
+                    self.parameters[key] = temp
```
sbenthall, line 408: Are these parameters, or model constructs, or both?

what if there was a self.constructs dictionary that was separate from parameters?

HARK/core.py
```diff
@@ -327,6 +329,164 @@ def __str__(self):
...
+        self._missing_key_data = missing_key_data
+        return
+
+    def describe_constructors(self, *args):
+        """
+        Prints to screen a string describing this instance's constructed objects,
+        including their names, the function that constructs them, the names of
+        those functions inputs, and whether those inputs are present.
+
+        Parameters
```
sbenthall, line 428: Python conventions for printed representations of things is to create a `__repr__` method, or to display docstrings.

HARK/core.py
```diff
@@ -327,6 +329,164 @@ def __str__(self):
...
+        N_keys = len(keys)
+        yes = "\u2713"
+        no = "X"
+        maybe = "*"
+        noyes = [no, yes]
+
+        out = ""
+        for i in range(N_keys):
+            key = keys[i]
+            has_val = hasattr(self, key)
```
sbenthall, line 451: could be looking up in parameters?

HARK/core.py
```diff
@@ -327,6 +329,164 @@ def __str__(self):
...
+        Parameters
+        ----------
+        *args : str
+            Optional list of strings naming constructed inputs to be described.
+            If none are passed, all constructors are described.
+
+        Returns
+        -------
+        None.
+        """
```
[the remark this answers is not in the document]
mnwhite: I think we want to leave it general/simple that any function can be a constructor, to keep barriers to entry low.

HARK/core.py
mnwhite: Yes, `construct()` does this but I forgot to put that line in here. Will fix.

HARK/core.py
```diff
@@ -327,6 +329,164 @@ def __str__(self):
...
+            keys = list(self.constructors.keys())
+        N_keys = len(keys)
+        yes = "\u2713"
+        no = "X"
+        maybe = "*"
+        noyes = [no, yes]
+
+        out = ""
+        for i in range(N_keys):
+            key = keys[i]
```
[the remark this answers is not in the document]
mnwhite: I do use `i` below, because I track which work items have been completed in a boolean vector.

HARK/core.py
```diff
@@ -327,6 +329,226 @@ def __str__(self):
...
+            Removes information from self.parameters dictionary and own namespace.
+
+        Returns
+        -------
+        None.
+        """
+        if param_name in self.parameters:
+            del self.parameters[param_name]
+        if hasattr(self, param_name):
+            delattr(self, param_name)
```
sbenthall, line 350: is there any way to avoid this redundant storage of the constructs?

Maybe constructs should not be stored in parameters, but in self.constructs (or something)

sbenthall:
This all looks very good to me.

The one thing which I think could be improved is the confusing way in which constructed objects are stored, both (!) as an entry in `parameters` and as a new attribute on the model object. Neither of these seem quite right to me.

What about a `self.constructs` dictionary?

HARK/core.py
mnwhite: That's for *deleting* things. Right now, I've written things to be intentionally redundant, because we're "between two worlds". The medium run goal is to remove the top level namespace at all, only putting them in `parameters`.

I don't want a separate `constructs` dictionary, because there's no meaningful distinction between `parameters` and `constructs`. It might be that `parameters` was the wrong name for that dictionary.

sbenthall:
Ok, I understand that removing object attribute storage entirely is a goal, but something to be implemented in a different PR.

I'll leave further notes on #889

mnwhite:
Fixed CHANGELOG

PR #1412, opened 2024-04-17
Multivariate Lognormal

sidd3888:
Added a class corresponding to the bivariate lognormal distribution (current features generalizable to the multivariate lognormal). Need to make sure things are in order before proceeding.

@alanlujan91 @llorracc


alanlujan91:
@sidd3888 this is a good start

alanlujan91:
@sidd3888 I added a few comments on the code, but I think what would be more useful is having tests and/or demonstration (a notebook) of what the code is doing to see if it's working well or as expected.

alanlujan91:
@sidd3888 pull changes from master and the errors should go away

PR #1415, opened 2024-04-23
MAINT: bump up to py3.12 for doc build

MridulS:


MridulS:
The test failures are coming from the changes in interpolation.py

alanlujan91:
can you take a look at this when you have time @Mv77?

Mv77:
I'd be happy to.

But to make sure I have the right version of everything, is there any yaml file with the environment that I should use while checking the tests?

alanlujan91:
`pip install -e .` should install all the dependencies, we got rid of `environment.yml` files

Mv77:
I'll run than in a clean python 3.12 env and report back.

Mv77:
Can confirm all econforge tests fail locally. Debugging.

Mv77:
It looks like there are now unrelated errors having to do with assertions?

It is strange that I did not get those locally.

Mv77:
https://github.com/econ-ark/HARK/pull/1420 fixes a typo that is making the 3.12 versions fail.

What I can't figure out is why 3.11 is failing. There are some failures related to properties of numpy arrays that I can not reproduce locally on a fresh 3.11 env.

Mv77:
Do they fail locally for you @alanlujan91 ?

alanlujan91:
> Do they fail locally for you @alanlujan91 ?

no, but they are failing on the mac tests and I have a windows, so I am trying to "guess" what a mac would like, we'll see if it passes!

Mv77:
MAC: It just doesn't work^{TM}

alanlujan91:
@Mv77 I think i'm getting close, only 1 failing test on mac now!

alanlujan91:
UGHHH everything is passing except for coverage

I agree with @MridulS that we should remove it

alanlujan91:
I think this will let us finally bump to python 3.12! 

Thank you @MridulS and @Mv77!

@mnwhite please review and merge when appropriate

PR #1416, opened 2024-04-24
Sequence Space Jacobian Type

alanlujan91:
Integrating @wdu9's work on SSJ. 

The results are very wrong... which is probably due to my lack of experience with SSJ

Would appreciate help from @wdu9 and @Mv77


alanlujan91:
hmm why are checks not running? @MridulS

MridulS:
I see this in the runs:

```
FAILED HARK/ConsumptionSaving/tests/test_SequenceSpaceJacobianType.py::test_Transition_Matrix_Methods::test_calc_tran_matrix - AttributeError: 'SequenceSpaceJacobianType' object has no attribute 'taxrate'. Did you mean: 'tax_rate'?
FAILED HARK/ConsumptionSaving/tests/test_SequenceSpaceJacobianType.py::test_Jacobian_methods::test_calc_jacobian - AttributeError: 'SequenceSpaceJacobianType' object has no attribute 'taxrate'. Did you mean: 'tax_rate'?
```

Mv77:
A minor comment is that, since SSJ is a method and not a model, I don't think that SSJType is appropriate. Maybe NewKeynesianType?

alanlujan91:
tried to fix the errors on the notebooks, but there's still something off with the "NewKeyensianType"

mnwhite:
I'm going to try to look at this ASAP. Alan's latest commit turns off some (probably unnecessary) pre-solve behavior. The change(s) that I made to `calc_jacobian` had interactions with `presolve`. Before my edits, the method would edit the attribute `_cFunc_terminal`, which is a *class* attribute, not instance. I also added functionality to `solve` so that the user can indicate that they *don't* want to run `presolve` as part of `solve`. That enables doing things like a custom "terminal solution", like in this case.

AMonninger:
After my last commit, the first column of the Jacobians was correct, but the others were still off.
Unfortunately, the changes in the meantime made them all wrong again. I assume it has something to do with the changes made to `IndShockConsumerType` and the updating process...
Steady states and policy functions of the infinite horizon solution look fine and unchanged. Only the Jacobian part is off.

mnwhite:
That's a good hint about what's going on. To further help me hypothesize
solutions while not actually looking at the code, which variable is "the
first column"?

AMonninger:
In the temporarily created notebook: IndShockConsumerType_HANK.ipynb
In cell 13, we solve:
CJACR, AJACR = Agent_GE.calc_jacobian("Rfree", T)
CJACN, AJACN = Agent_GE.calc_jacobian("labor", T)

The first column of CJACR should start at 0.4 and then converge to 0.

This part is created by the ZerothColAgent in the function `calc_jacobian`

mnwhite:
This material has been incorporated into HARK with #1447

PR #1414, opened 2024-04-23
move calibration out of DBlock definition, see #1371

sbenthall:
This separates out the calibration data from the dynamics in the DBlock

- [x] Tests for new functionality/models or Tests to reproduce the bug-fix in code.
- [x] Updated documentation of features that add new functionality.
- [x] Update CHANGELOG.md with major/minor changes.

PR #1417, opened 2024-04-24
portfolio consumer model with recursive block definition working with generic monte carlo

sbenthall:
This PR:
 - introduces a new 'new model', an unnormalized idiosyncratic shock model with (optional) portfolio share based on SolvingMicroDSOPS notes
 - an RBlock which can combine several DBlocks into one block.
 - passing smoke tests for running this model in the ATMC simulator

The model in `consumer.py` is an attempt to demonstrate how a value, such as `R`, can be either calibrated directly or set by a dynamic block (in this case, the portfolio block) while reusing the same consumption block data.

It also shows the use of a `tick` block to set $k_{t+1} = a_t$


sbenthall:
This PR builds on #1414, which should be merged first

sbenthall:
this is really work towards #1374

sbenthall:
TODO: Allow nested RBlocks

sbenthall:
TODO: Demo in a test the way that the consumption DBlock can be used with or without the RBlock

sbenthall:
I believe this PR is ready.
Requesting final review from @alanlujan91 or @mnwhite

This PR now supports recursively defined models, and demonstrates how a model can be composed of pre-made blocks.

Specifically, it shows how the portfolio consumption model can be built from the basic consumer model by adding one more block. This block makes the `R` rate of return an endogenous state that's a function of the risky return and `stigma`, overriding the default calibration of this value.

sbenthall:
Added normalized version of consumption block

sbenthall:
I've added some more functionality to the DBlocks in this PR pursuant to aligning it with the recent SolvingMicroDSOPs work.

A DBlock, which represents what we might have called a 'stage' at one point, can now:

- Given a continuation value function, provide a _state-rule value function_, which is the value of being at state $x$, with _realized_ shocks $z$, given a decision rule $\pi$
- Given a continuation value function and decision rule, provide a _decision value function_, which is the value of being in the (shock realized) state before the agent makes a decision. (This is the value at the 'decision' step/tac.)
- Given a continuation value function, a decision rule, and arguments for discretizing shocks, provide an _arrival value function_, which is the value of the (pre-shock) state at the beginning-of-period/arrival 'step'/tac.

These can all be computed directly using the dynamics and reward functions provided in the DBlock, and are intended to be helpful utilities to support solution algorithms.

PR #1419, opened 2024-04-24
[econforge interp] Change order -> k

Mv77:
Adjusts the name of a parameter that changed in the newer versions of econforge's interpolation


Mv77:
Note this is a PR to Mridul's PR https://github.com/econ-ark/HARK/pull/1415, not to master

PR #1425, opened 2024-05-15
Flatten default initializer dictionaries

mnwhite:
This PR has (or should have) no *functional* changes. All it does is reorganize the default initializer dictionaries for our AgentTypes so that all parameters for that class are present and explained in one place. All parameter values remain the same, and the dictionaries themselves should be exactly identical as before.

This branch builds from #1410 under the assumption that it will be merged in relatively soon.


mnwhite:
So I *said* it wouldn't change any dictionaries, but then I found an extraneous CubicBool=True and removed it. Test values changed by 0.0001 to 0.001.

mnwhite:
This branch is ready for review and merge. It extends #1410 , so that should be merged first.

There are very few functional changes, mostly just "flattening" of dictionaries so that each AgentType subclass is better documented in the file in which it appears. A few tests and examples had to have their formatting changed slightly to account for dictionary differences. Only one test had a *value* change, and that was due to removing a (likely mistaken) CubicBool=True in an example dictionary.

The tests that changed will be easier to see when #1410 is merged into master.

sbenthall:
I'll review this once #1410 is in. As is, it's hard to isolate the changes specific to this PR

PR #1427, opened 2024-05-20
allow use of string expressions instead of lambdas when definition new format models.

sbenthall:
This PR allows the user to define 'new format models' with string mathematical expressions instead of lambdas.


sbenthall:
This builds of #1417, which should be merged first

mnwhite:
@sbenthall Very minor file conflicts now that #1417 has been merged: changed import statements, changes to example model that show the "non-lambda" version, new test added (why does that register as a "conflict"?), one changelog line, etc. You can probably better select the right version than I can.

sbenthall:
This PR is ready for review.

PR #1434, opened 2024-05-28
Integration Example [WIP]

AMonninger:
Example notebook of Constructing End of Period (Marginal) Value Function using Transitionmatrices instead of Expectations.

This example notebook tests the idea of @mnwhite . So far, the two methods do NOT construct the same function. Could @mnwhite check for conceptual/ coding errors?

- [x ] Updated documentation of features that add new functionality.


mnwhite:
@AMonninger I just pushed a commit that revises the notebook. The method works as intended now, and displays comparisons at the bottom. Important changes that I put in (there might be more):

- You either need to implement an artificial borrowing constraint *OR* include a tiny probability of zero income. I put in UnempPrb=1e-10. Without this, the minimum allowable bNrm will vary between the two methods.
- The mNrmGrid and bNrmGrid had incorrect offsets. This is critical.
- Small, but helped with debugging: use different grid sizes for b and m so you can verify dimensions
- Did some variable relabeling to make it more accurate
- You were post-multiplying, not pre-multiplying
- Need to use the "pseudo-inverse trick" when representing the interim marginal value function
- And needs to have a point added at 0 (or fix this to generalize it).

The consumption functions constructed by the two integration methods now match as expected. The one using the new integration method is always *a little lower* than the old one, especially in the highly concave region of the cFunc. This is because the new method incorporates more uncertainty than the old one: as long as the theoretical probability of getting to some m from some b is positive, it is included in the expectation. That's not the case with a fixed set of integration nodes.

Take a look at the notebook and let me know if you want to meet to discuss. I didn't fix all typos or style issues.

llorracc:
Cool!

Looking forward to hearing about this-Fri?

AMonninger:
@mnwhite Thank you so much. Those changes are brilliant!
But if I understood you correctly, the method only works without Unemployment Probability (eg no pointmass of the pdf at 0/'IncUnemp'). Is there a way to circumvent that limitation?

mnwhite:
I am going to work on this. We liked it, and then it fell off the map.

PR #1447, opened 2024-06-03
[WIP] A third attempt to get SSJ running!

mnwhite:
This branch was developed on a Zoom call with Will Du on June 3, 2024 (1pm-3pm). It mostly copy-pastes necessary code from #1424 and makes formatting changes for current version of HARK. The SSJ example notebook runs correctly, except for the Nonlinear Response section, which Will thinks is more trouble than it's worth for now. Tests should run properly (maybe?), examples won't execute because of those cells. I will remove those cells on the next commit.


mnwhite:
@alanlujan91 @llorracc I made this branch this afternoon with @wdu9 and then cleaned it up just now. His example SSJ notebook now works in the latest version of HARK, and all tests/examples pass.

I'm not sure exactly what we want to do with this, versus Alan's PR #1416 , which moves some of the specific material out of ConsIndShockModel.py (which is very cluttered, in our opinion) and into a specialized file.

mnwhite:
I've now (more or less) reproduced the work Alan did to move the HANK-specific stuff into a separate file. If all tests pass, I'm going to rename the SSJ example file, because it's kind of weird as-is.

@alanlujan91 Your PR has additional notebook(s) I think, or other example files. Do we want to bring those over? I feel like the SSJ example notebook needs *a lot* more explanation and exposition about what's going on.

mnwhite:
Tests and examples should run properly now. The commits above moved more material from Alan's #1416 . I'm meeting with Chris and Will in 45 minutes to discuss next steps, but I *think* this is ready to merge, and further work can be done in a different branch.

mnwhite:
The tests are yielding an error for Documentation / Render (push) but not Documentation / Render (pull request). Looking at the output, it might be an internet connection failure during testing?

PR #1455, opened 2024-06-14
Update NARK so it compiles with new @resources

llorracc:
Some minor tweaks to fix compilation problems.


mnwhite:
The LaTeX files look like they refer to computer-specific files. Is that correct?

llorracc:
Shouldn’t be.  If so I need to fix.

PR #1456, opened 2024-06-14
Expand SSJ exposition

mnwhite:
This PR improves and extends the SSJ notebooks. It includes work from Will Du (already merged in from his branch).

As of this moment, there's a tiny bit more to do.


mnwhite:
Test targets have been adjusted, and numba version has been pinned below 0.60.0 (from master branch).

mnwhite:
This branch will be merged once tests finish running. The only failure before was due to a last minute variable name change that was missed, causing the final two cells of a notebook to note execute.

Tomorrow morning I will release HARK 0.15.1 so that the SSJ example(s) are in public HARK for CEF and Chris's presentation before CEF.

PR #1460, opened 2024-06-18
discretize() methods for blocks

sbenthall:
Introduces `discretize` methods for blocks, which apply the discretization parameters to the shocks.

In the RBlock case, this is done recursively.

The point of this is to provide an API which a user can call to modify the 'true' version of a model into a discretized one.
The discretized model can then be used for solution and simulation.

(in the future, a YAML configuration will interact with this API)

### DIFF
--- a/Documentation/CHANGELOG.md
+++ b/Documentation/CHANGELOG.md
@@ -8,6 +8,19 @@ For more information on HARK, see [our Github organization](https://github.com/e
 
 ## Changes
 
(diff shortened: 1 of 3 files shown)


sbenthall:
This PR is ready for review.

mnwhite:
Something is going on with GitHub-- now this branch is having trouble with auto-merge-checking after updating master.

Issue #857, opened 2020-10-22
Python model configuration


sbenthall:
Based on the meetings today: the simulation and solution aspects of HARK will need to both refer to a single, canonical description of a model.

This model configuration should define:
 - the model variables (including states, controls, and shocks)
 - initial values for the variables
 - transition equations for the state variables
 - exogenous distributions or processes for the shocks
 - anything needed by the solvers for the controls
 - constraints (Gamma)
 - parameters 

A major design question that has not been resolved yet is how to bundle this model description up into frames/stages. See #798 

Another question raised by this is how models should inherit configurations from each other. Currently, model inheritance is done through Python's class inheritance. But as more of a model's specification is removed from HARK class code and into a configuration object, it opens up other options. In any case, the class inheritance is not quite standardized yet across the library and could use some thought.

sbenthall:
Will need to revisit this towards the end of the push to 1.0 to see what smaller parts can be accomplished

sbenthall:
will split this into separate issues and close.

mnwhite:
Closing and archiving as for HARK 1 roadmap. See Seb's tags for where it's being actively worked on.

PR #1463, opened 2024-06-22
standalone model configuration file with model, approximation, solution, and simulation parameters

sbenthall:
The goal of this PR is to implement a working example of a HARK configuration file, in Python, as per #857. (Note the ticket originates from 2020).

Such a configuration should include information:
 - about "true" model dynamics (stages, blocks,etc.), in
 - approximations adopted for computational reasons
 - instructions about how to solve the model
 - parameters for simulating the model (with the given solution)

While this is functionality that's similar to the HARK 0.x models, this PR aims to implement model configuration subject to several constraints:

 - There's a single source of truth for any information about the model's structural equations
 - The configuration is a _data structure_, not a _program_, in the sense that it relies only on basic data types without cross-referencing between sections. This requirement is there because the next step after this PR is to implement a parser that will transform a YAML file into this kind of configuration object.
 - As much as possible, it actually operates with other not-model-specific components in HARK.

This work is informed by [work in late 2023 on the HARK 2.0 model specification language](https://github.com/econ-ark/OverARK/tree/master/Development).

Some tricky parts remaining to implement include:
 - Constructing shocks using mathematical expressions for input parameters from configuration objects
 - Settling on a syntax for aspects of the configuration which are really more like 'workflows' than like model details

### DIFF
--- a/Documentation/CHANGELOG.md
+++ b/Documentation/CHANGELOG.md
@@ -8,13 +8,14 @@ For more information on HARK, see [our Github organization](https://github.com/e
 
 ## Changes
 
-### 0.15.2 (in development)
+### 0.16.0 (in development)
 
 Release Date: TBD
 
 #### Major Changes
 
-none
+- Allows structural equations in model files to be provided in string form [#1427](https://github.com/econ-ark/HARK/pull/1427)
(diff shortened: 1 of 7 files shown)


sbenthall:
This PR builds on #1427 which should be merged first.

PR #1464, opened 2024-07-02
Construct shocks with mathematical expression arguments using calibration scope

sbenthall:
The "new model files" have had a problem, which is that they depend on fully constructed Distribution objects for each shock. That meant that things like, for example, the standard deviation of the transitory shock could not depend on a calibration dictionary passed in at simulation (let alone solution) time.

This PR fixes this issue.

Now, model files can define shocks with a tuple, e.g. `(Lognormal, {"mean": "Rfree + EqP", "std": "RiskyStd"})`.

When the Monte Carlo simulator is initialized with a calibration dictionary containing values for `Rfree`, `EqP`, and `RiskyStd`, the expressions in the strings are evaluated in the scope of calibration, and these values are passed through to the constructor for `Lognormal`.

Note that since we presume that the shocks are truly exogenous, these expressions need not be evaluated at simulation runtime in that dynamic context. (Though that can be done in principle at the expense of runtime performance).

Tests pass with new 'consumer.py' models that take advantage of this new feature.

p.s. one small change was to provide an alternative means of initializing a Lognormal distribution with mean/std directly, instead of mu/sigma. This could have been implemented in many ways.

### DIFF
--- a/Documentation/CHANGELOG.md
+++ b/Documentation/CHANGELOG.md
@@ -17,6 +17,7 @@ Release Date: TBD
 - Adds a discretize method to DBlocks and RBlocks (#1460)[https://github.com/econ-ark/HARK/pull/1460]
 - Allows structural equations in model files to be provided in string form [#1427](https://github.com/econ-ark/HARK/pull/1427)
 - Introduces `HARK.parser' module for parsing configuration files into models [#1427](https://github.com/econ-ark/HARK/pull/1427)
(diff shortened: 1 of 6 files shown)


sbenthall:
This is ready for review @mnwhite 

It's another incremental change, based on what we discussed earlier.

It's a requirement for an upcoming PR which should enable some basic YAML configuration of blocks.

PR #1472, opened 2024-07-11
Update notebooks

alanlujan91:
@DominicWC Chris wants to add some of the notebooks in examples/ConsNewKeynesian

I think that involves just adding simlinks to Documentation/example_notebooks, but is there anything else that needs to be changed in the documentation config?


DominicWC:
The notebooks need to be added to index.rst in the Documentation/overview folder. Additionally, the four added notebooks need to be checked. Sphinx won't let them be added if they:
1. Don't have a title
2. Have any headers that increase in size (go from ##-># or ###->##)

It looks like Jacobian_Example, SSJ_example, and Transition_Matrix_example all have headers out of order, which then throws an error.

Issue #1182, opened 2022-10-16
Improve the way we keep track of 'employment'


Mv77:
**Describe the bug**
The `get_shocks` method of `IndShkConsumerType` creates a variable that keeps track of whether an agent is employed or not. It determines whether the agent is employed by checking whether its transitory shock matches `IncUnemp`:
https://github.com/econ-ark/HARK/blob/ae96a0066563cf1f6020127232d31306be1dd425/HARK/ConsumptionSaving/ConsIndShockModel.py#L2302

This works fine if the problem is one of infinite horizon or a LC problem where `IncUnemp` is a constant. But we flexibilized the class to allow for time varying unemployment income, and this code won't work in that case (`IncUnemp` will be a list). Thus, for `self.EmpNow` to be informative in these settings, this line of code needs updating.

I might do this myself when I have a little more time but wanted to write it down so that we all remember.

mnwhite:
See also #984 ... This bit of code needs to be taken out entirely, because "unemployment" need not exist at all now. Users will need to make a subclass of IndShockConsumerType that adds this bit of accounting code if they need it for their purposes.

Issue #984, opened 2021-03-15
EmpNow in IndShockConsumerType ?


sbenthall:
It looks like there is some residual use of `EmpNow` in IndShockConsumerType even though it is not recorded as a shock.

https://github.com/econ-ark/HARK/blob/05ef324cfc7b0351b68063ce3cbcc70a47f62783/HARK/ConsumptionSaving/ConsIndShockModel.py#L2161-L2163

It looks like the cstwMPC code assume that this is there, and is available for inspection.

Is this part of the model, or not?
Can it be removed?

I can hack around it in the cstwMPC refactor

llorracc:
“Unemployment” in the model is just captured by a special value of the transitory shock.  Like, if there were no unemployment insurance it would be zero.  In the paper we tabulate various facts for ‘unemployed’ people, so it is useful to create a Boolean to mark them.  If the main HARK ConsIndShock model code doesn’t do this probably it would be worth adding it there and just using it in cstw but I don’t feel strongly about it

mnwhite:
Yes, that's how it's constructed in cstwMPC. It's not drawn as a shock
itself, but instead calculated as something like:

`self.EmpNow = self.TrankShkNow == self.IncUnemp`

Or whatever the correct phrasing is now with dictionaries. The
cstwMPC project wanted to calculate some statistics by employment state,
and this was the most straightforward way to do that.

mnwhite:
Chris wants this functionality to remain in HARK, but it actually needs to *leave* and be moved to cstwMPC (and other applications that use it). Going forward, the simulation code for ConsIndShock can't be sure that IncUnemp *exists*-- it's an input to a constructor, not a fundamental part of the model that's required to exist. So these two lines need to be excised and put elsewhere (currently 2048-2049 of ConsIndShockModel):

`self.EmpNow = np.ones(self.AgentCount, dtype=bool)`
`self.EmpNow[TranShkNow == self.IncUnemp] = False`

llorracc:
I strongly object to removing the unemployment shock, as a possibility, from the ConsIndShock model.  It is an extremely valuable tool for understanding the fundamental mechanics and math of the model.  The whole BufferStockTheory paper relies profoundly and intimately on this.

mnwhite:
That's not what I'm saying-- unemployment is not being removed as an
option. It's still in HARK, it's still the default, and it will remain that
way.

All that I'm flagging for removal is 2 lines in the simulation code that
[...]

PR #1473, opened 2024-07-13
Remove employment accounting lines

Mv77:
Two lines that were added to the income process of consindshock for `cstwMPC` are now causing trouble. See https://github.com/econ-ark/HARK/issues/984 and https://github.com/econ-ark/HARK/issues/1182.

This PR removes those lines. This does not break anything _within_ HARK itself.


Mv77:
I imagine this _might_ break the compatibility of the `cstwMPC` REMARK with the current version of HARK. However, it is also my understanding that REMARKs are now properly pinned to a specific version of HARK, so this should not be an issue.

mnwhite:
Mateo isn't the developer we deserve, but he's the developer we need.

Mv77:
Hah thanks. I take that as your approval and so I'll merge this in.

PR #1492, opened 2024-09-03
Fix identity function derivatives and remove ancient/dead test code

mnwhite:
This PR fixes a typo in the derivative methods of interpolation.IdentityFunc that has gone undetected for years. It also (finally) deletes hundreds of lines of unused tests that I wrote it in 2015 and early 2016.


sbenthall:
It would best if there was an automated test for this.

PR #1494, opened 2024-09-08
Update Parameters class implementation

alanlujan91:
This pull request updates the implementation of the Parameters class. It includes changes to the initialization, getitem, setitem, get, set_many, is_time_varying, to_dict, and update methods. The changes ensure better functionality and error handling in the Parameters class.


HARK/core.py
```diff
@@ -551,6 +647,353 @@ def describe_constructors(self, *args):
         return
 
 
+from typing import Any, Dict, Iterator, List, Set, Tuple, Union
+
+
+class Parameters:
```
sbenthall, line 653: It looks like there are now two copies of the Parameters class in this file?

PR #1495, opened 2024-09-09
Modernization fixes for calc_euler_error_func

mnwhite:
It looks like this method lacks tests and was never updated for changes in HARK.distribution. This commit fixes some notation so that the method works again.


mnwhite:
Arguably this PR should have a test added.

mnwhite:
@akshayshanker This can also be reviewed and merged, once tests finish.

PR #1499, opened 2024-10-01
Some expansion to the MonteCarloSimulator functionality

sbenthall:
This PR makes a few small additions and changes to the Monte Carlo simulation functionality.

- It introduces a new class, `MonteCarloSimulator`, which is a lot like the `AgentTypeMonteCarloSimulator`, except that this doesn't make any assumptions about birth/mortality, and it doesn't have the 'read_shocks" functionality. This is a more stripped down version of what's already there.
  -  It introduces some duplicate code, and maybe there's a way to make the implementation more efficient by thinking about what sort of abstract or super- class a Simulator should be. This will require some thinking about architecture which I think it would make sense to work with @mnwhite on

-  It makes a small change so that reward variables (i.e. `u`) get computed in simulations, and added into the `history`.  This is useful when trying to compare the outcomes of agents across different models and parameterizatiosn.

- The `apply_fun_to_vals` function compartmentalizes a coding pattern used widely in the Monte Carlo Simulator code. I didn't work it through the library yet since it replaces just one line of code with something arguably more readable.


mnwhite:
This class has a `state_vars` attribute that's defined at the class level and never seems to be used. If it's used, it should be at the instance level, but this looks like a harmless inclusion at this time.

PR #1519, opened 2025-01-09
Financial advice using LifeCycle Model demo.

akshayshanker:
Here is a notebook for financial advice analysts. 

This is a final draft, but some details are WIP and need input and discussion. Please do not merge yet. 


mnwhite:
Comments:

- Should we omit mention of bequest motives, because they're not part of a financial planner's fiduciary duty?
- "... to understand the average behaviour of a hypothetical population of" This sentence just ends.
- Should this notebook point to HARK instruction at all? As is, it says, "If you are new to the HARK toolkit, we recommend starting with the "Gentle Introduction to HARK" notebook." ... I'm not sure we want to derail people while they read this notebook.
- Expand on concave utility a little bit, and use the word concave.
- Need to say that the consumer wants to maximize the *expected discounted sum* of utilities they will experience.
- Theta is a transitory income shock, not transitory income.
- Missing "and" in the Extended to Portfolio Choice Bequest Motive heading
- "The basic LCP model assumes that all wealth is ultimately spent by the end of life." That's misleading or false. Bequests being "accidental" does not mean that all wealth is spent by the end of life in models.
- Soften language around bequest motive. You have "suggesting an additional bequest motive", but there are other explanations. E.g. "One way to account for this is with an additional bequest motive:"
- Do we want to use a normalized model for this, or use the un-normalized model to show off a more realistic SS benefits structure?
- In the text, risky shar is \varsigma, but it's S in the math.
- The value function definition is on 3 lines but could fit on 1 easily
- The model description is not correct. As written, the agent chooses S_t* every period, but it only actually *does anything* with some probability. In the actual HARK model, the agent *knows* whether they can adjust their risky portfolio share, and makes their consumption decision based on that.
- Using two different p_t symbols to mean "permanent income level" and "adjustment probability" is confusing.
- The Weierstrass p is never defined as a uniform RV on [0,1].
- The descriptive definitions of theta, gamma, and psi are incorrect.
- Let's come up with *some* symbols for BeqFac and BeqShift.
- rho_beq needs to be the same as normal CRRA or things fall apart.
- There's an extra "Details" section that's empty, right below "Mathematical Details" (which is not empty)
- The hump shape of wealth has nothing to do with "the standard portfolio allocation strategy implied by basic economic theory"
- You shouldn't need to run update() right after instantiating
- Note the very small bump in wealth-to-income ratio at retirement; that's because the permanent income drop at retirement is set to like 6%
- Should this describe our modeling assumptions a little more-- what went into the parameterization-- maybe as an optional/hidden box?
- Where is there a -25 in the code?
- Vertical axis for wealth-to-income plots is labeled as "Wealth (USD)"
- I'm really not sure we should use this as the baseline model, because rho is 9.2, and we describe 7 as "medium" risk aversion.
- As you note, we should expand the economic explanation, and maybe make an interactive graph.
- "How does one calibrate the parameters to match individual stated preferences of risk?" I think we'd need to ask preference-eliciting questions. Same thing with bequest motives.

PR #1529, opened 2025-02-16
Unwind init dependencies

mnwhite:
Up to now, each `AgentType` subclass had bespoke initialization and updating/constructing code that was run when it was called or the `update()` method was run. Because of the inheritance tree, the behavior of these methods often wound through several superclasses and was very confusing.

This PR unwinds all of that bespoke code and removes (essentially) all `__init__` and `update` methods from all of our classes. Default parameters and solver (`solve_one_period`) are now set in the class attribute `default_` as a two element dictionary that is read automatically when an AgentType is initialized. I also moved a few more things to constructors that were missed before, but there are still a (very) small number of (extremely minor) exceptions, which I will get to. On top of that, I got rid of all manipulations of `time_inv` and `time_vary` for solver-dependent attributes. If something should be time varying or time invariant and is used by the solver, the class says so up front.

As far as I know, the *only* exception to the "no `__init__` code" mantra is right in PerfForesight, from which everything descends. It calls a method that determines whether `Rfree` is time varying, which was put in at some point when I wasn't paying attention. I really wish this change had been implemented as simply making `Rfree` time-varying and throwing brackets around one number in a few dictionaries, rather than making a ton of really weird workaround code. I'm going to fix that issue, but on a different PR.

As of when I am putting in this PR, all tests pass for me locally *except* econforge_interp tests, and that's probably because I have an outdated version of the package on my desktop. However, I am quite sure that I broke a bunch of examples with these changes, and now I'm going to deal with that, one notebook at a time.

Some of the convenience methods I removed (like `update_income_process()` might get put back in as I work through this.


mnwhite:
Only 8 notebooks have errors, that's pretty good! Will fix on the next commit tomorrow morning, then go back and deal with the cubic and value function issue with the "fast" models.

mnwhite:
Most of the example notebook errors went away when I added two one-line methods: a "synonym" method `update` that duplicates `construct` and a basic `update_income_process` method that calls `construct` on the relevant attributes. Three notebooks just had formatting issues that were fixed easily; the fourth notebook has an *example* of the RiskyAssetLabeled model, even though there are no tests for it!

The commit I just put in will pass all tests except the last cell or two of that one notebook. I'm going to try to repair that model, but it's a bit of a mess with its constructed inputs.

mnwhite:
It was less of a mess than I feared. The main issue was just a swapped order of inputs on one call. The labeled portfolio model solver also has an unexpected need for `RiskyShareFixed` to exist, even though it's not used. This is one of the reasons I want to get away from inheritance in our models.

I'm going to let the tests run (successfully), and then turn to the cubic and vFunc issue with the "fast" solvers. I suspect I did something wrong when simplifying the terminal solution code. I'll then undo the target value changes in those tests (which changed because cubic interpolation was turned off).

mnwhite:
Some ConsAggShock test values have changed slightly because the MaggGrid changed when adding a constructor function.

Once tests run, this PR will be ready for review. I'm not sure we want to merge it for the upcoming HARK 0.16.0 release, but instead merge it after 0.16.1 (the packaging-only release) and hold it for later. But it does need to be merged before other significant work can be done, because a lot of the model files were *rearranged*, even if the code didn't change.

I'm going to do the work to remove `update_Rfree` in another branch off from this one.

examples/ConsPortfolioModel/example_ConsPortfolioModel.ipynb
alanlujan91, line 1: Funky results

examples/ConsumptionSaving/example_ConsRiskyContribModel.ipynb
alanlujan91, line 1: slight differences in graphs

examples/ConsPortfolioModel/example_ConsPortfolioModel.ipynb
mnwhite: Ah yes, this issue is fixed in the small extension to this PR, #1530 ... This arose from some issue with Rfree, I think. If you look at the same notebook on that PR, this figure is back to normal.

examples/ConsumptionSaving/example_ConsRiskyContribModel.ipynb
mnwhite: Yeah, it looks like the M-S limits somehow changed? Let me look at that. On the other figures with even smaller differences, I suspect it's due to a change in the number or spacing of gridpoints.

Some representative agent figures also changed in another place, but the old figures were wrong and the new ones are right.

mnwhite: I looked at the small changes in the policy function figures, and it appears that this change might pre-date this PR. The tests include this exact example (other than `RiskyStd` having a different value), and there were no changes to those results for this PR nor #1530 , so my suspicion is that the figure changed from some prior code change (default parameter change, probably), but the notebook wasn't re-run until this PR.

examples/ConsPortfolioModel/example_ConsPortfolioModel.ipynb
mnwhite: Nevermind, there is an issue that needs fixing.

mnwhite:
@alanlujan91 The "funky results" were resolved by removing the custom dictionary and just using the default one. I think something that was simplified in the main code didn't get copied over to the duplicative "example" dictionary.

The small differences in the M-S limit are due to the RiskyStd changing from 0.2 to 0.1836 in the parameters. I think we want to go with 0.2 as the default to use consistently simple values, but I don't want that to be a blocker for this PR.

mnwhite:
I would *like* this to be in version 0.16.0, but I'm willing to let it fall to 0.17.0. However, it needs to be merged before a bunch of other work can be tackled without generating massive conflicts. Marking as v0.17 for now.

llorracc:
I think we need to have a long session in which we go through the PR's that need to be merged jointly so we all understand what they do and why.  Maybe we should plan to start that (at least) on Wednesday?

alanlujan91 (approved):

PR #1543, opened 2025-03-31
Start solver from arbitrary solution object

Mv77:
Sometimes (e.g. when using the Fake News method to get Jacobians), it is useful to solve a model backwards starting from a solution object that one already has in hand.

For instance, imagine an infinite horizon model. I have already computed the steady state solution `ss_sol`. Now I want to know what the solution would be if I gave the model the sequence of inputs `[R_ss, R_ss, R_ss + dx, R_ss, R_ss, ...]`. I can solve my problem by doing backward induction on a three-period model with inputs `[R_ss, R_ss, R_ss + dx]` that takes `ss_sol` as its continuation value/ terminal solution.

This PR adds exactly this functionality, which I have had in my private branches for a while and has been very useful.

I imagine you can somehow already do this in HARK, manually overriding `solution_terminal`. @wdu9 managed to do this. But I think those manual overrides are hacky and it is also hard to know when `update_solution_terminal()` will be called and re-override the solution. I found this implementation clearer and more convenient.

### DIFF
--- a/Documentation/CHANGELOG.md
+++ b/Documentation/CHANGELOG.md
@@ -36,6 +36,7 @@ Release Date: TBD
 - Expands functionality of Cobb-Douglas aggregator for CRRA utility. [#1363](https://github.com/econ-ark/HARK/pull/1363)
 - Adds a new function for using Tauchen's method to approximate an AR1 process. [#1521](https://github.com/econ-ark/HARK/pull/1521)
 - Adds additional functionality to the CubicHermiteInterp class, imported from scipy.interpolate. [#1020](https://github.com/econ-ark/HARK/pull/1020/)
+- Allows users to pass a generic solution object to agent solvers to be used as the initial condition of backward induction. [#1543](https://github.com/econ-ark/HARK/pull/1543)
 
 ### 0.15.1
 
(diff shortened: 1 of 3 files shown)


mnwhite:
Thanks for this, I'll take a look at it. Yes, the behavior of `pre_solve` and `update_solution_terminal` is somewhat unpredictable as-is. One of my current PRs tries to make it much more predictable by moving *everything* that can be used with the `construct` method into that format, and eliminating *all* bespoke init code (etc). It should then be much easier to do something like:

`ThisType.solution_terminal = custom_solution`
`ThisType.solve(presolve=False)`

And be confident that it will solve exactly as-is, without replacing anything. But it might be that specific functionality for this should also be supported.

Mv77:
Great, thanks for checking it out. I'd be happy to discuss.

Mv77:
If it turns out that this is a functionality that is encompassed by a broader PR that might not come for some time, I would appreciate if it could just be merged in and overridden in the future. It is very self contained and makes calculating jacobians much cleaner.

mnwhite:
This looks good to me. The only thing I would add would be to copy-paste the docstring entry for `from_solution` in `solve_cycle` into the docstring for `solve` as well, so it's in the top-level method that people actually interact with. Otherwise, I think we should merge it; I don't think this bumps into the small-ish changes to `core` that I make in the other PRs.

Mv77:
Good point! The last commit adds the docstring. If/when tests pass, I'll merge this in.

mnwhite:
Thank you!

PR #1544, opened 2025-04-02
Use "from_solution" solves in NKAgent

Mv77:
Just wanted to give an example of how the functionality introduced in https://github.com/econ-ark/HARK/pull/1543 can be helpful. Here I am replacing two small bits of code in @wdu9 's Jacobian code.

This is not ground breaking work and might not even be worth merging, but it just makes the code marginally more readable, I think.

### DIFF
--- a/HARK/ConsumptionSaving/ConsNewKeynesianModel.py
+++ b/HARK/ConsumptionSaving/ConsNewKeynesianModel.py
@@ -560,9 +560,6 @@ class NewKeynesianConsumerType(IndShockConsumerType):
         # Add Rfree to time varying list to be able to introduce time varying interest rates
         FinHorizonAgent.add_to_time_vary("Rfree")
 
-        # Set Terminal Solution as Steady State Solution
-        FinHorizonAgent.solution_terminal = deepcopy(self.solution[0])
-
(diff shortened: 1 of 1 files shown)


mnwhite:
Funny, I was going to suggest putting this in the other PR. I haven't looked at this code yet, but this is the intended use-case for your other PR.

Mv77:
Yes, my reason not to include it is that maybe, just maybe, this is not what @wdu9 intended to _du_ and it passes tests by chance. Just wanted to have the option of not merging this... If @wdu9 has time he could review this tiny change.

But the functionality in general is very useful for general-ish Jacobians code.

llorracc:
@wdu9 it would be great if you could look at this

mnwhite:
@wdu9 We really want you to do this and it will only take about 5 minutes of looking.

wdu9:
> @wdu9 We really want you to do this and it will only take about 5 minutes of looking.

Looks good to me.

wdu9:
@mnwhite

wdu9 (approved):

mnwhite:
I'll merge this after the other two are in.

PR #1548, opened 2025-04-29
[WIP] Class representing discretized transitions

Mv77:
I'm making public versions of a few methods that I have used to work for simulations with discretized/matrix transitions.

The idea is to have a simple class that holds all the ingredients needed to, say, find the steady state distribution of an infinite horizon or life-cycle model, or to iterate a distribution forward a single time period.

Ideally, people will use this and make it more efficient and find useful sparse representations and who knows what else.


Mv77:
This is a work in progress, do not merge yet. I need to, for example, add tests.

mnwhite:
I finally had a chance to look at this. Main comments are related to both prongs of `find_steady_state_dstn`.

For the infinite horizon version, you don't need to loop on `_iterate_dstn_forward_ih` until convergence. You can just combine your single `living_tmat` with a "newborn transition matrix" that has `newborn_dstn` in every row, with weights `surv_prob` and `1-surv_prob`. That's a "master" transition matrix. Then transpose it and use `sp.linalg.eig` to get `k=1` eigenvectors, and normalize it. The largest eigenvector of (the transpose of) a non-singular, non-degenerate Markov matrix is always 1, and the corresponding (normalized) eigenvector is the steady state distribution. For extra speed, cast the transition matrix to `sp.sparse.csr_matrix` and use `sp.sparse.linalg.eigs` instead. Note that this method is used in `HARK.ConsumptionSaving.ConsNewKeynesianModel.NewKeynesianConsumerType.find_steady_state()`, minus the sparse arrays.

For the lifecycle model, your steady state age distribution is wrong. You can't just take the cumulative survival probabilities and normalize them, because that ignores death and replacement. Instead, make a square array A of "age transitions", with two non-zero elements in each row: A_{i,i+1} = LivPrb_i and A_{i,0} = 1-LivPrb_i. Just like above, take the transpose of A, then find the eigenvector associated with its largest eigenvalue and normalize it. That's the steady state age distribution. Your survival-conditional state distributions are correct. See L168-202 of estimation.py in the DistributionOfWealthMPC REMARK for where this was done in cstwMPC, but note that when I wrote the function there, I didn't know that the largest eigenvalue is always 1 and/or `eig` wasn't necessarily sorted back then.

mnwhite:
Addendum to the above: The lifecycle method assumes that you want a *constant* population, with newborns exactly replacing decedents. If you instead want a constant *growth rate* of the population, the math is slightly different (but you can probably work it out).

mnwhite:
Hit pause on my comment about the steady state age distribution. Normalized cumulative survival probabilities and the first normalized eigenvector might be equivalent. I'm on a train right now and can't check.

mnwhite:
Yeah nevermind, the methods are equivalent.

PR #1549, opened 2025-05-04
MAINT: Support python 3.13; remove numba, numpy upper pins

MridulS:

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -1,9 +1,10 @@
 interpolation>=2.2.3
 joblib>=1.2
+legacy-cgi; python_version >= '3.13'
 matplotlib>=3.6
 networkx>=3
-numba<0.60.0
-numpy<=1.26
+numba
+numpy
 optimagic
 pandas>=1.5
 pyyaml>=6.0
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -26,10 +26,10 @@ jobs:
       contents: write
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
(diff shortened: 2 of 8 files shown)


requirements/base.txt
```diff
@@ -1,9 +1,10 @@
 interpolation>=2.2.3
 joblib>=1.2
+legacy-cgi; python_version >= '3.13'
```
MridulS, line 3: interpolation uses a python feature that has been long deprecated and now removed from python 3.13. This package adds legacy support.

mnwhite:
I have a feeling some things that are in HARK but lack tests are going to
break with NumPy 2. I think it's related to treating a singleton array as
if it were a scalar, which raises a warning on NumPy 1.25 but I think will
break on newer versions. I'll look into it.

Is interpolation.py being maintained or updated? What in HARK uses it
rather than our stuff? Is it just the "fast" model?

MridulS:
> I have a feeling some things that are in HARK but lack tests are going to
break with NumPy 2. I think it's related to treating a singleton array as
if it were a scalar, which raises a warning on NumPy 1.25 but I think will
break on newer versions. I'll look into it.

Yeah, but hopefully the fixes should be quick :) Numpy 2 has been out for a while.

> Is interpolation.py being maintained or updated?

I don't see much activity there.

> Is it just the "fast" model?

Yes, I believe so.

mnwhite:
Exactly one test failed on Py 3.13, and it might have been due to a worker timeout. Re-running to see if that fixes it.

mnwhite:
The failing test ran successfully on the second attempt. Merging now after adding changelog entry.

Issue #708, opened 2020-06-03
automate PyPI/conda release using GitHub Actions.


MridulS:
- Setup GitHub Actions to build whenever a tag(release) is associated to a commit.
- (optional?) add the middle step of pushing it to TestPyPI server.
- upload it to PyPI using https://github.com/pypa/gh-action-pypi-publish.
- conda forge is already automated through fetching PyPI release, just a PR needs to be a merged at https://github.com/conda-forge/econ-ark-feedstock

PR #1551, opened 2025-05-07
Support PyPI Trusted Publisher for making releases

MridulS:
Fixes https://github.com/econ-ark/HARK/issues/708

Keeping it in draft mode for now as we still need to turn on trusted publisher on pypi side https://docs.pypi.org/trusted-publishers/adding-a-publisher. I can do that if folks think we should go ahead with this :)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,34 @@
+name: Build Wheel and Release
+on:
+  workflow_dispatch:
+  release:
+    types:
+      - published
+jobs:
+  build-and-inspect-package:
+    name: Build & inspect package.
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: hynek/build-and-inspect-python-package@v2
+
+  publish:
+    name: Publish release to PyPI
+    if: github.repository_owner == 'econ-ark' && github.event_name == 'release' && github.event.action == 'published'
+    runs-on: ubuntu-latest
+    environment: release
+    permissions:
+      id-token: write
+      attestations: write
+      contents: read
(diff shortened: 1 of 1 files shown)


mnwhite:
So the idea is that rather than doing the old process of manually pushing to PyPI and concurrently making a release commit / tagging a version on GitHub, we would just do a version tag here and that would *automatically* propagate to a PyPI release?

MridulS:
> we would just do a version tag here and that would automatically propagate to a PyPI release?

Indeed! This is the recommended security practice too now.

mnwhite:
@llorracc @akshayshanker @alanlujan91 See above. I vote for this.

llorracc:
@MridulS, thanks for the input.  This sounds like a good idea.

mnwhite:
Alan pointed out that this change is part of the (oddly named) (PR #1514)[https://github.com/econ-ark/HARK/pull/1514]. The current plan is that after we release v0.16.0, we will then quickly do v0.16.1 with *only* packaging / technology changes. So let's put this on hold for (hopefully) a short while.

alanlujan91:
"cookie" https://github.com/scientific-python/cookie

mnwhite:
I'm sure it means something, but my point is that the one word title with no verb isn't clear. My understanding/definition of "cookie" is:

1) the round sweetdiscs that we om-nom-nom
2) a file that websites deposit in your webcache

The cookie PR is related to neither, and the word "cookie" doesn't appear in the description anywhere.

MridulS:
@alanlujan91 I was just overwhelmed by looking at all the merge conflicts it will create so trying to move in baby steps :D

Issue #1534, opened 2025-02-25
Incorporate "arrival" variables in simulation


mnwhite:
This issue is reproduced from communications among team members.

--------------------------------------------------------------

MNWHITE writes:

As is, our simulator methods "start" each period with *last* period's so-called post-state variables (information at the very end of the period). Specifically, they take these values from the `state_prev` dictionary and then use them in the simulator method to make things that will go in `state_now`. E.g. the typical consumption-saving model uses `aNrmPrev` and `pLvlPrev` at the very start of a simulated period before using them to get `bNrm` and `pLvl` for the current period.

A side effect of this structure is that newborns initialize with "last period" post-state variables for a split second before they are converted to current period values. This is a little awkward and overall not great.

This work item is to implement in HARK something like the "twist" or "connector" concept that we will have in the StageClass / new structure. That is, an explicit mechanism for relabeling (e.g.) `aNrm` as `kNrm` "between" periods; nothing can "happen" in this interstitial step other than relabeling.

The general principle we want to adhere to is that symbols are never "recycled" within a period: each symbol means exactly one thing, nothing more and nothing less.

-------------------------------------------------------------

LLORRACC writes:

It's not clear to me that we need the "MNW twist" (not a popular new dance) to get started on this.  Or even to finish.

The right way to get started is just to modify the model to give a new name (not `aNrmPrev`) to the incoming state variable.  I think this requires any no coding, and so should be the first step.  

I use the weasel words 'something related to k' because there's an irritating subtlety here.  If the MNW twist says $k_{t+1} = a_{t}$, the problem is that $a_{t}$ is $A_{t}/P_{t}$ and $k_{t+1}$ is therefore $K_{t+1}/P_{t}$, while for every other `{t+1}`- dated variable we want to think of it as being normalized by $P_{t+1}$.  This is particularly awkward in the LaTeX documents because there is no obvious way to notationally signal the point that k is normalized by the _previous period_'s permanent income.  

This is easier in the code, because we can just give the variable a name like `kNrmPrePermShk` to transparently signal the point that when we move from the arrival to the decision stage this variable will evolve according to `bNrm=(kNrmPrePermShk/(PermGroFac*PermShk))*Rfree` and a code comment explaining what is going on will suffice.  

[...]

PR #1553, opened 2025-05-16
Update newborn distribution format

mnwhite:
All of our AgentType subclasses have hardcoded `sim_birth` methods for initializing "newborns" at model birth. Each of those methods relies on particular attributes and forces distributional assumptions onto newborns; this means that a user would need to make a new subclass just to (e.g.) make their newborns have custom distribution of wealth at birth, rather than lognormal.

Consistent with moving the construction of complex model inputs from methods to constructor functions (in the `constructors` attribute), this PR is moving the newborn/initial distributions to constructors as well. The `sim_birth` methods will remain (in the legacy simulation system), but they will just involve making N draws from the appropriate distributions. These distributions can then also be used in my new simulation system in the `initialize` block (see #1545 ). As a bonus feature, I'm *mostly* fixing the "aNrm means two things" problem in the current simulation system.

It will also eliminate custom code for the `reset_rng()` method in AgentType subclasses. Instead, distributions should be named in the `distributions` class attribute, and then `AgentType.reset_rng()` will automatically reset them.

I'm just beginning work on this, so there will be tons and tons of test and notebook failures. Like, a lot.


mnwhite:
This is related to #1534

mnwhite:
All tests are now passing (but I had to change the target on two ConsAggShock tests because they're sensitive to initial conditions). A *lot* of notebooks will need small revisions, and I should do a pass to change references to parameters in the documentation.

mnwhite:
Turns out only two notebooks failed, and it's because those types have no tests, only examples / demos. Everything should work now, but I'm going to look for obvious places where the documentation notebooks should change.

mnwhite:
This should be ready for review and merge once tests finish running. All notebooks that refer to initialization attributes have been updated.

mnwhite:
Copilot suggestions are actually helpful, both on this PR and the others.

HARK/ConsumptionSaving/tests/test_IndShockConsumerType.py
```diff
@@ -634,10 +634,10 @@ def transition(self):
     # Parameters only used in simulation
     "AgentCount": 500,  # Number of agents of this type
     "T_sim": 100,  # Number of periods to simulate
-    "aNrmInitMean": np.log(1.3) - (0.5**2) / 2,  # Mean of log initial assets
-    "aNrmInitStd": 0.5,  # Standard deviation of log initial assets
-    "pLvlInitMean": 0,  # Mean of log initial permanent income
-    "pLvlInitStd": 0,  # Standard deviation of log initial permanent income
+    "kNrmInitMean": np.log(1.3) - (0.5**2) / 2,  # Mean of log initial assets
```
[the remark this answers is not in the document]
mnwhite: Fixed.

HARK/core.py
```diff
@@ -978,17 +979,28 @@ def solve(self, verbose=False, presolve=True, from_solution=None):
...
 
-        Parameters
-        ----------
-        none
-
-        Returns
-        -------
-        none
+        1) The target is a single distribution object
+        2) The target is a list of distribution objects (probably time-varying)
```
alanlujan91, line 986: Don't we have a TimeVaryingDistribution object for this purpose?

alanlujan91 (approved):
looks good to me!

HARK/core.py
mnwhite: We do, but it might not be used for all time-varying distributions, and some of my greatest difficulties have come from managing objects from that class.

alanlujan91 (approved):

PR #1556, opened 2025-05-19
Fix typo in core error message

alanlujan91:
## Summary
- reference `initialize_sim()` in error string

## Testing
- `pip install -e .[dev]` *(fails: Could not find a version that satisfies the requirement setuptools>=61.2)*


mnwhite:
Will merge after tests run, but this just changes an error string.

PR #1561, opened 2025-05-19
Enable parallel simulation for AgentPopulation

alanlujan91:
## Summary
- allow parallel execution in `AgentPopulation.simulate`
- set a flag to warn only once if parallel execution fails
- replace variable docstring with a comment

## Testing
- `python -m pytest -n auto`


mnwhite:
This will only fail if the AgentType instance somehow can't be pickled.

PR #1562, opened 2025-05-20
Improve numpy usage

alanlujan91:
## Summary
- vectorize kernel regression computations
- streamline epanechnikov kernel
- use numpy indexing to get interest rates
- evaluate polynomials with `np.polyval`
- eliminate loops in idiosyncratic-shock consumer functions
- use boolean arrays directly in `core.AgentType`

## Testing
- `ruff check --fix HARK/core.py HARK/ConsumptionSaving/ConsIndShockModel.py`
- `pytest -n auto` *(fails: KeyboardInterrupt)*


mnwhite:
In the second change the AI made in ConsIndShockModel, it adds an unnecessary "if any" test inside the loop it modified. If it's looking at some `t` value, there *must* be some agent with `t_cycle=t` because `t in np.unique(t_cycle)`.

PR #1563, opened 2025-05-20
Clean distance metric warnings, stop simulate() from returning history

mnwhite:
As long as we're doing small pull requests, this PR fixes two things that have been irking me for a while.

1) Someone added warnings to the distance metric sub-functions when they encounter "unusual" cases. Some of those cases really are strange, but others are part of normal HARK solvers. Particularly, when lists or arrays aren't the same length / shape, there's no need to raise a warning. This happens in `PerfForesight` because the terminal criteria is reached (in most cases) when some maximum number of gridpoints is reached (after that many periods of solving). Before that point, each successive solution has 1 more gridpoint than the last one, and the distance metric reports that there is *no way* that these are the same thing.

2) Someone made `simulate()` return the history dictionary, which led to some utterly massive notebook output that we don't want (when `ThisType.simulate()` is in the last line of a cell). I deleted that line.

### DIFF
--- a/HARK/core.py
+++ b/HARK/core.py
@@ -1528,8 +1528,6 @@ class AgentType(Model):
                             )
                 self.t_sim += 1
 
-            return self.history
-
     def clear_history(self):
         """
         Clears the histories of the attributes named in self.track_vars.
--- a/HARK/metric.py
+++ b/HARK/metric.py
@@ -5,61 +5,47 @@ import numpy as np
 
 def distance_lists(list_a, list_b):
     """
-    If both inputs are lists, then the distance between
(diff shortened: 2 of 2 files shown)


mnwhite:
Tests and examples should pass. One notebook relied on `history` being returned rather than retrieved, so it was lightly edited. Ready for review and merge.

mnwhite:
The "low confidence" copilot comments aren't helpful. The docstring already does what it says, and we already have automated testing on Py 3.10-3.13.

alanlujan91 (approved):

mnwhite:
Merging this once tests finish running. Only edit was to changelog.

Issue #1574, opened 2025-06-27
AgentType.construct() needs to check for dependencies


mnwhite:
I'm too tired to write this out fully right now, but there's a bug in AgentType.construct() when there are "layers" of constructed objects and you *change* parameters after instantiation.

Suppose object C depends on object B, which in turn depends on parameter A. If A is changed after instantiation and update() or construct() is run, *and* the constructor dictionary happens to list C before B in its keys, then it will re-construct C *before* B is re-constructed, and thus the change to A won't be properly reflected in the new C.

The quick/dumb workaround is to call construct() twice in a row. The smarter workaround involves either tracking dependencies during the method and doing a second sub-pass if necessary, or building a dependency tree in advance.

mnwhite:
The mostly-works fix for this is to move the call to `self.del_param(key)` to *before* the main construct loop, not inside it: wipe clean anything that's about to be constructed before trying to construct, to prevent this kind of "incomplete update". I'll make a PR for this tonight.

What this won't catch is any "impossible" constructions, such as if A depends on B and B depends on A-- neither can be made without the other already being made first! The `construct()` method will fail gracefully if you use its built-in debugging tools (instantiate your agent with `construct=False` and then call `construct()` after it exists), and the contradiction will be revealed in `_missing_key_data`.

PR #1575, opened 2025-06-29
Move where pre-constructed objects are deleted

mnwhite:
This prevents some "incomplete updates" when running construct() if there are dependencies among constructed objects, fixing #1574 

### DIFF
--- a/HARK/core.py
+++ b/HARK/core.py
@@ -573,6 +573,10 @@ class Model:
         if N_keys == 0:
             return  # Do nothing if there are no constructed objects
 
+        # Delete pre-existing constructed objects, preventing "incomplete" updates
+        for key in keys:
+            self.del_param(key)
+
         # Get the dictionary of constructor errors
         if not hasattr(self, "_constructor_errors"):
             self._constructor_errors = {}
@@ -597,7 +601,6 @@ class Model:
                     constructor = self.constructors[key]
                 except Exception as not_found:
                     errors[key] = "No constructor found for " + str(not_found)
-                    self.del_param(key)
                     if force:
                         continue
                     else:


mnwhite:
Ok bugbot, you got me.

HARK/core.py
```diff
@@ -679,6 +694,9 @@ def construct(self, *args, force=False):
                 if keys_complete[i]:
                     continue
                 msg += " " + keys[i] + ","
+                if keys[i] in backup.keys():
+                    setattr(self, key, backup[key])
+                    self.parameters[key] = backup[key]
```
alanlujan91, line 700: what is this doing?
seems like resetting the backup again, shouldn't this have been done above already?
you know more about this, when can a constructor be None? and if the constructor is None, why assign a back up instead of a default?

mnwhite: No, that's restoring any prior data *from* the backup dictionary to the agent itself.

I think I allowed None to be a constructor to provide a second way to make an attribute non-constructed. If something has a constructor in the default dictionary, and you want to make an instance where it's not constructed (maybe because you're going to build it elsewhere), you can delete the constructor dictionary entry *or* set it to None.

HARK/core.py
```diff
@@ -611,8 +622,12 @@ def construct(self, *args, force=False):
                     else:
                         raise ValueError("No constructor found for " + key) from None
 
-                # If this constructor is None, do nothing and mark it as completed
+                # If this constructor is None, do nothing and mark it as completed;
+                # this includes restoring the previous value if it exists
                 if constructor is None:
+                    if key in backup.keys():
+                        setattr(self, key, backup[key])
+                        self.parameters[key] = backup[key]
```
alanlujan91, line 630: there are 2 places where the back up is restored

alanlujan91: my question is, why is this done twice?

mnwhite: That's only if the constructor is None. The one at the very bottom catches any *incomplete* keys and restores their backup. Incomplete keys happen if there was a constructor error-- usually a failure to find the data.

alanlujan91: Seems like it could lead to unexpected results, failure to find enough data should rais a warning instead of silently failing and giving back the prior construct

mnwhite: That's what it did until I took cursor's advice!

mnwhite: Well, it would throw an error unless you passed `force=True`.

mnwhite: Let me correct that: it *does* raise an error unless `force=True` is passed. I think it works how you want it to:

- if a constructor is missing but it's told to run it, it raises an error unless force=True
- if a constructor is None but it's told to run it, the prior value is used (if any)
- if a constructor is present but it fails to run due to missing data, the prior value is used (if any), and an error is raised unless force=True

Passing force=True is a debugging option. You want to know why your `construct()` call is failing, so you run it in a mode that collects information about its failures and puts it into the `_missing_key_data` dictionary. I meant for it to assign `self._constructor_errors = errors` but I forgot to put that line in.

PR #1595, opened 2025-10-07
Add extensive margin medical care model

mnwhite:
This PR adds a new medical spending model to HARK. The model extends the persistent shocks model (really: general income process model) to include 2D medical shocks: jointly drawn medical cost and utility shock, specified as bivariate lognormal. The agent makes a binary choice of whether to pay the cost or suffer the utility shock, and consumes and saves as normal.

This version of the model has no representation for health as a persistent state; I made it as an initial starting point or proof of concept. It also has no insurance, but I think I left an input for `CoinsRate` that isn't actually used (will fix).

As of right now, the solver works for *most* parameter sets that I've tested, but it can encounter a numeric glitch when I turn up the stdev of utility shocks, This might be due to a sign error on one term, or it might be a result of some extreme numeric math; I'll figure it out.

Simulation in the legacy and new model-file system both work, but the former will have an indexing problem for `pLvlNextFunc` in finite horizon. It's easily fixable, but requires copy-pasting a sim method to remove *two characters*.

Not ready to merge right now, because there are no tests nor example notebook, and the model itself isn't formally stated in the docstring. Test code also needs to be removed from if-main block.


mnwhite:
I should note that the assumption of bivariate lognormal medical costs and utility shocks is "hardcoded" at this time. The code can easily be adapted to have any distribution of medical costs, but the distribution of utility shocks *conditional on med cost* must be lognormal. That distribution has a closed form conditional expectation, allowing much faster computation of the expected utility shock conditional on not buying care.

In principle, the code could be complexified to allow a general distribution of utility shocks conditional on medical cost, as long as it has a method for computing conditional (truncated) expectation.

mnwhite:
ALSO: This PR already includes #1591 because I needed its features. That (much smaller, and ready-for-review) PR should be merged into master *first*, which will make some of the "changes" in this PR go away.

PR #1598, opened 2025-10-17
Simplify lognormal approximation math

mnwhite:
The lognormal discretizer that we've had for ten years had some convoluted math. I've simplified that, and adjusted the imports in that file.

I *think* it's reproducing the same values as before, but I forgot to actually write down old values. Running tests to verify.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,2 @@
+ignore:
+  - "HARK/ConsumptionSaving/LegacyOOsolvers.py"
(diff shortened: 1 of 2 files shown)


mnwhite:
Well this is a weird failure: one notebook fails only on Py 3.13. The only changed method returns the same output (to like 11 digits) as before, but somehow breaks in some Py 3.13 usage.

PR #1603, opened 2025-10-18
New line escapes in python 3.13 for @sj.solved

MridulS:


MridulS:
loool, this is indeed the issue 🤣 

Python 3.13 probably changed the way the string that is returned by the `inspect` function and SJ makes some assumption about new lines. I just moved around the function to a new cell and cleaned up some new lines.

This should fix the failing tests in the open PRs.

PR #1606, opened 2025-10-21
Significantly improve test coverage

mnwhite:
This PR is going to drive up our test coverage number. It's going to be a percentage like nobody's ever seen before. Just incredible.


mnwhite:
I'm going to merge this after tests finish, then do some substantive cleanup on ConsRiskyAssetModel in a different PR.

PR #1607, opened 2025-10-22
Collapse fixed share model into RiskyAsset

mnwhite:
We had two models that were nearly identical: one in which the agents are stuck at RiskyShare=1.0, and another in which they were stuck at some RiskyShare, which may or may not be 1.0. The "fixed share" solver had its name changed to replace the original RiskyAsset solver, which has been deleted. One parameter was added to the AgentType to make it work.

Tests pass for the model file, but need to see if all examples work.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -8,13 +8,16 @@ For more information on HARK, see [our Github organization](https://github.com/e
 
 ## Changes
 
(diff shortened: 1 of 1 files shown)


mnwhite:
@alanlujan91 This is the class reduction we discussed yesterday. I'll let you look at it and merge when ready.

PR #1610, opened 2025-10-28
Expand test coverage even more

mnwhite:
Do not merge. I am creating this PR now so that I can verify tests pass as I add them and delete old unused code.


mnwhite:
No matter what I do, test_simulate.py::testMarkovEvents::test_markov_SSJ crashes the automated tester. It runs locally just fine. I have turned the grid sizes *way* down, and reduced income shock counts as well (because they contribute to memory usage). I also turned the time horizon very far down. Same result every time. I have no idea what's going on.

The test has been commented out, but that means there will be sections of the SSJ code that show up as untested. I'd like to fix this.

mnwhite:
More or less finished all tests that I want to add for this PR.

PR #1617, opened 2025-10-30
Test coverage expansion, part 3

mnwhite:
There are more tests coming, don't worry.


mnwhite:
The only remaining task that I want to include on this PR is to get tests for (more or less) everything in HARK.rewards. That will be... tedious.

Issue #1611, opened 2025-10-28
Duplicate `Multivariate[Log]Normal` definitions


mnwhite:
I forgot about `HARK.distributions.multivariate` and defined very basic multivariate (log)normal distribution classes in `HARK.distributions.continuous` for use by `MedExtMargConsumerType`.

Whatever functionality I put in those classes needs to be rectified with the classes that already existed, and my new versions deleted.

PR #1618, opened 2025-11-03
Delete duplicate multivariate classes, adjust code

mnwhite:
The new MedExtMarg model mistakenly created new, duplicate multivariate (log) normal classes. These have now been deleted, and the model points to the original ones.

The multivariate code was changed slightly because draw() was returning the wrong size output for multivariate continuous distributions.


mnwhite:
This addresses #1611

PR #1626, opened 2025-11-05
Add test for agent with parameters

alanlujan91:
@mnwhite is this what you meant?


mnwhite:
That should do it, yes. It should cover lines 1831-1860 of `HARK.core`

PR #1629, opened 2025-11-17
Expose functions and classes at higher level

mnwhite:
All AgentType and Market subclasses are now aliased at HARK.models and HARK.ConsumptionSaving, to make it easier to import them. Likewise, a few common / core calibration tools are aliased at HARK.Calibration, rather than buried.

I also fixed a discrepancy between the mortality parser and the income parser. This might break examples or DEMARKS, but the fix is trivial. CHANGELOG has a note on this.


alanlujan91 (approved):

mnwhite:
I fixed a formatting discrepancy in one notebook. Will merge after tests run again.

PR #1648, opened 2025-12-03
Fix #1331

### LINKED ISSUE #1331: Distance Metrics
We need to make it much more transparent how we are comparing the distances between our solutions and solution objects. Currently the way to do it is very labyrinthian and to figure it out you have to know how to follow a long rabbit hole. 

We should be able to specify clearly what the distance criteria is and the user should be able to understand how that is happening readily. 

@llorracc

mnwhite:
This PR adds the method `describe_distance()` to `MetricObject`, which recursively generates a description of how the distance metric is computed. Prints output to screen by default, but can return it as a string instead. Has an option to specify maximum depth, because some solution objects are complicated.

I should add some tests so that most of the new code gets covered.

### DIFF
--- a/HARK/metric.py
+++ b/HARK/metric.py
@@ -122,6 +122,9 @@ def describe_metric(thing, n=0, label=None, D=100000):
         A generic object.
     n : int
         Recursive depth of this call.
+    label : str or None
+        Name/label of the thing, which might be a dictionary key, attribute name,
+        list index, etc.
     D : int
         Maximum recursive depth; if n > D, empty output is returned.
 
@@ -185,7 +188,10 @@ def describe_metric(thing, n=0, label=None, D=100000):
             desc += pad * (n + 1) * " " + "SUPPRESSED OUTPUT\n"
         else:
             for key in my_keys:
-                desc += describe_metric(getattr(thing, key), n + 1, label=key, D=D)
+                if hasattr(thing, key):
+                    desc += describe_metric(getattr(thing, key), n + 1, label=key, D=D)
+                else:
+                    desc += key + " (missing): CAN'T COMPARE\n"
 
     else:
         # Something has gone wrong


HARK/metric.py
```diff
@@ -112,6 +112,85 @@ def distance_metric(thing_a, thing_b):
...
+        desc += (
+            "(" + type(thing).__name__ + "): largest distance among these attributes:\n"
+        )
+        if len(my_keys) == 0:
+            desc += pad * (n + 1) * " " + "NO distance_critera SPECIFIED"
+        if n == D:
+            desc += pad * (n + 1) * " " + "SUPPRESSED OUTPUT\n"
+        else:
+            for key in my_keys:
+                desc += describe_metric(getattr(thing, key), n + 1, label=key, D=D)
```
alanlujan91: Typo: `distance_critera` → `distance_criteria`. Also missing trailing `\n` unlike other messages.

HARK/metric.py
```diff
@@ -112,6 +112,85 @@ def distance_metric(thing_a, thing_b):
...
+
+    else:
+        # Something has gone wrong
+        desc += "WARNING: INCOMPARABLE\n"
+
```
alanlujan91, line 199: `getattr()` will raise `AttributeError` if attribute doesn't exist. Consider try/except or `getattr(thing, key, None)` with handling.

HARK/metric.py
```diff
@@ -112,6 +112,85 @@ def distance_metric(thing_a, thing_b):
...
+    thing : object
+        A generic object.
+    n : int
+        Recursive depth of this call.
+    D : int
+        Maximum recursive depth; if n > D, empty output is returned.
+
+    Returns
+    -------
+    desc : str
```
alanlujan91, line 133: Type says `int` but default is `1e10` (float). Should be `int or float`.

HARK/metric.py
```diff
@@ -112,6 +112,85 @@ def distance_metric(thing_a, thing_b):
     return 1000.0
 
 
+def describe_metric(thing, n=0, label=None, D=1e10):
+    """
+    Generate a description of an object's distance metric.
+
+    Parameters
+    ----------
```
alanlujan91, line 120: The `label` parameter is not documented in the docstring.

HARK/metric.py
```diff
@@ -112,6 +112,85 @@ def distance_metric(thing_a, thing_b):
...
+        if n == D:
+            desc += pad * (n + 1) * " " + "SUPPRESSED OUTPUT\n"
+        else:
+            for key in my_keys:
+                desc += describe_metric(thing[key], n + 1, label=key, D=D)
+
+    elif isinstance(thing, MetricObject):
+        my_keys = thing.distance_criteria
+        desc += (
+            "(" + type(thing).__name__ + "): largest distance among these attributes:\n"
```
alanlujan91, line 183: If a key in `distance_criteria` doesn't exist in dict, this raises `KeyError`.

HARK/metric.py
```diff
@@ -112,6 +112,85 @@ def distance_metric(thing_a, thing_b):
...
+                desc += describe_metric(getattr(thing, key), n + 1, label=key, D=D)
+
+    else:
+        # Something has gone wrong
+        desc += "WARNING: INCOMPARABLE\n"
+
+    return desc
+
+
 class MetricObject:
```
alanlujan91, line 203: Consider using `warnings.warn()` for consistency with other warning patterns in this file.

alanlujan91 (commented):
Nice feature addition! A few small items to address before merging.

HARK/metric.py
mnwhite: Fixed

HARK/metric.py
mnwhite: Changed default to 100000. It's just an arbitrary large number.

HARK/metric.py
mnwhite: Good catch, added an exception to flag the missing `distance_criteria` item.

HARK/metric.py
mnwhite: The text "warning" is intentional, because this method is just for generating text feedback for the user in an interactive session. If something about the distance criteria is not working, I want them to be told right in `describe_distance`.

HARK/metric.py
mnwhite: Fixed

HARK/metric.py
mnwhite: Nice catch, fixed

PR #1685, opened 2026-01-04
Fix wheel building and remove warnings

MridulS:
@mnwhite this should fix the release and remove all the warnings in the build logs. We had some long deprecated usage in it :)


mnwhite:
Thank you! I will merge in a few hours, then attempt a new release. Have a
very important engagement this afternoon (read: child's birthday party at
bowling alley).

PR #1701, opened 2026-01-17
Fix array broadcasting in 2D interpolation classes

llorracc:
## Problem

`HARKinterpolator2D` methods (`__call__`, `derivativeX`, `derivativeY`) fail when called with mixed scalar/array inputs:

```python
from HARK.interpolation import LinearInterpOnInterp1D, LinearInterp
import numpy as np

# Create interpolator
y_list = np.array([0.8, 1.0, 1.2])
x_interps = [
    LinearInterp([0, 1, 2, 5], [0, 1, 2, 5]),
    LinearInterp([0, 1, 2, 5], [0, 1, 2, 5]),
    LinearInterp([0, 1, 2, 5], [0, 1, 2, 5]),
]
lioi = LinearInterpOnInterp1D(x_interps, y_list)

# This works:
lioi(np.array([1.0, 2.0]), np.array([1.0, 1.0]))  # ✓

# This fails with IndexError:
lioi(np.array([1.0, 2.0]), 1.0)  # ✗
```

**Error:**
```
IndexError: boolean index did not match indexed array along dimension 0; 
dimension is 2 but corresponding boolean dimension is 1
```

## Root Cause

In `HARKinterpolator2D.__call__`:

```python
xa = np.asarray(x)
ya = np.asarray(y)
return (self._evaluate(xa.flatten(), ya.flatten())).reshape(xa.shape)
```

When `x = [1, 2]` (array) and `y = 1.0` (scalar):
- `xa.flatten()` → `[1, 2]` (length 2)
- `ya.flatten()` → `[1.0]` (length 1)

The `_evaluate` method assumes both inputs have the same length, causing the `IndexError`.

## Fix

Add `np.broadcast_arrays` to ensure inputs have matching shapes before flattening:

```python
xa = np.asarray(x)
ya = np.asarray(y)
xa, ya = np.broadcast_arrays(xa, ya)  # NEW
return (self._evaluate(xa.flatten(), ya.flatten())).reshape(xa.shape)
```

The same fix is applied to `derivativeX` and `derivativeY`.

## Affected Use Cases

This bug affects complex nested interpolator structures commonly created by consumption-saving solvers, such as:

```
LowerEnvelope2D(
    VariableLowerBoundFunc2D(LinearInterpOnInterp1D(...)),
    BilinearInterp(...)
)
```

These structures are evaluated with mixed inputs during simulation (e.g., `cFunc(mNrm_array, Cratio_scalar)`).

## Testing

Added `tests/test_interpolation_broadcasting.py` with comprehensive tests for:
- `LinearInterpOnInterp1D` with (array, scalar)
- `VariableLowerBoundFunc2D` with (array, scalar)
- `LowerEnvelope2D` with nested interpolators
- `derivativeX` and `derivativeY` methods
- Reversed (scalar, array) inputs

## Discovery

This issue was discovered during migration of an external project (HAFiscal) from HARK 0.14.1 to 0.17.0.

### DIFF
--- a/HARK/interpolation.py
+++ b/HARK/interpolation.py
@@ -360,8 +360,9 @@ class HARKinterpolator3D(MetricObject):
         x : np.array or float
             Real values to be evaluated in the interpolated function.
         y : np.array or float
-            Real values to be evaluated in the interpolated function; must be
-            the same size as x.
+            Real values to be evaluated in the interpolated function. If multiple
+            inputs are arrays, they must be broadcastable to the same shape.
+            Scalar inputs will be broadcast to match array inputs.
         z : np.array or float
             Real values to be evaluated in the interpolated function. If multiple
             inputs are arrays, they must be broadcastable to the same shape.


llorracc:
Thanks for your quick review.  

This took me quite a while to run down, but was the core reason that my upgrade from HARK 0.14.1 to 0.17.0 failed.  

If you think our original code (which worked) should not have worked, I'd appreciate your help in rewriting it.  But even if that is true, I guess there is a danger that other bits of working code will break when people upgrade to 0.17.0?

mnwhite:
I just verified that the docstrings for `HARKinterpolator2D`, `HARKinterpolator3D`, and `HARKinterpolator4D` do in fact specify in both `__call__` and derivative methods' docstrings that the input arrays need to be the same shape (sometimes mistakenly using the word "size" instead of "shape", which is misleading because of `np.array.size` and `np.array.shape`).

If we're going to make this change, then:

1) the docstrings should be updated to reflect the new acceptable arguments

2) the change should be made in `HARKinterpolator3D` and `HARKinterpolator4D` as well

3) the tests can just be put in tests/test_interpolation.py, not a new file just for this one specific thing

mnwhite:
Sorry, our messages / posts crossed. Let me look at HARK 0.14.1. I really didn't expect that code to work in *any* version of HARK after about 2018.

llorracc:
OK, I've added the deprecation warnings and docstrings.  You should see the revisions now.

mnwhite:
I just checked the "blame" history for that section of code. The lines `xa = np.asarray(x)` and `ya = np.asarray(y)` literally have not changed since 2015, and the line below that (flattening and reshaping) was only edited ~8 years ago when we first started using to enforce `black` for code style (whitespace, etc).

It looks like three months ago I removed a bunch of kludgy code in HARK.interpolation that would handle the case where the arguments were scalars, but those code sections were wholly unnecessary because the top-level interpolator methods pre-processed in a way that made all calls compatible with the main method code. But each of those kludgy case-handlers *only* looked at whether the first argument was a scalar, not any of the others. So at best, it could *maybe* handle `this_interpolant(5.0, some_array)` but not `this_interpolant(some_array, 5.0)`. I'm surprised it *ever* worked.

mnwhite:
I don't think we need any warning to deprecation notice to be raised. If we allow this input format, and the docstrings are updated to say that (which they now are), then it's fine.

llorracc:
OK, thanks for your help.  I'll do that.

alanlujan91 (approved):

PR #1704, opened 2026-01-19
Fixes to webpage example documentation

mnwhite:
Some model notebooks were not rendering on the website because they were not correctly indexed. This PR fixes that and tweaks the links and formatting in the "model directory" notebook.


mnwhite:
The previously unindexed notebooks have a bunch of small issues, which I'm working through.

mnwhite:
Alan ended up fixing the offending notebook by "roundtripping" it to markdown and back. On net, this appears to have just added $$'s around the equation environments (and deleted results cells). Currently running checks after restoring results cells, should be ready to review if tests pass.

PR #1708, opened 2026-01-21
Revamp PrefShockModel

mnwhite:
This PR is the companion to #1706 , revising the utility format in ConsPrefShockModel.py analogously. Roughly, it moves the preference shock from as a factor outside the CRRA term to a divisor inside the CRRA term. This has the effect of making the distribution of shocks be less sensitive to CRRA if you want to maintain the same variation in consumption. It's just an algebraic rearrangement.

The only commit so far fully converts `PrefShockConsumerType`. Because the default CRRA is 2.0, the math works out that the solution is exactly the same in the tests, which is nice validation.

On the next commit (tonight or tomorrow morning), I will revise `KinkyPrefConsumerType`, which will be very easy because I can follow the template of changes from `PrefShockConsumerType`. The documentation still needs a little bit of revision as well.


mnwhite:
This is now complete and ready for review.

PR #1761, opened 2026-04-14
A bunch of mop-up test cases

mnwhite:
Most of these cover only a few lines, but they get us closer to 100% coverage.

A couple more additions are coming on the next commits.


mnwhite:
An additional test will be added to the bequest models branch, but this is ready to review @alanlujan91

alanlujan91 (approved):

PR #1777, opened 2026-05-20
Add HARK.interpolation_jax — opt-in JAX counterparts to interpolation primitives

llorracc:
> ⚠️ **CI note:** the one failing job, `build (ubuntu-latest, 3.12)`, is a **pre-existing breakage on unmodified `main`** — scipy 1.18.0's spline objects cannot be `deepcopy`-ed, which detonates in the existing test suite wherever that scipy resolves. Root cause, three-line repro, and the one-line fix: **#1788** (merge that first). Every other job here is green; once #1788 merges I will refresh this branch so the matrix shows fully green.

> **Non-disruption note (2026-08-10):** this PR adds a new module and its tests only — no existing code file's behavior changes (a changelog entry is the sole edit to an existing file). Now marked ready for review as part of a short series of self-contained upstreaming PRs (see also #1783, #1784).
>
> - [x] Tests added for new functionality (18 cases)
> - [x] Documented (numpydoc docstrings)
> - [x] CHANGELOG.md updated

## Summary

Adds a new opt-in module, `HARK.interpolation_jax`, that ports the interpolation primitives used by `ConsAggShockModel`-style solvers to JAX. The new module is functional (data + query arguments, no classes) so the functions can be `vmap`-ed, `jit`-ed, and differentiated — the building blocks needed for a GPU EGM solver in a follow-up PR.

## Why

HAFiscal (Carroll et al.) has a working JAX EGM solver kernel for HARK's `solve_agg_cons_markov_alt` that validates to <1e-3 vs HARK at recession scale. The interpolation primitives in that kernel have lived as `HAFiscal/jax_hark_interp.py`. This PR upstreams them in HARK-quality shape so any downstream project can reuse them.

Sets the foundation for:
- **PR-2**: `ConsIndShockModelJAX` — JAX EGM for the basic single-state consumer
- **PR-3**: `ConsAggShockModelJAX` — JAX EGM for the Markov + aggregate-state variant (matches HAFiscal's solver structure)

## What this PR adds

New module **`HARK/interpolation_jax.py`** (192 lines) — JAX counterparts to:

| JAX function | HARK reference |
|---|---|
| ` linear_interp_1d ` | `LinearInterp` |
| ` bilinear_interp ` | `BilinearInterp` |
| ` bilinear_interp_derX ` | `BilinearInterp.derivativeX` |
| ` linear_interp_on_interp_1d_shared_xgrid ` | `LinearInterpOnInterp1D` (shared x) |
[...]

PR #1778, opened 2026-05-20
Add ConsIndShockModelJAX — JAX EGM for basic ConsIndShockModel

llorracc:
## Summary

Adds `ConsIndShockModelJAX`, a JAX-accelerated solver for the basic `ConsIndShockModel` (CRRA utility, idiosyncratic permanent + transitory income shocks, single Markov state). Drop-in opt-in alternative to `ConsIndShockModelFast` (numba), using the same wrapping pattern.

## Depends on

PR #1777 (`Add HARK.interpolation_jax …`) — this PR uses `linear_interp_1d` from `HARK.interpolation_jax`. Base branch is set accordingly; will rebase / retarget to `main` once PR-1 merges.

## What this PR adds

New module **`HARK/ConsumptionSaving/ConsIndShockModelJAX.py`**:
- `_egm_kernel` — JIT-ed pure-JAX EGM step
- `solve_one_period_ConsIndShock_jax` — drop-in for `solve_one_period_ConsIndShock`
- `IndShockConsumerTypeJAX` — drop-in subclass of `IndShockConsumerType` that overrides only the `solver` entry in `default_`

```python
# Default (unchanged)
from HARK.ConsumptionSaving.ConsIndShockModel import IndShockConsumerType
agent = IndShockConsumerType(**params)
agent.solve()  # numpy/numba solver

# Opt-in JAX
from HARK.ConsumptionSaving.ConsIndShockModelJAX import IndShockConsumerTypeJAX
agent = IndShockConsumerTypeJAX(**params)
agent.solve()  # JAX solver
```

New tests **`tests/test_ConsIndShockModelJAX.py`** (6 tests, all pass):
- 4 numerical parity tests vs `IndShockConsumerType` (rtol=1e-4)
- 2 `NotImplementedError` tests for unsupported flags

## Scope limitations (raise NotImplementedError)

- `CubicBool=True` — linear cFunc only for now
- `vFuncBool=True` — no value function for now

Both flags are common in standard HARK usage but neither is needed for HAFiscal's core code path. Both will be ported in a follow-up; current scope covers the primary use case (single-state, linear cFunc, no v).

## Numerical precision

The JAX kernel evaluates next-period `cFunc` by lifting it to a 10000-point linear tabulation each iteration (because `cFunc` is typically a `LowerEnvelope` of `LinearInterp`s and not directly representable as JAX-friendly arrays). Single-period interp error is ~1e-7; compounded through ~100 EGM iterations to steady state, this gives ~1e-5 relative agreement — tighter than EGM truncation noise on `aXtraGrid`, loose enough to avoid FP-non-associativity false alarms.

Tests use `rtol=1e-4` to leave headroom.

## Risk

Zero for existing users. The module adds new files; no existing imports change. Users who don't `import HARK.ConsumptionSaving.ConsIndShockModelJAX` see no difference.

## Performance note

This PR is about API correctness, not raw speed. The basic single-state model is small enough that JAX's JIT compilation overhead can dominate per-call wall time on first solve. Real GPU benefits appear in PR-3 (`ConsAggShockModelJAX`) where the per-iter work scales with `StateCount × aCount × ShockCount` and the GPU's parallelism pays off.

## Test plan

- [x] All 6 tests pass locally (Python 3.11, JAX 0.4.x, x64 mode)
- [ ] CI parity tests pass
- [ ] Tests skip cleanly on CI without JAX

### DIFF
--- a/HARK/ConsumptionSaving/ConsIndShockModelJAX.py
+++ b/HARK/ConsumptionSaving/ConsIndShockModelJAX.py
@@ -1,0 +1,292 @@
+"""
+JAX-accelerated solver for the basic ``ConsIndShockModel`` (CRRA utility,
+idiosyncratic permanent and transitory income shocks, single Markov state,
+no aggregate state).
+
+This module is the JAX counterpart to ``ConsIndShockModelFast`` (numba opt-in)
+and follows the same wrapping pattern: a drop-in subclass of
+``IndShockConsumerType`` that swaps in a JAX-based ``solve_one_period``.
+
+JAX is an optional dependency. Importing this module without JAX installed
+raises ``ImportError`` with installation instructions.
+
+Currently supports the basic case: ``CubicBool=False`` and ``vFuncBool=False``.
+Cubic interpolation and value-function computation are not yet ported and
+raise ``NotImplementedError`` when requested.
+
+The interpolation primitives used here live in ``HARK.interpolation_jax``.
+"""
+from __future__ import annotations
+
+try:
+    import jax
+    import jax.numpy as jnp
+except ImportError as _e:  # pragma: no cover
+    raise ImportError(
+        "HARK.ConsumptionSaving.ConsIndShockModelJAX requires JAX. "
+        "Install with: pip install jax  (or 'jax[cuda12]' for GPU)"
+    ) from _e
+
+import numpy as np
+
+from HARK.ConsumptionSaving.ConsIndShockModel import (
+    ConsumerSolution,
+    IndShockConsumerType,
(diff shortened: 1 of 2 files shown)


llorracc:
## GPU verification

All 6 parity tests pass on NVIDIA RTX 4080 (CUDA 12.6) as well as CPU. Setup:

```
>>> import jax
>>> jax.devices()
[CudaDevice(id=0)]
>>> jax.default_backend()
'gpu'
```

The default JAX backend (`'cpu'` or `'gpu'`) determines where the kernel runs — no code change needed in this PR. For users who want CPU only despite having a GPU, set `JAX_PLATFORM_NAME=cpu` in the environment.

PR #1779, opened 2026-05-20
[DRAFT] Add ConsAggShockModelJAX — JAX EGM for AggShockMarkov + aggregate state

llorracc:
## Status: Draft

Opened as draft for early visibility. Both parity tests pass, but the lift-resolution / Mgrid-density design choices deserve a review pass before marking ready-for-review.

## Summary

Adds `ConsAggShockModelJAX`, a JAX-accelerated solver for `AggShockMarkovConsumerType` — consumers facing idiosyncratic + aggregate shocks under a discrete Markov macro state. Drop-in opt-in alternative to `solve_ConsAggMarkov`, following the same wrapping pattern as PR-2.

## Depends on

- PR #1777 (`HARK.interpolation_jax`) — uses `bilinear_interp`
- PR #1778 (`ConsIndShockModelJAX`) — base branch; same opt-in pattern

Will rebase / retarget to `main` after both predecessors merge.

## Architecture

The two outer loops (per next-state `j`, per current-state `i`) stay in plain Python — they're the structure that returns HARK's `ConsumerSolution` shape exactly. The inner expectation-over-shocks step is JIT-compiled (`_expected_vp_next_2d`) and called from inside the outer loops. Mirrors the architecture of HAFiscal's existing JAX solver kernel.

## Tests

`tests/ConsumptionSaving/test_ConsAggShockModelJAX.py` — both pass:
- `test_cfunc_state0_at_MSS` — first Markov state cFunc at MSS query point
- `test_cfunc_grid_per_state` — per-state cFunc on a (m, M) grid (rtol=1e-3)

## Numerical precision

The inner kernel evaluates `vPfuncNext[j]` via a 512×128 bilinear table (lift). Two subtleties hardened:

1. **NaN propagation in bilinear interp.** Bilinear at any query inside a cell whose 4 corners include NaN returns NaN (since `0 * NaN = NaN` in IEEE FP). `vPfuncNext` returns NaN for `m < mNrmMinNext(M)`, so the lift's `m_lift_min` must sit at or above `max(mNrmMinNext(M))` over the M-lift range — otherwise the bottom row of the table is NaN and poisons queries in the adjacent row. Caught only by a careful debug pass; the earlier all-corners-of-(R,a,PST,wEff*tran) bound silently extended below the natural borrowing constraint.

2. **Resolution.** 256×64 lift gave ~5e-4 relative error (just over the 1e-3 test tolerance). Bumping to 512×128 gives well within tolerance with negligible memory/time cost. Future optimization could use jax.pure_callback to skip the lift entirely.

## What this PR is NOT

- **Not** changes to existing modules — opt-in only.
- **Not** a new core dependency — JAX import is gated.
- **Not** the full Krusell-Smith solver — `solve_KrusellSmith` would be a separate PR.
- **Not** GPU-required — works on CPU; GPU is a config flag at the JAX level.

## What needs reviewer judgment

- **Lift resolution as default.** 512×128 closes the test tolerance but is arbitrary. Larger may be needed for problems with very tight constraint structures; the `_DEFAULT_M_LIFT` / `_DEFAULT_AGG_LIFT` constants are easily tunable.
- **Pure-Python outer loop trade-off.** I kept the per-state loops in Python for transparency and easy comparison to `solve_ConsAggMarkov`. A full-JAX outer loop (vmap over states) would be faster but harder to read; happy to add if reviewers prefer.

## Test plan

- [x] Both parity tests pass locally (Python 3.11, JAX 0.4.x, x64)
- [ ] CI parity tests pass
- [ ] Tests skip cleanly on CI without JAX
- [ ] Benchmark numbers (will add as PR comment)

### DIFF
--- a/HARK/ConsumptionSaving/ConsAggShockModelJAX.py
+++ b/HARK/ConsumptionSaving/ConsAggShockModelJAX.py
@@ -1,0 +1,392 @@
+"""
+JAX-accelerated solver for ``AggShockMarkovConsumerType`` — consumers facing
+idiosyncratic and aggregate shocks plus a discrete Markov macro state.
+
+This is the JAX counterpart to ``solve_ConsAggMarkov`` in
+``HARK.ConsumptionSaving.ConsAggShockModel``. It follows the same drop-in
+opt-in pattern as ``ConsIndShockModelJAX`` (PR-2).
+
+JAX is an optional dependency. Importing this module without JAX installed
+raises ``ImportError`` with installation instructions.
+
+Scope of the JAX kernel
+-----------------------
+The two outer loops (per next-state ``j``, per current-state ``i``) stay in
+plain Python/numpy. Each iteration's "expectation over shocks" is what
+benefits most from JAX vectorization, so that step is JIT-compiled and
+called from inside the outer loops. The cFunc and vPfunc objects returned
+are built using HARK's existing numpy interpolator classes, preserving
+compatibility with downstream code that walks ``solution.cFunc.functions``.
+
+This matches the architectural pattern of ``ConsIndShockModelFast``
+(numba) and of HAFiscal's existing JAX kernel for ``solve_agg_cons_markov_alt``.
+"""
+
+from __future__ import annotations
+
+try:
+    import jax
+    import jax.numpy as jnp
+except ImportError as _e:  # pragma: no cover
+    raise ImportError(
+        "HARK.ConsumptionSaving.ConsAggShockModelJAX requires JAX. "
+        "Install with: pip install jax  (or 'jax[cuda12]' for GPU)"
+    ) from _e
+
+import numpy as np
+
+from HARK.ConsumptionSaving.ConsAggShockModel import AggShockMarkovConsumerType
+from HARK.interpolation import (
+    BilinearInterp,
+    ConstantFunction,
+    LinearInterp,
+    LinearInterpOnInterp1D,
+    LowerEnvelope2D,
+    MargValueFuncCRRA,
+    UpperEnvelope,
+    VariableLowerBoundFunc2D,
+)
+from HARK.interpolation_jax import bilinear_interp
+
+__all__ = [
+    "solve_ConsAggMarkov_jax",
+    "AggShockMarkovConsumerTypeJAX",
+]
+
+
+# ============================================================
+# Helpers — lift HARK 2D vPfuncNext to a JAX-evaluatable table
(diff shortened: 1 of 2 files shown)


llorracc:
## GPU verification

Both parity tests pass on NVIDIA RTX 4080 (CUDA 12.6) as well as CPU. The inner kernel (`_expected_vp_next_2d`) does substantial vectorizable work per call (Mcount × aCount × ShkCount lookups via bilinear interp, summed over shock axis), which is exactly the shape that benefits from GPU parallelism.

A full speedup benchmark vs the numpy `solve_ConsAggMarkov` requires solving a converged AggShockMarkov problem end-to-end — out of scope for the parity tests but happy to add as a follow-up benchmark script if useful.

llorracc:
## End-to-end solve speedup vs numpy

Standard `CobbDouglasMarkovEconomy` test fixture (2 Markov states, default grids, AgentCount=1000). Wall time of `.solve()` only — agent construction not included.

| Backend | numpy cold | JAX cold (with JIT) | JAX warm | Speedup (warm) |
|---------|-----------:|--------------------:|---------:|---------------:|
| CPU     |     17.65s |               4.15s |    3.88s |          4.55x |
| GPU     |     17.87s |               5.33s |    4.28s |          4.18x |

(Linux x86_64, NVIDIA RTX 4080, Python 3.11, JAX 0.4.x, x64 mode, `OMP_NUM_THREADS=1`.)

### Observations

1. **JAX wins ~4× even on this small problem.** The pure-JAX expectation kernel is meaningfully faster than HARK's per-Markov-state numpy/numba loop, even before any GPU benefit.

2. **CPU is slightly faster than GPU at this scale.** Per-call dispatch + GPU sync overhead is ~1ms, which is non-trivial when each individual EGM step is ~milliseconds. GPU should pull ahead at problem sizes where the per-call vectorized work is much larger (more Markov states, finer aXtra/Mgrid, longer T).

3. **JIT compile cost is ~0.3s** (warm vs cold difference). That's a one-time cost per session; amortizes immediately for any iterative solve.

### What would help GPU

The bilinear lift currently moves data CPU↔GPU each iteration (`vp_table` is rebuilt in numpy then transferred). Moving the lift to JAX itself (`vmap`-ed evaluation of the vPfunc on the lift grid as a single JIT'd operation) would keep everything on-device. Left as a follow-up — the current 4× CPU speedup is already meaningful for the basic-AggShockMarkov use case this PR targets.

PR #1783, opened 2026-08-10
Add HARK.dual_measure: opt-in Harmenberg neutral-measure parallel tracking

llorracc:

This PR adds a single new module (plus its test suite) and **modifies no existing code file** (changelog entry only) — and composing the mixin without enabling it is covered by a bit-identity test, so default behavior is provably unchanged.

`DualMeasureMixin` composes with `IndShockConsumerType`-family classes and, when explicitly enabled via `setup_Q_measure()`, runs a parallel Q-measure (Harmenberg 2021 neutral-measure) state update reusing the P-pipeline's mortality draws, Markov transitions, and base uniforms. Aggregates computed under Q need no pLvl weighting, which removes the permanent-income contribution to sampling noise (3–17× variance reduction on aggregate consumption estimates in our downstream use). Standalone helpers `compute_mean_pLvl` / `compute_pLvl_factor` provide the analytical E[pLvl] anchor and its path deviation under unemployment shocks.

Design note: the mixin replicates `AgentType.sim_one_period` internally (the Q-step must run before time advancement) and exposes a no-op, forward-compatible `post_state_hook()` extension point between `get_states` and `get_controls` — no `HARK.core` changes are required or made.

Tests: Q-reweighting math; CDF inversion; a dual-mode smoke checked against the Harmenberg identity; a closed-form E[pLvl] case; and a **default-path invariance test** (mixin composed but not enabled ⇒ histories bit-identical to the plain agent).

This is the first of a short series of self-contained PRs upstreaming general-purpose machinery developed for the HAFiscal project; happy to open an umbrella issue describing the series if that would be useful.

- [x] Documented (numpydoc docstrings + module usage example)
- [x] CHANGELOG.md updated


alanlujan91:
Chris, I ran a multi-agent review over this. Three issues look like blockers, and there is a fourth point about merge order that is really a question for you rather than a defect.

## 0. Merge order: this PR is inert until #1799 lands, and nothing on the PR says so

I initially wrote this up as "the coupling mechanism is never implemented." That was wrong, and I want to correct it before anything else. `_base_shock_draws` is populated by #1799, "Wire the dual-measure base-draw cache through IndShock/Markov get_shocks", and #1799's body declares the merge order explicitly. The mechanism exists.

What I want to flag is the position rather than the code. #1799 is the third entry in stack #1804 (#1789 -> #1792 -> #1799), while this PR sits on `main` with no stack edge at all. So the producer is buried at the top of a three-deep stack and the consumer is loose on main, which means the natural merge order, small standalone PRs first, lands this one before the thing that feeds it. Merged that way it is green and inert:

```python
base_draws_dict = getattr(self, "_base_shock_draws", {})
```

returns `{}`, both `_draw_Q_shocks_*` functions take the independent `IncShkDstnQ.draw_events(N)` branch, and `_cdf_invert` is unreachable. `getattr` with a default degrades rather than raising, so CI, the merge box, and the stack map all stay quiet. I take it this is the graceful degradation you intend, and I have no objection to the design.

The part I would still change is the documentation. The module docstring, the class docstring, and the PR description all state the P/Q common-random-numbers coupling as something the code does. Between this PR merging and #1799 merging, that describes behaviour which is not happening, and the merge-order note lives in #1799's body, so whoever merges this one sees nothing. Either a one-line "requires #1799" in this PR's body and a conditional in the docstrings, or hold this until #1799 is ready and merge them together.

The tests do measure P/Q correlation of 0.98 with the wiring absent, but that comes from `make_Q_measure_dstn` cloning `dstn.seed`, so the two streams start in lockstep. It is a side channel rather than the documented mechanism, and item 1 shows it does not survive.

## 1. The accidental coupling decays after the first `initialize_sim()`

`AgentType.reset_rng` resets only the distributions named in the class attribute `distributions`, which for `IndShockConsumerType` is `["IncShkDstn", "PermShkDstn", "TranShkDstn", "kNrmInitDstn", "pLvlInitDstn"]`. `IncShkDstn_Q` is not in that list and `setup_Q_measure` does not add it, so each `initialize_sim()` rewinds P while Q keeps advancing.

Measured on a 500 agent, 8 period run, calling `initialize_sim(); simulate()` three times on the same object:

```
run1  corr(P PermShk, Q PermShk):  0.982
run2  corr(P PermShk, Q PermShk): -0.056
run3  corr(P PermShk, Q PermShk): -0.017
```

`history["cNrm"]` is bit identical across runs; `history_Q["cNrm"]` is not. So Q is not reproducible from the seed, and any estimation loop or repeated `initialize_sim()` silently loses the variance reduction. Adding `"IncShkDstn_Q"` to `self.distributions` in `setup_Q_measure` would fix it.

## 2. Enabling dual mode corrupts the P-measure history

The `simulate()` override reimplements the base recording loop but drops the final `else` branch that `AgentType.simulate` has:

```python
else:
    self.history[var_name][self.t_sim, :] = getattr(self, var_name)
```

Any tracked variable that is a plain attribute rather than a key in `state_now` / `shocks` / `controls` is therefore never recorded once `dual_measure=True`. `MPCnow` is the obvious case, since `ConsIndShockModel` sets `self.MPCnow` as an attribute and it is routinely tracked.

Measured with `track_vars = ["cNrm", "MPCnow", "who_dies"]`:

```
dual_measure=False   history["MPCnow"] all NaN = False
dual_measure=True    history["MPCnow"] all NaN = True
```

This is the one that worries me most, because it contradicts the PR's premise that the P pipeline is untouched. Calling `super().simulate()` for the P recording, rather than copying the loop, would avoid the whole class of problem.

## 3. `history_Q` records uninitialized memory for `bNrm` and `kNrm`

`_lag_Q_states` blanks every entry of `state_now_Q` with `np.empty(self.AgentCount)`, which is uninitialized rather than NaN. The Q pipeline writes `pLvl`, `mNrm`, `aNrm`, and conditionally `aLvl`, but never `kNrm` or `bNrm`, and `simulate()` then records them verbatim.

Measured with `track_vars = ["cNrm", "pLvl", "bNrm", "kNrm"]`:

```
history_Q[bNrm]  finite=False  min=0  max=1.635
history_Q[kNrm]  finite=False  min=0  max=1.181
```

Those ranges look entirely reasonable for `bNrm`/`kNrm`, which is what makes it dangerous. `_transition_Q` even computes `bNrmNow` and discards it. Filling with `np.nan` instead of `np.empty` would make any future omission fail loudly.

## Also worth fixing

**Markov newborns.** In `_draw_Q_shocks_markov` the comment says "deterministic PermShk = PermGroFac, TranShk = 1 (same as P)", but `MarkovConsumerType.get_shocks` draws newborn permanent shocks randomly from `IncShkDstn[0][j]`. The Q side hardcodes psi = 1, so Q newborns get zero permanent shock dispersion, which biases `mNrm_Q` through `ReffNow`. Separately, `TranShkQ[newborn] = 1.0` is unconditional here while the IndShock branch gates it on `NewbornTransShk`.

**Period indexing.** `t_key = t - 1 if self.cycles == 1 else t` diverges from the base, which uses `t = s - 1` unconditionally, and from `_draw_Q_shocks_markov`, which uses `[t-1]`. For a `cycles=0`, `T_cycle>1` model, P and Q read different periods. Not reachable at the tested calibration since `T_cycle == 1` there.

**Silent degradation.** `make_Q_measure_dstn` returns the input unchanged when `E_perm <= 0 or np.std(perm_atoms) < 1e-12`, and `setup_Q_measure` then sets `dual_measure = True` regardless. A negative or zero mean permanent shock is a modeling error rather than a benign edge case and deserves a warning.

**Unicode em-dashes**, 1 in `dual_measure.py` and 5 in `test_dual_measure.py`. Project rule is ASCII hyphens.

## The tests cannot see any of this

`test_dual_mode_smoke_and_harmenberg_identity` asserts `0.9 < ratio < 1.1`. That ratio computes to 1.0012 with the Harmenberg reweighting removed entirely, so the test passes whether or not the feature works. `test_default_off_is_bit_identical_to_plain_agent` only exercises the disabled path and tracks only `cNrm`/`pLvl`, which is why item 2 got through.

The tests that would have caught these: assert the Q draws are a deterministic function of the P draws, assert P-history equivalence with `dual_measure=True` over a `track_vars` list containing an attribute-backed variable, and assert `history_Q` finiteness for every tracked state rather than just `cNrm`.

The default-off bit-identity tests are the right idea and well built. The gap is that nothing tests whether the enabled behavior is correct.

alanlujan91:
Chris, the three blocking items from my earlier review are fixed and pushed as `82c67cc6`. Two of them turned out to be worse than I first measured.

## Enabling dual mode was dropping tracked variables

The `simulate()` override reimplemented the base recording loop and dropped its final `else` branch, so any tracked variable that is a plain attribute rather than a key in `state_now`/`shocks`/`controls` was never recorded. Measured, 400 agents, T=15, `track_vars = ["cNrm","MPCnow","who_dies"]`:

| config | variable | all-NaN | n_nan |
|---|---|---|---|
| dual OFF | MPCnow | False | 0 |
| dual ON, before | **MPCnow** | **True** | **6000** |
| dual ON, after | MPCnow | False | 0 |
| dual OFF | who_dies | False | 400 |
| dual ON, before | who_dies | False | **800** |
| dual ON, after | who_dies | False | 400 |

`who_dies` was a second casualty I had not flagged: the base falls into the same `else` for `t_sim <= 1`, so rows 0 and 1 were lost as well.

The fix deletes the copied P loop rather than repairing it: `simulate()` now calls `super().simulate(1)` once per period and appends only the Q recording, so it cannot drift from `AgentType.simulate` again, which matters because this is the second time that copy has diverged. Checked the MRO first: no `AgentType` subclass in `HARK/` overrides `simulate`, so per-period delegation is behaviourally identical. Three of the base's argument-validation raises are kept, since looping `simulate(1)` would otherwise turn `sim_periods > T_sim` into an `IndexError` instead of the base's message.

## `history_Q` was recording uninitialized memory

Worse than the earlier numbers suggested. Pre-fix, tracking the full state vector:

| variable | min | max |
|---|---|---|
| kNrm | 0 | **1.13073e+277** |
| bNrm | 0 | 1.46307 |

`history_Q['bNrm'][2, :4]` was `[6.22e-310, 2.81e-315, 0, 0]` - subnormals sitting next to plausible values. Definitional violations pre-fix: `mNrm - (bNrm + TranShk)` = 1.445, and `bNrm*PermShk - Rfree*kNrm` = 1.165e+277. Post-fix those are 2.2e-16 and 5.6e-17.

The clinching detail: the maximum differed run to run on the *same seed* (1.635, 2.73, 1.13e+277 across three runs). That nondeterminism is itself the proof it was reading uninitialized memory rather than a computed value.

Both halves are fixed. `_transition_Q` now stores `kNrm` and `bNrm` - it was computing `bNrmNow` and discarding it - and `_lag_Q_states` blanks with `np.nan` instead of `np.empty`, so a future omission fails loudly. Same pattern as [#1809](https://github.com/econ-ark/HARK/issues/1809), referenced in the code comment.

## The coupling decay is already fixed, by #1799

Re-measured on the merged tree rather than assumed:

| run | corr(P PermShk, Q PermShk) | P cNrm reproducible | Q cNrm reproducible |
|---|---|---|---|
| 0 | 0.974692 | True | True |
| 1 | 0.974692 | True | True |
| 2 | 0.974692 | True | True |

`IncShkDstn_Q` is in `self.distributions` and `_cache_base_shock_draws` is armed, both from #1799. The old 0.982 / -0.056 / -0.017 sequence is gone. The correlation sits at 0.97 rather than 1.0 by design, since P and Q invert the same uniforms through different CDFs. No change was needed here.

## The smoke test could not fail, and now proves its own power

`test_dual_mode_smoke_and_harmenberg_identity` asserted `0.9 < ratio < 1.1`, and that passes with the Harmenberg reweighting deleted entirely:

| variant | old ratio | old passes | new statistic | analytic target | new abs error | new passes |
|---|---|---|---|---|---|---|
| reweighting ON | 0.999416 | True | 1.009336 | 1.009359 | 2.3e-05 | True |
| reweighting DELETED | 1.001207 | **True** | 1.000000 | 1.009359 | 9.4e-03 | **False** |

The old assertion cannot separate them because in this calibration `cNrm` is nearly independent of `pLvl`, so the weighted and unweighted means coincide. The replacement checks the realized Q permanent-shock mean against the analytic `E[psi^2]/E[psi]^2` at tolerance 1e-3, sized from 10 seeds (max sampling error 2.8e-4, so 3.6x headroom and 9.4x below the sabotaged value). A `_NoReweight` control runs inside the test itself and must fail the same assertion, so the test demonstrates its own power on every CI run rather than only when someone checks.

No variance-reduction test shipped. One was built and measured: with `E[pLvl]` estimated from the same simulation, the P/Q variance ratio came out 0.92-1.01x, i.e. no measurable reduction. Demonstrating Harmenberg's 3-17x needs an analytic `E[pLvl]`, a stationary calibration and a per-period cross-sectional comparison, which is a research exercise; a flaky variance assertion would be worse than none.

## Rejection tests

Every fix was confirmed by reverting only that hunk:

| reverted | failures |
|---|---|
| `simulate` delegation | 1 |
| `kNrm`/`bNrm` writes | 2 |
| `np.nan` back to `np.empty` | 1 |
| Markov newborn redraw | 1 |
| both warnings | 2 |

Worth noting row 3: with only `np.empty` restored and the writes intact, the transition-identity test still passes. That is the right separation - the NaN blank is defence in depth, the missing writes were the substantive bug - and it shows the two tests are not redundant.

## Also fixed

**Markov newborns.** `MarkovConsumerType.get_shocks` redraws newborn permanent shocks from `IncShkDstn[0][j]`, while the Q branch hardcoded psi = 1 and pinned `TranShk` unconditionally. It now redraws from `IncShkDstn_Q[0][j]` and gates `TranShk` on `NewbornTransShk`, matching P. It reads `("newborn", j)` from `_base_shock_draws` when present but falls back to its own draw, because #1799's Markov `get_shocks` does not record newborn uniforms (the IndShock one does, under `"newborn"`). So Markov newborn P and Q permanent shocks remain uncoupled; that asymmetry belongs to #1799 rather than here, and is documented in the docstring rather than patched into `ConsMarkovModel.py`.

**Silent degradation.** `make_Q_measure_dstn` now warns on both degenerate branches with distinct messages, and `setup_Q_measure` warns separately when no distribution admitted a neutral measure, since that is the case where dual mode costs 2x for identical results. A test pins that ordinary calibrations stay quiet, so the warnings cannot start firing unnoticed.

Em-dashes are gone, imports are module-top, and `__all__` names 4 symbols that all resolve.

## Verification

**Default-off bit identity is exact**, max absolute difference 0.000e+00 across `cNrm pLvl mNrm aNrm MPCnow bNrm kNrm aLvl`, with `RNG.bit_generator.state` and `IncShkDstn[0]._rng.bit_generator.state` both identical, so no stream was consumed and discarded.

Full suite 839 passed, 0 failed. `tests/test_dual_measure.py` 20 passed, up from 12. `uvx ruff@0.11.8 format --check` clean across all 149 files, and `ruff check` clean.

## One thing flagged rather than fixed

**`KinkedRconsumerType` composition would be silently wrong.** `_transition_Q` calls `self.get_Rport()`, and `KinkedRconsumerType.get_Rport` (`ConsIndShockModel.py:3067-3069`) reads `self.state_prev["aNrm"]`, the P side, to choose between `Rboro` and `Rsave`. A Q agent would therefore get the rate implied by the P agent's asset position. Harmless for IndShock, which is state-independent, and for Markov, which is indexed by the shared `Mrkv` - both within this mixin's documented scope - but it is a real wrong-number bug the moment anyone composes with `KinkedRconsumerType`. Fixing it means a Q-aware `get_Rport`, which is scope expansion into a model this PR does not claim to support, so it is flagged rather than done silently.

alanlujan91 (approved):
Approving. All three blocking defects fixed and verified by reverting each hunk individually. Default-off bit identity exact at 0.000e+00 including RNG stream position; full suite 839 passed; ruff@0.11.8 clean across all 149 files.

The hidden dependency is also resolved: this PR reads `_base_shock_draws`, which #1799 now writes, and #1799 merged earlier today. Re-measured rather than assumed - corr(P, Q) holds at 0.974692 across three repeated `initialize_sim()` calls with both histories bit-reproducible, where it previously decayed 0.982 / -0.056 / -0.017.

One item flagged rather than fixed, deliberately: `KinkedRconsumerType` composition would take the borrowing-vs-saving rate from the P agent's asset position, since `get_Rport` reads `state_prev["aNrm"]`. Out of this mixin's documented scope (IndShock and Markov are unaffected), so it is documented rather than silently patched.

PR #1784, opened 2026-08-10
Add HARK.simulation.normalization: opt-in moment-normalization mixins (Markov-capable)

llorracc:

This PR adds one new module (plus tests) inside the existing `HARK/simulation/` package and **modifies no existing code file** (changelog entry only) — a bit-identity test covers the composed-but-disabled default, so default behavior is provably unchanged.

Two strictly opt-in variance-reduction mixins:

- `ShockNormalizationMixin` rescales drawn `PermShk`/`TranShk` so cross-sectional means are exactly 1 each period (per Markov state when present).
- `PermanentIncomeNormalizationMixin` pins per-age-cohort `log(pLvl)` moments to analytical targets via a rank-preserving affine transform in log space, inversely rescaling `mNrm`/`bNrm` so level quantities are preserved.

Markov-capable by design: vector (state-dependent) `PermGroFac` is supported via the stationary-distribution-weighted mean log growth, and the moments mode automatically degrades to **mean-only** under vector growth — the simple analytical variance formula omits the Markov-path variance contribution, so rescaling the cross-sectional spread to it would shrink genuine heterogeneity (the rationale is recorded in the module docstring). Shock log-moments are computed directly from the simulated discrete approximation's atoms (exact), rather than a lognormal formula.

Self-contained (no `HARK.core` changes); composes with `HARK.dual_measure`'s `DualMeasureMixin` when that module lands (`DualMeasureMixin` first in the MRO; a guard raises a clear error on the wrong order; the composition test is `importorskip`-gated until then).

Tests: exact shock means; scalar-growth cohort log-moments pinned to 1e-8; mean-only mode semantics; Markov drift/shape regressions; a stationary-distribution unit check; and the **default-path invariance test**.

- [x] Documented (numpydoc docstrings + module usage example)
- [x] CHANGELOG.md updated


alanlujan91:
Chris, multi-agent review on this one too. Two issues look like blockers, and both are on the default `IndShockConsumerType` at its own default parameters rather than in some corner case.

## 1. Normalizing `PermShk` to mean 1.0 deletes permanent income growth

`_normalize_shock_means` treats `shocks["PermShk"]` as a mean-one shock and rescales it to exactly 1.0. But HARK folds the growth factor into that array. From `ConsIndShockModel.py:2182`, with HARK's own comment:

```python
PermShkNow[idx] = (
    IncShks[0, :] * PermGroFacNow
)  # permanent "shock" includes expected growth
```

So the correct cross-sectional mean is `PermGroFac[t_cycle-1]`, not 1.0. Forcing it to 1.0 does not remove sampling noise, it removes the growth factor.

Measured on `IndShockConsumerType` at defaults (`PermGroFac = [1.01]`), `AgentCount=20000`, `T_sim=8`, same seed:

|  | plain | `normalize_shocks=True` |
|---|---|---|
| `mean(shocks["PermShk"])` | 1.010687 | 1.000000 |
| aggregate `pLvl` growth over the run | 1.06739 | 0.99973 |

Permanent income growth goes to zero. Since `transition()` computes `pLvlNow = pLvlPrev * shocks["PermShk"]`, this is a bias rather than variance reduction, and it hits every model descended from `IndShockConsumerType`.

`ConsMarkovModel` does the same per state, which makes it worse in the Markov case: normalizing within each state forces every state's mean to 1.0 and annihilates exactly the state dependent growth this PR advertises. `_effective_log_PermGroFac` then works to recover a stationary weighted mean log growth that `ShockNormalizationMixin` has already removed.

The target needs to be the group's `PermGroFac`, or `PermShk` should be left alone and only `TranShk` normalized. `TranShk` atoms are genuinely mean one, so 1.0 is the right target there.

Note that `test_shock_means_exact_when_enabled` asserts `mean(PermShk) == 1.0` to 1e-12. It passes, and it pins the behavior in place.

## 2. `_analytical_log_pLvl_moments` undercounts shock periods, collapsing young cohorts

```python
eff_periods = max(int(age_k) - 1, 0)
mu_k = p_init_mean + int(age_k) * g_log + eff_periods * e_log_psi
sigma_k = np.sqrt(p_init_std**2 + eff_periods * var_log_psi)
```

Normalization runs inside `get_states`, before the `t_age` increment, so an agent reading `t_age == k` has been through `k+1` transitions, each multiplying `pLvl` by `psi * G`. The count should be `k+1` for both the growth term and the shock term. The code uses `k` for growth and `max(k-1, 0)` for shocks, which is also internally inconsistent since those two counts should never differ.

The docstring justifies `max(k-1,0)` on the grounds that newborns receive `PermShk = 1`. `ConsIndShockModel.py:2199` redraws a random permanent shock for newborns:

```python
PermShkNow[idx] = (
    IncShkDstnNow.atoms[0][EventDraws] * PermGroFacNow
)
```

Only `TranShk` is set to 1 for newborns, and only when `NewbornTransShk` is False.

Measured, `AgentCount=40000`, `T_sim=20`, same seed:

| `t_age` | sd(log pLvl) normalized | sd(log pLvl) plain |
|---|---|---|
| 1 | 0.000000 | 0.096914 |
| 2 | 0.000000 | 0.136009 |
| 3 | 0.096672 | 0.175118 |
| 4 | 0.136715 | 0.197792 |
| 7 | 0.216166 | 0.266932 |

The two youngest cohorts are driven to identical permanent income, and every older cohort reproduces the plain series lagged by exactly two periods. This is the harm the module docstring says it is avoiding in the Markov case, happening in the scalar growth default path.

## 3. Lifecycle models are silently wrong

`_effective_log_PermGroFac` reads `self.PermGroFac[0]`, `_permshk_log_moments` reads `self.PermShkDstn[0]`, and `_analytical_log_pLvl_moments` uses `int(age_k) * g_log` for every cohort. All of that assumes growth and shock moments are constant over the life cycle. HARK ships calibrations where they are not, for instance `PermGroFac = 14*[1.03] + 25*[1.0] + [0.7] + 9*[1.0]`.

`_growth_is_vector` detects only the Markov case, so age varying growth is not detected at all, and `_normalize_shock_means` groups by `Mrkv` only, never by `t_cycle`, so a lifecycle model pools agents facing different `PermGroFac` into one target.

For an opt-in variance reduction feature I would rather this refuse to enable when `T_cycle > 1` with non-constant parameters than silently produce a fabricated age profile.

## 4. `sim_one_period` re-inlines boilerplate that main now factors out

`HARK/core.py` on main now has `_sim_period_prologue()` and `_sim_period_epilogue()` precisely so subclasses stop duplicating the loop, and the docstring there says so. This mixin copies the whole thing, so any future change to the prologue or epilogue silently diverges for normalized agents. It should collapse to:

```python
def sim_one_period(self):
    self._sim_period_prologue()
    self.get_states()
    self.post_state_hook()
    self.get_controls()
    self.get_poststates()
    self._sim_period_epilogue()
```

Related: overriding `sim_one_period` in a mixin means it shadows any model specific `sim_one_period`, and `ConsRiskyContribModel` has one. Using only `post_state_hook` would be structurally safer than shadowing the orchestrator.

## Also worth fixing

**Markov convention is ambiguous.** `_normalize_shock_means` normalizes within each Markov state; `post_sim_normalize_pLvl` cohorts on `t_age` only and applies a single stationary weighted drift, so it is pooled across states. Both are defensible, but using one for shocks and the other for `pLvl` in the same module without saying so is confusing. The pooled target is also only valid at chain stationarity, and a young cohort has not mixed yet, which is where the item 2 error bites hardest.

**`normalize_shocks=True` silently no-ops under `read_shocks`.** Normalization hooks `get_shocks()`, and `core.py` branches to `read_shocks_from_history()` instead when `read_shocks` is True. The handcrafted shock history path therefore disables the feature with no warning.

**Silent skips.** `_normalize_shock_means` skips groups with `mask.sum() < 2` and `post_sim_normalize_pLvl` skips cohorts below `_pLvl_norm_min_cohort`, both without warning, while the class docstring promises exactness "regardless of population size". A Markov state with one agent silently voids the guarantee.

**Unicode em-dashes**, 3 in `normalization.py` and 1 in `test_normalization.py`.

**Dead composition test.** `test_composition_with_dual_measure` is behind `pytest.importorskip` on `HARK.dual_measure`, which does not exist on main, so the MRO guard is entirely unverified. Worth landing with #1783 rather than here.

## On the tests

`test_scalar_growth_pins_cohort_log_moments` normalizes the data to `_analytical_log_pLvl_moments(...)` and then asserts the result equals `_analytical_log_pLvl_moments(...)`. That verifies the assignment executed; it passes for any target, right or wrong, which is why items 1 and 2 both got through.

The comparison that would catch both is against an un-normalized agent's realized cohort moments. The default-off bit-identity test is good and worth keeping as is.

alanlujan91 (approved):
Approving. Both economic defects from my earlier review are fixed, and I verified the headline one independently rather than taking the fix on trust.

**PermShk now targets PermGroFac, not 1.0.** HARK folds growth into that array (`PermShkNow = IncShks[0,:] * PermGroFacNow`, with HARK's own comment saying the permanent "shock" includes expected growth), so rescaling to 1.0 was deleting growth rather than removing noise. Measured on the merged branch at defaults, 20000 agents:

| | plain | normalized |
|---|---|---|
| mean(PermShk), last period | 1.010524 | **1.010000000000** |
| aggregate pLvl growth over the run | 1.06459 | 1.06591 |

That figure was 0.99973 before the fix, i.e. growth gone. Markov case verified per state too: with `PermGroFac=[array([1.03, 0.98])]`, state means come out at exactly 1.03 and 0.98 rather than both at 1.0, which is the state-dependent growth this PR is for. TranShk keeps the 1.0 target since its atoms genuinely are mean-one, with newborns whose TranShk the model pinned to 1.0 held out rather than rescaled. Grouping is now by income-process index and Markov state, where before it was Markov only.

**The cohort shock-period undercount is fixed.** Normalization runs inside `get_states`, before the `t_age` increment, so an agent at `t_age == k` has been through `k+1` transitions; the old code used `k` for growth and `max(k-1,0)` for shocks, which cannot both be right. Now an explicit per-period accumulation over the cohort's realized index history. sd(log pLvl) at ages 1 and 2 was exactly 0.000000 before, with every older cohort reproducing the plain series lagged by two; now 0.096672 and 0.136715 against plain 0.098149 and 0.136818.

**Lifecycle is now age-aware** rather than reading `[0]` everywhere, with the two cases age-awareness cannot repair warning instead of fabricating a profile.

**The `sim_one_period` override is gone entirely.** Main's `AgentType.sim_one_period` already calls `post_state_hook()` in exactly the right place, so the override bought nothing and shadowed models with their own `sim_one_period`. Models the hook does not reach now warn at `initialize_sim`.

**On the tests, which is what let both defects through.** The old `test_scalar_growth_pins_cohort_log_moments` normalized to `_analytical_log_pLvl_moments(...)` and then asserted equality with `_analytical_log_pLvl_moments(...)`, so it verified the assignment executed and would pass for any target, right or wrong. It now rebuilds the moments independently in the test file from `PermShkDstn[0].atoms/.pmv`, and a second test compares against a plain agent's realized cohort spreads. Every new test was checked to FAIL against the old implementation: 7 failed there, 12 pass now.

Default-off bit identity is exact, max abs difference 0.0 across pLvl, cNrm, aNrm, mNrm, PermShk and TranShk on both the infinite-horizon defaults and `init_lifecycle`, with the RNG bit-generator state identical.

One judgement call to flag, since it touches #1783: the DualMeasureMixin ordering guard was removed. With the `sim_one_period` override gone, MRO order no longer matters, and I confirmed `DualMeasureMixin.sim_one_period` calls `self.post_state_hook()` at `dual_measure.py:208`, so the hook fires either way. Had the assertion stayed it would have started failing the moment #1783 merges.

Full suite green on the merged branch.

PR #1786, opened 2026-08-11
Add opt-in low-variance draw modes (shuffle/draws/replicates) to the distributions layer

llorracc:

This PR adds opt-in low-variance draw modes to the distributions layer. It modifies two existing files, and the non-disruption case is proven two independent ways:

1. **Textual identity**: every modified method dispatches its new behavior behind a new keyword argument and keeps the default path textually identical to the previous code — `MarkovProcess.draw`'s original body survives verbatim as `_draw_iid`, `draw_events`'s default branch is unchanged, and `DiscreteDistribution.draw` only gains a `replicates=` prologue that is inert when the argument is omitted.
2. **RNG-stream goldens**: two new tests pin the exact default-path draw sequences as captured on current `main` (`a25d3ae0`) at fixed seeds. The default streams are byte-identical, not merely distributionally equivalent — the property downstream reproducibility actually depends on (cf. the 0.17.2 release-note experience with draw-order changes).

What the new modes do:

- `MarkovProcess.draw(shuffle=True)` — deterministic floor-plus-leftover target counts per source state (eliminates sampling noise in transition counts), with `sort_key=` (systematic sampling within source states) and `draws=` (rank-based stratified inverse-CDF assignment: quota-exact and asymptotically iid per agent — built for common-random-number workflows across counterfactual scenarios).
- `DiscreteDistribution.draw_events(shuffle=True)` — same construction for event draws.
- `DiscreteDistribution.draw(replicates=k)` — exact full-coverage samples: k copies of the minimal sample size J_min at which every atom's count is an integer, with a clear error when probabilities are not cleanly rational.

Tests: five behavioral tests for the new modes (CRN across matrices, quota preservation, convergence-to-iid, mutual-exclusion errors) plus the two stream-invariance goldens; the full `tests/test_distribution.py` suite passes (53) on top of main's recent distributions refactor.

Part of the same self-contained upstreaming series as #1777, #1783, #1784, #1785.

- [x] Documented (numpydoc parameter docs)
- [x] CHANGELOG.md updated


alanlujan91:
Chris, reviewing this stack as a unit (#1786 then #1793, evaluated at #1793's head merged with current main), because the two are only meaningful together. One blocking correctness issue and one blocking test-coverage issue.

## 1. `MarkovProcess._draw_shuffled` reintroduces a bug that was just fixed next door

`HARK/distributions/base.py:326-337` contains the pre-#1808 leftover-slot allocator verbatim, including

```python
eps = 1.0 / N_j
Q = K_exact - eps * K
```

That residual is wrong by construction: `K_exact` is in slot units and `eps * K` is in probability units, so `Q` is approximately `K_exact` and the leftover slots go out in proportion to the row itself rather than to the fractional remainders. The successive-sampling loop that follows (`Q[idx] = 0.0`) is the second half of the same defect: sampling without replacement proportional to size does not give inclusion probabilities proportional to size.

`DiscreteDistribution.draw` in `discrete.py` was fixed for exactly this in #1808, which merged today, so the two functions now disagree with each other.

Measured E[count] over 20000 repetitions, single source state:

| row | N | M | measured E[count] | target `N*p` | max abs |
|---|---|---|---|---|---|
| [.7,.2,.1] | 13 | 1 | [9.696, 2.205, 1.099] | [9.1, 2.6, 1.3] | 0.596 |
| [.55,.25,.15,.05] | 26 | 2 | [14.856, 6.611, 3.395, 1.139] | [14.3, 6.5, 3.9, 1.3] | 0.556 |
| [.7,.2,.1] | 7 | 2 | [4.896, 1.407, 0.696] | [4.9, 1.4, 0.7] | 0.007 |
| [.55,.25,.15,.05] | 6 | 2 | [3.292, 1.503, 0.908, 0.297] | [3.3, 1.5, 0.9, 0.3] | 0.009 |

The N=13 row is analytic rather than sampling noise: `Q = [8.408, 2.446, 1.223]` normalizes to `[0.696, 0.203, 0.101]`, so the one leftover slot lands in state 0 with probability 0.696 where it should be 0.1. The modal state is over-allocated by 6.5% and the rare state under-allocated by 15%. Rows where `N*p` happens to be integral give max abs 0.00000, which is why nothing currently fails, and the two near-zero rows above are coincidental cancellation of the two errors rather than correctness.

What makes this worse here than the same bug was on main: variance reduction turns a negligible bias into the dominant error term. Long-run share in state 0 with `TM = [[0.93, 0.07], [0.40, 0.60]]`, true stationary `pi_0 = 0.85106`, 1000 periods after 200 burn-in:

| N | mode | mean share | bias | sd across periods |
|---|---|---|---|---|
| 1000 | iid | 0.85061 | -0.00046 | 0.01115 |
| 1000 | shuffle | 0.85228 | +0.00122 | 0.00060 |

Under iid the bias is about 4% of the period-to-period standard deviation and invisible. Under shuffle it is roughly twice the standard deviation, because removing the noise is the whole point. A user turning this on to reduce Monte Carlo error gets a systematic error instead.

The fix is not to paste #1808's version into a second place. Both call sites should share one helper, since two copies of this logic is what produced the situation.

## 2. Both of #1793's parameters can be deleted with a green suite

Disconnecting each feature at its dispatch site and re-running:

| disconnected | tests failing |
|---|---|
| `markov_shuffle` in `get_markov_states` | **0 of 50** |
| `balanced_transitions` in `get_markov_states` | **0 of 50** |
| `sort_key` inside `MarkovProcess.draw` | 1 of 56, and that one is the mutual-exclusion `ValueError`, not behaviour |
| `MarkovProcess.draw(shuffle=)` | 3 of 106 |

`test_markov_consumer_shuffle` asserts only `shocks["Mrkv"].shape == (1000,)`. `sort_key` has no behavioural assertion anywhere, which also means the `state_prev` correction has no regression test guarding it.

The two `testMarkovTransitionShuffle` count tests are calibrated so the remainder path never executes: `9500 * 0.95 = 9025` and `500 * 0.5 = 250` are both integers, so `M = 0` in every row and the defective code is never reached. `test_shuffle_with_draws_preserves_quotas` asserts `|count - N_j * p| <= 1`, which floor-plus-anything satisfies unconditionally. `test_markov_shuffle_consistent_over_time` uses `atol=0.02`, roughly 16 times the measured bias.

The assertion that would have caught it is the one #1808 added for `DiscreteDistribution` and that was not mirrored here: repeat a small-N draw many times and compare the realized marginal against `p`, at a calibration where `N*p` is not integral.

`replicates=` has no tests at all.

## Also worth fixing

**Deferred import.** `from fractions import Fraction` inside `_resolve_replicates` (`discrete.py:259`). Project rule is all imports at module top.

**Zero-probability atoms raise.** `pmv=[0.5, 0.0, 0.5]` fails the `c > 0` check with a message about probabilities not being cleanly representable as rationals, which is not the actual reason, and `1.0 / P` divides by zero before that check runs. Zero-mass atoms are legitimate.

**`replicates=0` and `replicates=-1`** both return an empty array without validation.

**The "joint distribution detected" warning misfires.** It triggers on plain `[0.05, 0.95]` because `1/0.95` is not integral, so any non-uniform two-point distribution warns, and realistic income grids would warn on essentially every call.

**Silent no-ops.** `sort_key=` and `draws=` are ignored when `shuffle=False`. Confirmed: `balanced_transitions=True` with `markov_shuffle` at its default is bit-identical to the default path. A user who sets one flag and not the other gets no signal.

**Non-ASCII characters**, project rule is ASCII only: em-dashes at `base.py:289`, `discrete.py:248`, `test_distribution.py:619, 629, 657, 663`; multiplication signs at `discrete.py:237, 238`; plus-minus at `test_distribution.py:690, 710`.

## Verified good

`DiscreteDistribution.draw(shuffle=True)` is untouched by this stack and #1808's fix is intact: realized marginals over 40000 reps give max abs error 0.0019 / 0.0005 / 0.0011 / 0.0008 at N = 2/3/5/7 for `[.7,.2,.1]`, and 0.0003 at N=6 for the four-atom case, all within about one standard error. A reintroduced bias there would have shown as roughly 0.02 at N=6.

`draw(0, shuffle=True)` returns `shape=(0,)` with integer dtype rather than raising, and `draw_events` inherits the fixed path.

**Default-off bit identity against main is exact**, worst max abs difference 0.0 across ten arrays including a 500-agent 12-period `MarkovConsumerType` history and the RNG stream position.

**The `state_prev` correction in `get_markov_states` is present and working.** Instrumented, 15 of 15 calls received a sort key, all finite and strictly positive, ranging [0.293, 2.782] at t=14 and exactly 1.0 at t=0. `_sim_period_prologue` rotates `state_now` into `state_prev` before `get_shocks()`, and `get_mortality()` runs before that rotation, so newborns' fresh `pLvl` is in `state_prev` as well.

`replicates` gives exact counts where it succeeds: equiprobable J=3 at k=2 gives [2,2,2]; `[0.05, 0.95]` gives N=20 and [1,19]; a 0.07 unemployment crossed with 7 equiprobable states gives N=700; and 1/3, 1/6, 1/2 gives N=6 and [2,1,3].

One structural note. Merging main into this head conflicts only in the two test files, both pure end-of-file additions. `HARK/distributions/discrete.py` merges cleanly, which is exactly why the first finding needed measuring rather than waiting for a conflict marker to surface it.

alanlujan91:
Merge-order note, since this branch and #1801 both contain the affected code.

The biased leftover-slot allocator in `MarkovProcess._draw_shuffled` is fixed on **#1793's branch** (the top of this stack), not on this one:

```
baseline-q1-distribution-draw-modes   'eps = 1.0 /' present
shuffle-b2-markov-transitions          allocate_remainder_slots
```

Since #1793 descends from this branch, merging the stack as a unit carries the fix. What does not carry it is **#1801**, which holds its own copy of these commits rather than descending from this branch, so it still has the biased version. Merging #1801's stack before this one would put the biased allocator on main, and nothing in CI would notice, because the existing `test_shuffle` draws N=1000 where both allocators agree exactly.

Full detail and the measured marginals are on [#1793](https://github.com/econ-ark/HARK/pull/1793).

alanlujan91 (approved):
Approving as part of stack #1805, reviewed and merged whole. Full review on this PR; fixes pushed and verified.

Headline: `MarkovProcess._draw_shuffled` carried a verbatim copy of the pre-#1808 leftover-slot allocator, which git merged cleanly because it lives in a different function from the one #1808 fixed. Now factored into one shared `allocate_remainder_slots` in `HARK/distributions/base.py`, called by both sites. Measured E[count] at row [0.7,0.2,0.1], N=13: [9.696, 2.205, 1.099] -> [9.101, 2.597, 1.301] against a target of [9.1, 2.6, 1.3]. Long-run state-0 share bias went from about twice the period-to-period sd to roughly a eighteenth of it.

Both of #1793's parameters previously survived being disconnected with a fully green suite (0 of 50 failing); they now have end-to-end tests that fail when disconnected. Default-off bit identity against main is exact at 0.0 across ten arrays.

PR #1788, opened 2026-08-11
Exclude scipy 1.18.0: its spline objects cannot be deepcopy-ed (fixes the failing py3.12 CI job)

llorracc:
scipy 1.18.0's `PPoly`-family objects (`CubicHermiteSpline` et al.) raise `TypeError: cannot pickle 'module' object` under `copy.deepcopy`. HARK's `ValueFuncCRRA` deepcopies the interpolant it wraps, so any environment that resolves scipy 1.18.0 — currently the `build (ubuntu-latest, 3.12)` CI job — fails almost immediately in the **existing** test suite. Reproduced on unmodified `main`; minimal repro:

```python
from scipy.interpolate import CubicHermiteSpline
import copy, numpy as np
s = CubicHermiteSpline(np.arange(3.), np.arange(3.)**2, 2*np.arange(3.))
copy.deepcopy(s)   # TypeError on 1.18.0; fine on 1.17.x
```

With `scipy<1.18` the full pre-existing `tests/ConsumptionSaving/test_IndShockConsumerType.py` passes on py3.12 (22 passed). The pin is spelled `!=1.18.0` so a fixed 1.18.1 will be picked up automatically. (A HARK-side hardening — e.g. a `__deepcopy__` on `CubicHermiteInterp` that reconstructs rather than pickles — could follow separately; this unblocks CI now. Worth reporting upstream to scipy as well.)

This is why the `ubuntu-3.12` job currently fails on every open PR that runs the build matrix — including the unmodified-code series #1777/#1783/#1784/#1786/#1787. Merging this first turns those matrices fully green.

- [x] CHANGELOG.md updated

PR #1789, opened 2026-08-11
Add opt-in income_shuffle to IndShockConsumerType (exact per-period shock frequencies)

llorracc:

Adds an opt-in `income_shuffle` parameter to `IndShockConsumerType.get_shocks`: per-period and newborn income shocks drawn with `DiscreteDistribution.draw(N, shuffle=True)` — exact floor-plus-leftover shock frequencies with random assignment — eliminating cross-sectional sampling noise in the shock composition.

**Non-disruption:** default False; both draw sites keep the original code **verbatim** in the `else` branch, and a stream-golden test pins the exact default PermShk/TranShk draws as captured on `main` (`a25d3ae0`) — byte-identical, not merely distributionally equivalent. Full `tests/ConsumptionSaving/test_IndShockConsumerType.py` green (26).

Part of the shuffle family following up #1786 (the distributions layer); series context in the comment on #1788.

- [x] Tests (the feature suite + the stream-invariance golden)
- [x] Documented (docstring + parameter comment)
- [x] CHANGELOG.md updated


alanlujan91:
Reviewed as part of stack #1804 (#1789 -> #1792 -> #1799); the full review is on [#1799](https://github.com/econ-ark/HARK/pull/1799). One finding belongs to this PR.

**`income_shuffle` is untested.** Replacing both `if getattr(self, "income_shuffle", False):` guards with `if False:`, so the feature is entirely disconnected:

| model | result with the feature off |
|---|---|
| `ConsIndShockModel` | 60 passed, 0 failed |
| `ConsMarkovModel` | 60 passed, 0 failed |

The cause is a tolerance rather than a missing test. `test_shuffle_empirical_frequencies` uses `atol=0.02` at N=5000, but the measured maximum per-atom frequency deviation over 60 seeds is **0.00946 under iid draws** and **0.00106 under shuffle**. The tolerance sits 2.1x above the noise it is meant to detect, so it passes whichever code path runs. `atol=0.003` separates them cleanly. The other three tests are shape-and-positivity smoke.

For contrast, the same mutation applied to #1799's `_cache_base_shock_draws` guard fails 2 tests, so that half of the stack is genuinely covered.

**Verified good on this PR's own behaviour:** the draws are unbiased at every N I tried, including small ones. Against a 56-atom `IncShkDstn` over 3000 repetitions the max z on E[count] is 2.69 at N=5, 1.34 at N=100 and 0.15 at N=1000. Worth noting for the docstring though: the variance reduction is only about 20% at N=5, and the claim of "eliminating cross-sectional sampling noise" holds once N comfortably exceeds the atom count rather than in general. Markov makes small groups routine, since at `AgentCount=20` the smaller Markov group measured 4 agents against 56 atoms.

Default-off bit identity against main is exact, max absolute difference 0 across twelve fingerprints including both models' histories and the post-run RNG positions.

alanlujan91 (approved):

PR #1790, opened 2026-08-11
Add opt-in death_shuffle: deterministic death counts (IndShock + Markov)

llorracc:

Adds an opt-in `death_shuffle` parameter (`PerfForesightConsumerType.sim_death`, mirrored in `MarkovConsumerType.sim_death`): each death-probability group's death count is computed by floor-plus-remainder (unbiased expected deaths) with random selection of which agents die — removing binomial noise in mortality while preserving expected counts exactly.

**Non-disruption:** default False; both sites keep the original `DeathShks` code **verbatim** in the `else` branch (same RNG consumption in the same place), and a behavior-golden test pins a default simulation bit-identical to `main` (`a25d3ae0`). Full test file green (26).

Part of the shuffle family; series context in the comment on #1788.

- [x] Tests (feature suite + behavior golden)
- [x] Documented
- [x] CHANGELOG.md updated


alanlujan91:
Chris, the algorithm here is statistically sound and I could not break it. The problem is the tests: the feature can be entirely disconnected and every test still passes.

## Blocking: the tests do not exercise the feature

I replaced the dispatch at both `sim_death` sites (`ConsIndShockModel.py:1395`, `ConsMarkovModel.py:923`) with `if False:`, making `death_shuffle` a no-op:

| mutant | `-k DeathShuffle` | `test_IndShockConsumerType.py` + `test_ConsMarkovModel.py` |
|---|---|---|
| dispatch removed from both files | 4 passed | 36 passed |

`test_death_shuffle_deterministic_count` calls `agent._sim_death_shuffled(DiePrb_arr)` directly with a hand-built array, so it bypasses `sim_death` entirely. `test_death_shuffle_runs` asserts nothing beyond the absence of an exception. The remaining two exercise only the default path. So no assertion fails if the wiring breaks, and there is no Markov test at all even though `ConsMarkovModel.py` is modified.

The smallest fix that would have caught it: simulate with `T_age=None` and `death_shuffle=True`, then assert `history["who_dies"].sum(axis=1)` is constant across periods. That fails on the mutant. The Markov analogue would cover the second path.

## Blocking: `test_death_shuffle_deterministic_count` fails on correct code

The assertion `len(counts) == 1` holds only because `5000 * (1 - 0.98)` lands exactly on an integer. Measured, 200 calls each:

| AgentCount | LivPrb | N*p | distinct counts |
|---|---|---|---|
| 5000 | 0.98 | 100.0000000000 | {100} |
| 5001 | 0.98 | 100.0200000000 | {100, 101} |
| 5000 | 0.985 | 75.0000000000 | {75} |

At `AgentCount=5001` the test fails on code that is behaving exactly as designed. The scheme guarantees an unbiased expectation, not a fixed count, so the assertion is stronger than the contract. Asserting the mean over many calls, or pinning the calibration with a comment saying why it must stay integral, would both work.

Related: that test also runs `for _ in range(20): agent.sim_one_period()` and then discards the result, with trailing comments describing an approach that was replaced. Twenty periods are simulated for nothing.

## Non-blocking

**The parameter is registered on one of the three classes that use it.** `death_shuffle` is in `IndShockConsumerType_simulation_default` but not in `PerfForesightConsumerType_simulation_defaults`, where the method actually lives, nor in `init_indshk_markov`, which dispatches to it. Measured defaults: `PerfForesightConsumerType()` missing, `IndShockConsumerType()` False, `MarkovConsumerType()` missing. The `getattr` guard keeps it safe and the kwarg works, so this is only about the parameter not appearing in listings for two of the three.

**New lint hit:** RUF046 (`int(round(...))`) in the test, the only ruff finding not already present on `main`.

**Unrelated churn:** the two `Delta_mNrm_ZeroFunc` lambda reformats at ~1763 and ~2511.

## Worth qualifying in the CHANGELOG

The variance reduction scales with `N_g * p_g` per group, so it is strong for one homogeneous group and weak when mortality is age-graded and `AgentCount` is modest. Lifecycle age distribution from `init_lifecycle`, 65 age groups, 4000 reps:

| AgentCount | median group | groups with N*p>=1 | sd Bernoulli | sd shuffled | reduction |
|---|---|---|---|---|---|
| 250 | 3 | 0/62 | 1.826 | 1.778 | 5.2% |
| 1000 | 17 | 0/65 | 3.703 | 3.016 | 33.7% |
| 5000 | 87 | 31/65 | 8.259 | 3.297 | 84.1% |
| 20000 | 347 | 49/65 | 16.775 | 3.416 | 95.9% |

`T_age` dilutes it too, since old-age deaths are OR'd in afterward (mean sd over 8 seeds, N=2000, T_sim=400):

| T_age | shuffle=False | shuffle=True |
|---|---|---|
| None | 6.094 | 0.000 |
| 80 | 8.273 | 5.406 |
| 40 | 13.583 | 11.966 |

"Removing binomial noise in mortality" is exact only for `T_age=None` with large groups, so a qualifier would help users predict what they get.

## Verified clean

**Default-off is bit-identical to `origin/main`, max absolute difference 0.0.** Checked `cNrm`, `mNrm`, `aNrm`, `pLvl`, `t_age`, `who_dies` on an infinite-horizon run (N=500, T_sim=25, seed=1234) and a lifecycle run (`init_lifecycle`, T_age=11, seed=77), with identical NaN masks. Post-simulation RNG draws also match exactly, which is the check that catches a changed draw count rather than a changed value.

**Marginal death probability is exactly `p`.** N=1000, p=0.0235, 40000 reps: realized 0.023503. Across 10 index deciles, max |z| = 2.05, so the remainder allocation shows no bias toward low- or high-index agents. Heterogeneous case, 4 groups of 300 at p = 0.01 / 0.05 / 0.10 / 0.333: per-group bias +0.000000 for all four.

**The induced correlation is the intended one:** within-group pairwise correlation -0.01012 at N=100, against the theoretical -1/(N-1) = -0.01010.

**Edge cases:** N=1 p=0.5 gives mean 0.4966; p=0 gives no deaths; p=1 kills all; N=50 p=0.001 with `N*p < 1` gives mean 0.0488 against a target of 0.05. Markov per-state unbiased at LivPrb = [0.99, 0.90]: realized 0.01003 and 0.10012, total-death sd 10.011 to 2.552.

No Unicode dashes in any added line, all imports at module top, and no timing regression (20000 agents, 30 periods: 0.123s to 0.108s infinite-horizon, 0.157s to 0.144s lifecycle).

Where nothing turned up, the specific things looked for were index and age bias in the remainder allocation, float error in the `floor` (the default `1 - 0.98` leaves a 8.5e-14 remainder, so the extra coin is drawn but effectively never fires), and mutation of the read-only view from `np.broadcast_to`.

Fix the tests and this is good to merge.

alanlujan91 (approved):
Algorithm verified sound and I could not break it: marginal death probability exactly `p` (realized 0.023503 against 0.0235 over 40000 reps), no bias across index deciles, induced within-group correlation -0.01012 against the theoretical -0.01010, and correct at the edges (N=1, p=0, p=1, N*p<1). Markov unbiased per state.

The blocking issue was test coverage, now fixed: the new tests go through `sim_death()` rather than calling `_sim_death_shuffled()` directly, and a mutation check confirms they discriminate. With the dispatch disabled at both sites, 5 of the feature tests fail and the 3 that still pass are the default-path ones that should be insensitive. The brittle `len(counts)==1` assertion, which failed on correct code at `AgentCount=5001`, now has both a guard that the calibration keeps `N*p` integral and a separate test stating the real unbiasedness contract.

Default-off bit-identity re-confirmed against current main at max abs difference 0.0, including RNG stream position.

PR #1791, opened 2026-08-11
Add opt-in init_shuffle: exact-marginal initial-state draws (IndShock + Markov)

llorracc:

Adds an opt-in `init_shuffle` parameter (`PerfForesightConsumerType.sim_birth`, mirrored for `MarkovConsumerType`'s initial Markov states): initial states drawn with exact-marginal matching (`DiscreteDistribution.draw(N, shuffle=True)`), eliminating sampling noise in the initial cross-section — a residual noise source per-period flags cannot reach.

**Non-disruption:** default False, and the `shuffle` kwarg is passed **only when enabled** (`{"shuffle": True} if flag else {}`), so duck-typed continuous init distributions whose `draw()` lacks the kwarg keep working at the default. A behavior-golden test pins a default simulation bit-identical to `main` (`a25d3ae0`). Full test file green (27).

Part of the shuffle family; series context in the comment on #1788.

- [x] Tests (feature suite + behavior golden)
- [x] Documented
- [x] CHANGELOG.md updated


alanlujan91:
Chris, this one turned up something bigger than the PR. The wiring here is right and the default path is provably untouched, but `DiscreteDistribution.draw(shuffle=True)` on `main` has an arithmetic bug, and this PR is what routes small-`N` traffic into it. I would hold this until that is fixed, then it should be fine as written.

## The bug is on main, at `HARK/distributions/discrete.py:248`

```python
K_exact = N * P                 # slot units, order N*P
K = np.floor(K_exact).astype(int)
eps = 1.0 / N
Q = K_exact - eps * K           # slot units minus probability units
```

`Q` is meant to be the leftover fractional mass, which is `K_exact - K` in slot units, or `P - K/N` in probability units. Either normalizes to the same `Q_adj`. What the line computes is `N*P - K/N`, and since `K/N` is approximately `P`, that term is negligible against `N*P`. So `Q_adj` collapses to roughly `P`, and the `M` leftover slots get allocated in proportion to `P` itself rather than to the remainders. The result is that `E[count_j] != N*P_j`.

By hand at `P = [0.7, 0.2, 0.1]`, `N = 3`: `K = [2,0,0]`, `M = 1`. The remainders are `[0.1, 0.6, 0.3]`, which would give `E = [2.1, 0.6, 0.3]`, exactly `N*P`. The code instead gets `Q = [1.433, 0.6, 0.3]`, so `E = [2.614, 0.257, 0.129]`.

Measured, repeated `draw(N, shuffle=True)`, target `0.700 / 0.200 / 0.100`:

| N | as-is on main | with `Q = K_exact - K` |
|---|---|---|
| 1 | 0.700 / 0.200 / 0.099 | 0.700 / 0.200 / 0.099 |
| 2 | 0.800 / 0.134 / 0.066 | 0.699 / 0.201 / 0.100 |
| 3 | 0.872 / 0.086 / 0.042 | 0.700 / 0.200 / 0.099 |
| 5 | 0.738 / 0.238 / 0.023 | 0.701 / 0.200 / 0.099 |
| 100 | 0.700 / 0.200 / 0.100 | 0.700 / 0.200 / 0.100 |

The error is `O(M/N)`, so it vanishes at `N = AgentCount` and dominates at rebirth-cohort sizes. That is why nothing has caught it: every existing caller draws a full population.

## Why this PR makes it live

`sim_birth` is called every period from `get_mortality` with `N = who_dies.sum()`, so this is the small-`N` regime. Driving `draw` with cohort sizes from the mortality process over 4000 periods:

| AgentCount | LivPrb | E[cohort] | shuffle=True | shuffle=False |
|---|---|---|---|---|
| 200 | 0.98 | 4 | 0.764 / 0.186 / 0.050 | 0.701 / 0.198 / 0.101 |
| 500 | 0.99 | 5 | 0.753 / 0.197 / 0.050 | 0.699 / 0.200 / 0.102 |
| 1000 | 0.98 | 20 | 0.714 / 0.199 / 0.087 | 0.699 / 0.201 / 0.101 |
| 10000 | 0.98 | 200 | 0.701 / 0.200 / 0.099 | 0.700 / 0.200 / 0.100 |

The rarest initial state is under-represented by half at typical cohort sizes. Turning the flag on to remove sampling noise would substitute a systematic bias that averaging cannot remove. This is `MrkvInitDstn`'s regime specifically, since `MrkvPrbsInit` is rarely equiprobable; `kNrmInitDstn` and `pLvlInitDstn` are equiprobable, so symmetry happens to spare them.

## Second blocker: `init_shuffle=True` crashes on a zero-death period

`sim_birth` runs every period including when nobody dies. At `N = 0` the shuffle branch builds an empty float64 array and `atoms[..., indices]` raises.

```python
a = IndShockConsumerType(AgentCount=100, T_sim=3, LivPrb=[1.0], init_shuffle=True, seed=0)
a.solve(); a.initialize_sim(); a.simulate()
# IndexError: arrays used as indices must be of integer (or boolean) type
```

A direct `draw(0, shuffle=True)` raises `ZeroDivisionError` at `eps = 1.0 / N`; the agent path passes `np.int64(0)`, which gives `inf` and falls through to the `IndexError`. Per-period probability is `LivPrb ** AgentCount`, so at `AgentCount=200, LivPrb=0.98` it is 0.0176 per period, which is 83% within 100 periods. The tests miss it only because they use `AgentCount=1500`.

## Suggested sequencing

Both defects live in `discrete.py`, not in this diff. Fixing `Q = K_exact - K` and adding an `N == 0` early return would make this PR correct as written. Since #1786 is already reworking that function, it may be cleanest to fix it there, or directly on `main` given that `shuffle=` is already shipped.

## Non-blocking

**The duck-typing claim holds only at the default.** With `init_shuffle=True` and a continuous init distribution, `Lognormal` raises `TypeError: Distribution.draw() got an unexpected keyword argument 'shuffle'`. The CHANGELOG says continuous init distributions keep working, without that qualifier.

**Unicode em-dashes** at `tests/ConsumptionSaving/test_IndShockConsumerType.py:920, 948, 962, 1029`, plus non-ASCII math glyphs in docstrings and two assertion messages.

**Unrelated reformat.** The `Delta_mNrm_ZeroFunc` lambda reparenthesization at `ConsIndShockModel.py` ~1734 and ~2482 is semantically identical churn and could come out.

## Tests

They do discriminate against silent disablement: forcing `_kw = {}` fails `test_init_shuffle_exact_frequencies` (atom counts 88-115 rather than 99-101) and `test_init_shuffle_reduces_seed_variance` (SD ratio 1.0 against a required 0.1). That is the right design.

What they do not cover is exactly what let both blockers through: nothing exercises the mortality path with the flag on, nothing uses `N=0`, nothing uses a non-equiprobable `pmv`, and there is no `MarkovConsumerType` test for `init_shuffle` at all, so the second of the two code paths this PR adds is untested. A `LivPrb=[1.0]` smoke test and a long-run marginal test on `MrkvPrbsInit=[0.7,0.3]` at `AgentCount=200` would cover both.

## Verified good

Default-off bit-identity holds: against a separately built `origin/main` tree at `AgentCount=200, T_sim=8, seed=555`, the `cNrm`, `aNrm` and `pLvl` histories are bit-identical and the `kNrmInitDstn` RNG counter matches exactly. Omitting the parameter and passing `False` explicitly also agree exactly.

No index correlation is induced: at `AgentCount=1500` over 6 seeds, `max|rho(idx, kNrm)| = 0.049` and `max|rho(kNrm, pLvl)| = 0.027`, indistinguishable from the `False` path. The trailing `permutation` does its job and the two init distributions hold distinct RNG objects.

alanlujan91:
Correction to my review above, on two points.

**The `Delta_mNrm_ZeroFunc` lambda reformats are not stray churn, and I was wrong to ask for them to be dropped.** They are a ruff version artifact. CI runs pre-commit, which pins **ruff 0.11.8** (`.pre-commit-config.yaml`), while a recent local ruff (0.15.x, 0.16.x) wants the other form. Reverting them would make a modern local `ruff format --check` fail while satisfying CI. Nothing to do here; ignore that bullet.

Worth knowing generally for this repo, since it silently invalidates the obvious check: **`ruff format --check HARK/ tests/` reports "No Python files found" and exits without checking anything.** `pyproject.toml` sets `[tool.ruff] include = ["*.ipynb"]`, which *replaces* ruff's default file discovery rather than extending it, so directory arguments match only notebooks. Pre-commit passes explicit paths, so CI is unaffected, but a hand-run directory-form check is a no-op that looks like a pass. Check explicit file paths, and with `uvx ruff@0.11.8` if you want to match CI.

**The two blocking findings in my review were not defects in this PR.** Both live in `DiscreteDistribution.draw(shuffle=True)` on main: the biased remainder allocation and the `N=0` crash. This PR is written correctly; it is simply the first caller to route small-cohort traffic into them. Both are fixed in #1808, which also carries the regression tests. Once that lands, rebase and I expect this to be clean.

The branch is now up to date with main.

alanlujan91 (approved):
Both blocking findings from my earlier review are resolved, and neither was a defect in this PR: they were in `DiscreteDistribution.draw(shuffle=True)` on main, now fixed by #1808. Verified against this branch merged onto current main rather than assumed.

Zero-death periods no longer raise. `LivPrb=[1.0]` with `init_shuffle=True` runs clean, as does a 400-period run at `AgentCount=200, LivPrb=0.98`, which previously had an 83% chance of hitting the `N=0` path within 100 periods.

Small-cohort marginals are now exact at the sizes `sim_birth` actually redraws, which is the `MrkvInitDstn` regime where `MrkvPrbsInit` is rarely equiprobable:

| cohort size | realized | target |
|---|---|---|
| 4 | 0.700 / 0.200 / 0.100 | 0.700 / 0.200 / 0.100 |
| 5 | 0.701 / 0.200 / 0.099 | 0.700 / 0.200 / 0.100 |
| 20 | 0.700 / 0.200 / 0.100 | 0.700 / 0.200 / 0.100 |

Previously the rarest state came out about half as often as it should at these sizes.

Also confirmed from the earlier review and still holding: default-off bit-identity is exact, including the `kNrmInitDstn` RNG counter position, and no correlation is induced between agent index and initial state (max |rho| 0.049 over 6 seeds at `AgentCount=1500`).

47 tests pass on the merged branch, and `ruff@0.11.8` (the version CI pins) is clean. The `Delta_mNrm_ZeroFunc` reformatting I asked about is a ruff version artifact, not churn; ignore that note.

Remaining non-blocking items are yours: the CHANGELOG says continuous init distributions keep working, which holds only at the default (with `init_shuffle=True` a `Lognormal` raises `TypeError` on the unexpected `shuffle` kwarg), there is still no `MarkovConsumerType` test for `init_shuffle`, and there are Unicode em-dashes in the test file.

PR #1792, opened 2026-08-11
Extend income_shuffle to MarkovConsumerType (stacked on the IndShock PR)

llorracc:

> 🔗 **Stacked on #1789** (income_shuffle for IndShockConsumerType) — its base is that PR's branch, so this diff shows only the Markov increment. Merge #1789 first, then retarget/merge this.

Extends `income_shuffle` to `MarkovConsumerType.get_shocks`: both the per-state income loop and the newborn redraw gain the same opt-in shuffle branch (`DiscreteDistribution.draw(N, shuffle=True)` for exact shock frequencies).

**Non-disruption:** default False; the original `draw_events` RNG paths are preserved **verbatim** in the `else` branches. Full `tests/ConsumptionSaving/test_ConsMarkovModel.py` green.

- [x] Tests · [x] Documented · [x] CHANGELOG.md updated


alanlujan91:
Chris, same note as on #1793: I pushed a commit to this branch (`608647c6`) without explaining it at the time.

It registers the flag in the parameters dictionary:

```python
"income_shuffle": False,  # Exact per-period shock frequencies when True (see get_shocks)
```

`get_shocks` reads it with `getattr(self, "income_shuffle", False)`, and `init_indshk_markov` did not define it, so `MarkovConsumerType` had no default for a parameter the class consults. It worked when passed as a kwarg, but the flag did not appear in the type's parameter listing and there was no default to inspect. Also dropped two Unicode em-dashes.

Worth knowing for review purposes: that `getattr` pattern is not a dead read here, because `AgentType.__init__` assigns every parameter-dict entry as an attribute, so the dict key is the assignment. I flagged it as a dead read initially and was wrong; it caught me again on #1790 and #1791.

This is the middle of `stack #1804` (#1789 -> #1792 -> #1799), so it merges with the chain rather than alone, and neither #1789 nor #1799 has been reviewed yet.

alanlujan91 (approved):

PR #1793, opened 2026-08-11
Add opt-in markov_shuffle / balanced_transitions to MarkovConsumerType (stacked on the distributions PR)

llorracc:

> 🔗 **Stacked on #1786** (the distributions-layer draw modes) — its base is that PR's branch (`MarkovProcess.draw` needs the `shuffle=`/`sort_key=` kwargs). Merge #1786 first, then retarget/merge this.

Adds opt-in `markov_shuffle` and `balanced_transitions` to `MarkovConsumerType.get_markov_states`: quota-exact Markov transitions via `MarkovProcess.draw(shuffle=True)`, optionally with systematic sampling by pLvl within each source state (pLvl deliberately, never wealth — a wealth sort key creates a poverty-trap feedback loop; documented at the call site).

**Non-disruption:** both flags default False and are read with `getattr` (no AttributeError on instances whose parameter dicts lack the keys); the default call path is unchanged. Full `test_ConsMarkovModel.py` green on the stack.

- [x] Tests · [x] Documented · [x] CHANGELOG.md updated


alanlujan91:
Chris, I pushed a one-token fix to this branch (`4c57e664`) and should have said so at the time rather than leaving it in the log unexplained.

```diff
-                sort_key = self.state_now["pLvl"][right_age]
+                sort_key = self.state_prev["pLvl"][right_age]
```

`balanced_transitions` sorts agents by permanent income to assign Markov transitions, but at the point `get_shocks()` runs, `state_now["pLvl"]` has already been blanked by `_sim_period_prologue`, which fills it with `np.empty`. That is uninitialized memory rather than NaN, so the sort key was whatever bytes happened to be at that address. In a simulation loop those bytes are usually a previously freed array of the same quantity, so the values look like plausible `pLvl` figures and nothing fails loudly. `state_prev` holds the values the sort actually wants.

Three agents reached the same conclusion by different routes: reading `_sim_period_prologue`, instrumenting the array before the sort, and diffing the assignment against the sibling `income_shuffle` path in #1792, which already used `state_prev`.

Also dropped one Unicode em-dash for the ASCII rule.

Note this PR is stacked on #1786 (`stack #1805`), so it merges as a unit with it rather than on its own. #1786 itself has not been reviewed yet.

alanlujan91:
Reviewed as part of stack #1805 together with #1786; the full review is on [#1786](https://github.com/econ-ark/HARK/pull/1786#issuecomment-latest). Two points concern this PR directly.

**This PR turns on a biased allocator.** `MarkovProcess._draw_shuffled` in #1786 carries the pre-#1808 leftover-slot code verbatim, and `markov_shuffle` is what routes traffic into it. Measured at row `[0.7, 0.2, 0.1]`, N=13: E[count] = [9.696, 2.205, 1.099] against a target of [9.1, 2.6, 1.3]. That row is analytic rather than noise. It matters more here than the same bug did on main, because reducing the variance is the point: the long-run state-0 share bias goes from 4% of the period-to-period sd under iid draws to about 200% of it under shuffle.

**Both parameters this PR adds can be silently deleted with a green suite.** Disconnecting `markov_shuffle` at its dispatch site fails 0 of 50 tests; disconnecting `balanced_transitions` also fails 0 of 50. The existing count tests are calibrated at `9500 * 0.95` and `500 * 0.5`, both integral, so `M = 0` and the remainder path never runs.

That second point matters for the fix I pushed here earlier (`4c57e664`), which changed the `balanced_transitions` sort key from `state_now["pLvl"]` to `state_prev["pLvl"]` because `_sim_period_prologue` blanks `state_now` with `np.empty` before `get_shocks()` runs. I verified it is live: 15 of 15 calls receive a finite, strictly positive sort key, range [0.293, 2.782] at t=14. But nothing tests it, so it can silently regress.

Fixes for both are in progress and will be pushed to these branches.

alanlujan91:
Chris, we were working on this branch at the same time, so a few things to sync up. Nothing of yours was overwritten: your rebase is the base, my work sits on top as a fast-forward.

## Your rebase dropped a correctness fix

The rebase discarded commit `4c57e664`, and the new head had gone back to

```python
sort_key = self.state_now["pLvl"][right_age]
```

in `balanced_transitions`. That reads uninitialized memory. `_sim_period_prologue` blanks `state_now` with `np.empty` before `get_shocks()` runs, so the sort key is whatever bytes were at that address, and in a simulation loop those are usually a previously freed array of the same quantity, which is why the values look like plausible `pLvl` figures. `state_prev` holds what the sort actually wants.

I re-applied it as `7dc2ee80`. Flagging it because a further rebase would drop it again the same way.

## The allocator bug, and why it needed measuring rather than reading

`MarkovProcess._draw_shuffled` in #1786 contained a verbatim copy of the leftover-slot code that #1808 fixed on main last night:

```python
eps = 1.0 / N_j
Q = K_exact - eps * K
```

`K_exact` is in slot units and `eps * K` is in probability units, so `Q` is approximately `K_exact` and the leftovers go out in proportion to the row rather than to the fractional remainders. Because it lives in a different function from the one #1808 touched, git merged it cleanly and no conflict marker ever appeared.

Measured E[count] over 20000 repetitions:

| row | N | before | after | target |
|---|---|---|---|---|
| [.7,.2,.1] | 13 | [9.696, 2.205, 1.099] | [9.101, 2.597, 1.301] | [9.1, 2.6, 1.3] |
| [.55,.25,.15,.05] | 26 | [14.856, 6.611, 3.395, 1.139] | [14.298, 6.500, 3.900, 1.301] | [14.3, 6.5, 3.9, 1.3] |

Max absolute error goes from 0.596 to 0.0026 on the first row. That row is analytic rather than noisy: the single leftover slot was going to state 0 with probability 0.696 where it should be 0.1.

What makes this matter more here than the same bug did on main is that variance reduction inverts the signal-to-noise. Long-run share in state 0, `TM = [[0.93, 0.07], [0.40, 0.60]]`, true `pi_0 = 0.85106`:

| N | mode | bias before | bias after | sd |
|---|---|---|---|---|
| 1000 | iid | -0.00046 | -0.00046 | 0.01115 |
| 1000 | shuffle | +0.00122 | **+0.00004** | 0.00072 |

Before the fix the bias was roughly twice the period-to-period standard deviation; now it is about a eighteenth of it. Someone turning shuffle on to reduce Monte Carlo error was getting a systematic error in exchange.

Rather than copy #1808's version into a second place, which is what produced this, `allocate_remainder_slots(K_exact, K, M, rng)` is now module-level in `HARK/distributions/base.py` and both `DiscreteDistribution.draw` and `MarkovProcess._draw_shuffled` call it. Its docstring records why it is shared. It still draws `M` uniforms while using only the first, so RNG stream position is unchanged, and default-off bit identity against main is exact at 0.0 across ten arrays including `DiscreteDistribution.draw`'s own output.

## `MrkvArray` passed to the constructor is silently ignored

Worth knowing independently of this PR. `MarkovConsumerType(MrkvArray=[[0.9, 0.1], [0.1, 0.9]])` does not do what it looks like: `MrkvArray` is a constructed parameter, so the agent runs on the default matrix and the argument is discarded without a warning. `test_markov_consumer_shuffle` passes exactly that and has been testing the default chain. Assigning after construction works. This cost a full round of confidently wrong measurements before anyone checked `agent.MrkvArray`.

## Test coverage

Both of this PR's parameters could be deleted with a completely green suite:

| disconnected | failures before | failures now |
|---|---|---|
| `markov_shuffle` in `get_markov_states` | 0 of 50 | both new end-to-end tests |
| `balanced_transitions` in `get_markov_states` | 0 of 50 | `test_balanced_transitions_makes_movers_representative` |
| sort key back to `state_now` | (no test existed) | same test |

The existing count tests were calibrated at `9500 * 0.95` and `500 * 0.5`, both integral, so `M = 0` in every row and the defective remainder path never executed. The new marginal test uses a fractional case for that reason.

## Smaller items

`from fractions import Fraction` moved to module top. Zero-probability atoms no longer raise in `_resolve_replicates` and the error message names the offending atoms. `replicates <= 0` and non-integral values raise instead of returning an empty array. `sort_key=` and `draws=` now warn when `shuffle=False` rather than being silently ignored.

The "joint distribution detected" warning was re-keyed rather than removed. Its old test, `1/p_j` non-integral, is true of nearly every non-uniform pmv, so `[0.05, 0.95]` warned. Covering an atom of probability `p_min` needs `ceil(1/p_min)` draws as arithmetic, so the informative event is `J_min` exceeding that floor. It is now silent on `[0.05, 0.95]` and on 0.05 crossed with 3 equiprobable states, and warns on 0.07 unemployment crossed with 7 equiprobable states, where the rarest atom needs 100 draws but the LCM needs 700.

Also fixed: the two test files were not ruff-formatted after the rebase, and two Unicode dashes had come back in `testMarkovTransitionShuffle`. Both would have failed CI independently of any of the above. Note CI pins ruff 0.11.8 in `.pre-commit-config.yaml`, which formats differently from a current local ruff, so `uvx ruff@0.11.8 format --check <explicit paths>` is the check that matches. Passing a directory instead of explicit paths silently checks nothing, because `[tool.ruff] include = ["*.ipynb"]` replaces ruff's file discovery.

Full suite green: 795 passed, 1 skipped.

One judgement call: resolving the cherry-pick conflict I kept your file layout, where `testMarkovTransitionShuffle` moved to the end of the file, and appended only the new class. Restoring the old position would have left three copies of that class.

alanlujan91 (approved):
Approving as part of stack #1805, reviewed and merged whole. Full review on this PR; fixes pushed and verified.

Headline: `MarkovProcess._draw_shuffled` carried a verbatim copy of the pre-#1808 leftover-slot allocator, which git merged cleanly because it lives in a different function from the one #1808 fixed. Now factored into one shared `allocate_remainder_slots` in `HARK/distributions/base.py`, called by both sites. Measured E[count] at row [0.7,0.2,0.1], N=13: [9.696, 2.205, 1.099] -> [9.101, 2.597, 1.301] against a target of [9.1, 2.6, 1.3]. Long-run state-0 share bias went from about twice the period-to-period sd to roughly a eighteenth of it.

Both of #1793's parameters previously survived being disconnected with a fully green suite (0 of 50 failing); they now have end-to-end tests that fail when disconnected. Default-off bit identity against main is exact at 0.0 across ten arrays.

PR #1794, opened 2026-08-11
Add opt-in shuffle= to AgentType.make_shock_history (thin wrapper)

llorracc:

`AgentType.make_shock_history` gains an opt-in `shuffle=` keyword (default False): the pre-drawn shock history can be generated with the low-variance draw modes (`income_shuffle`/`markov_shuffle`, where the composed class defines them) temporarily enabled and then restored — so `read_shocks` reproductions can carry a variance-reduced history. The flag composes with #1789/#1790/#1791/the Markov PRs and is harmless before they land (it toggles attributes only where they exist).

**Non-disruption, by construction:** re-authored as a **thin wrapper** — the original method body survives **verbatim** as `_make_shock_history` (the diff renames one `def` line and adds the wrapper; no re-indent of the 85-line body), the default path delegates unchanged, and a stream-golden test pins the default pre-drawn PermShk history bit-identical to `main` (`a25d3ae0`). Full `tests/test_core.py` green (135).

- [x] Tests (stream golden + flag toggle/restore) · [x] Documented · [x] CHANGELOG.md updated


alanlujan91:
Forwarded onto current main. The merge needed a real resolution rather than a mechanical one: main now has `post_state_hook` from #1787 immediately where this PR splits `make_shock_history` into a wrapper plus `_make_shock_history`, so both methods had to be kept and reordered so the original docstring stays attached to `_make_shock_history`. `tests/test_core.py` was a pure append on both sides. 141 tests pass and the formatter is clean.

The reason I have not marked this ready is that **`shuffle=True` is currently a silent no-op**, measured on the merged branch:

```python
a = IndShockConsumerType(AgentCount=20, T_sim=3, seed=0)
[k for k in ('income_shuffle','markov_shuffle') if hasattr(a, k)]   # -> []
```

```python
saved = {k: getattr(self, k) for k in ("income_shuffle", "markov_shuffle") if hasattr(self, k)}
```

`income_shuffle` arrives with #1789 and `markov_shuffle` with #1793, neither of which is merged. Until then `saved` is empty, the loop sets nothing, and `make_shock_history(shuffle=True)` pre-draws exactly the same history as `shuffle=False` without saying so. The `hasattr` guard makes that degrade silently rather than raise, so no test can see it and CI stays green.

Two options, either is fine by me: merge this after #1789 and #1793, or keep the guard and `warn()` when `shuffle=True` finds no flags to toggle, so a user who asks for the variance reduction learns they did not get it.

Same shape as #1783, which is inert until #1799 lands. Worth a line in each PR body stating the dependency, since in both cases the note lives only in the producer's PR and whoever merges the consumer sees nothing.

alanlujan91:
Forwarded onto current main and merged. The silent no-op I flagged earlier is resolved, because its dependencies landed: `income_shuffle` came with #1789/#1792 and `markov_shuffle` with #1793, all merged today.

Verified on the merged branch rather than assumed:

```
IndShock  flags present: ['income_shuffle']
Markov    flags present: ['income_shuffle', 'markov_shuffle']

make_shock_history(shuffle=False) vs (shuffle=True) identical:  False
sd of per-period mean PermShk    iid 0.002491   shuffle 0.000276
```

So the `hasattr` loop now finds the flags, the pre-drawn history genuinely differs, and the variance reduction it promises is real, about 9x on the per-period mean at AgentCount=2000.

Worth recording why this was invisible before: the enabling loop is

```python
saved = {k: getattr(self, k) for k in ("income_shuffle", "markov_shuffle") if hasattr(self, k)}
```

Because the flag names are a loop variable rather than string literals, no static scan for `getattr(self, "literal")` can see the dependency, and `hasattr` returning False degrades silently rather than raising. The only way to establish it either way was to construct an agent and check. That is worth a line in the docstring: with neither flag defined, `shuffle=True` is a no-op by design.

186 tests pass, `ruff@0.11.8` clean repo-wide.

alanlujan91 (approved):
Approving. Dependencies now on main, verified the flag takes effect and delivers the variance reduction (per-period mean sd 0.002491 -> 0.000276).

PR #1796, opened 2026-08-11
Hierarchical-Markov helpers: general [i][j] format, extract inverse, KS plumbing

llorracc:

Additive helpers for hierarchical (macro × micro) Markov structures:

- `make_hierarchical_mrkv_array` auto-detects a **general nested `[i][j]` format** (conditional micro matrices depending on both source and destination macro state — Krusell-Smith style) alongside the existing flat destination-conditioned format. The existing-format path is unchanged — pinned by an exact-equality test.
- **`extract_cond_mrkv_arrays`** — the inverse: recovers the `[i][j]` blocks from a combined transition matrix (round-trip tested; zero-probability blocks return as zeros).
- KS plumbing: `KrusellSmithEconomy.make_MrkvArray` additionally stores `MacroMrkvArray`/`CondMrkvArrays`, and `KrusellSmithType.market_vars` distributes them. Purely additive — nothing consumes them yet (the hierarchical refactor that follows does).

Full `tests/ConsumptionSaving/test_ConsAggShockModel.py` green on this branch (16, including the seeded KS assertions).

- [x] Tests (5 new unit tests + the KS suite) · [x] numpydoc · [x] CHANGELOG.md updated


alanlujan91:
Chris, this one is in good shape. The inversion is mathematically correct, the round trip is exact, and the tests genuinely discriminate. Nothing here blocks merge; the items below are hardening.

## Verified

Round trip exact at every size tried, including non-square `M != N`:

| case | max abs error, recovered conditionals | max abs error, rebuild |
|---|---|---|
| M=3, N=4, random | 5.55e-17 | 0.0 |
| M=2, N=2, with a zero macro block | - | 0.0 |
| Krusell-Smith economy | - | 0.0 |

The hardcoded `2` at the KS call site is structurally correct rather than coincidental: `make_MrkvArray` builds a fixed 4x4 with `StateCount = 2`, and the `BU, BE, GU, GE` ordering is exactly `combined = N*macro + micro`. Block `(i,j)` row sums equal `MrkvAggArray[i,j]` to machine precision for all four blocks, every extracted conditional is row-stochastic with strictly positive entries (min 0.0167), and the rebuild is bit-exact. `update()` calls `make_MrkvArray()` before `give_agent_params()`, so the new `market_vars` entries always exist; I ran `give_agent_params()` and confirmed both reach the agent. 16 tests pass.

Format auto-detection handles flat list, nested list, `(M,N,N)` and `(M,M,N,N)` ndarray, and the `M=1` and `N=1` degenerate cases. Wrong-length input raises `IndexError` rather than returning garbage.

The tests catch all six mutations I applied: dropping the division (3 failed), transposing the extracted block (3), swapping `i`/`j` in the slice (3), `MacroMrkvArray[j,i]` (2), builder using `CondMrkvArrays[j][i]` (1), removing the zero guard (1). `test_extract_round_trip` is the workhorse, since it compares recovered conditionals directly and so survives a defect that would cancel between the two functions.

## Worth adding

**The helper never checks the structure it assumes.** `extract_cond_mrkv_arrays` returns `block / p_macro` unconditionally. That is correct exactly when `block.sum(axis=1) == MacroMrkvArray[i,j]` row by row, which is necessary and sufficient for the result to be row-stochastic. When it does not hold, non-stochastic matrices come back silently:

| input | recovered `[0][0]` row sums | raised |
|---|---|---|
| product form | 1.0, 1.0 | - |
| block scaled by 0.123 | 0.183, 0.183 | nothing |
| block rows summing to 0.30 and 0.35 against `p = 0.9` | 0.333, 0.389 | nothing |

One check covers all of it:

```python
if not np.allclose(block.sum(axis=1), p_macro):
    raise ValueError(f"block ({i},{j}) is not p_macro times a stochastic matrix")
```

That matters because a downstream caller doing `probs / probs.sum()` gets a silently wrong distribution rather than a crash.

**`N` is not checked against the array shape.** The slicing assumes `MrkvIndArray.shape == (M*N, M*N)` but never verifies it, so a wrong `N` produces well-formed garbage. On a 12x12 array whose true `N` is 4, passing `N=3` returns nine `(3,3)` blocks with row sums 0.672, 0.521, 0.978 and no error. Too large is worse: `N=5` on a 4x4 array returns blocks shaped `(4,4)`, `(4,0)`, `(0,4)`, `(0,0)`. An assert costs nothing, and the only call site passes a literal.

**Zero macro transitions return a matrix that is not a transition matrix.** When `p_macro == 0` the fibre is the whole set of row-stochastic matrices, so any choice round-trips. The code picks `np.zeros((N,N))`, which is the one class of choice that is not row-stochastic. `np.full((N, N), 1.0 / N)` round-trips identically, measured rebuild error 0.0, and stays stochastic, so a consumer that later normalizes gets uniform rather than `nan`. Your docstring documents the current behaviour so this is a judgement call, but the zeros are a landmine for the `[i][j]` consumer in #1801.

**Import inside a test method.** `test_hierarchical_mrkv_helpers.py:75` has `from HARK.ConsumptionSaving.ConsAggShockModel import KrusellSmithEconomy` inside the method. There is no cycle to justify it, since `ConsAggIndMarkovModel` imports only numpy and `HARK.AgentType`, so it belongs at module top.

**An unfalsifiable assertion.** `ConsAggShockModel.py:2872` does `self.MacroMrkvArray = MrkvAggArray`, which aliases rather than copies, so `economy.MacroMrkvArray is economy.MrkvAggArray`. The test's `assert_array_equal(MacroMrkvArray, MrkvAggArray)` therefore cannot fail unless the assignment line is deleted outright. Either `.copy()` on assignment, or drop that assertion as uninformative.

## Merge order

This is registered as standalone with base `main`, but `extract_cond_mrkv_arrays` is consumed by #1801 on `q5d-hierarchical-reparent`, the third entry of stack #1803. The dependency points the harmless way round, since merging standalone PRs first satisfies it for free, but the constraint is real: **this needs to land before #1801.** Worth a line in #1801's body so whoever merges that one knows.

The em-dashes in `ConsAggIndMarkovModel.py` predate this branch, and the `convertKtoY`/`wFunc` reformatting is semantically identical.

alanlujan91 (approved):
Approving. The inversion is mathematically correct and the tests genuinely discriminate: all six mutations I applied earlier (dropping the division, transposing the extracted block, swapping i/j in the slice, `MacroMrkvArray[j,i]`, builder using `CondMrkvArrays[j][i]`, removing the zero guard) are caught by the existing suite.

Round trips are exact at every size tried, including non-square `M != N` and the degenerate `M=1`/`N=1` cases: max error 5.55e-17 on the recovered conditionals and exactly 0.0 on the rebuild. The hardcoded `2` at the KS call site is structurally correct rather than coincidental, since `make_MrkvArray` builds a fixed 4x4 with `StateCount = 2` and the `BU, BE, GU, GE` ordering is exactly `combined = N*macro + micro`.

I pushed one commit (`372febcb`) adding the two validations from my review, since the helper was assuming a structure it never checked:

- `MrkvIndArray.shape != (M*N, M*N)` now raises `ValueError` naming actual and required shape. Previously a wrong `N` returned well-formed garbage: on a 12x12 array whose true `N` is 4, `N=3` gave nine (3,3) blocks with row sums 0.672 / 0.521 / 0.978 and no error.
- Each block is checked against `np.allclose(block.sum(axis=1), MacroMrkvArray[i,j])`, which is necessary and sufficient for the recovered block to be row-stochastic, raising with the offending `(i,j)`, row sums and macro probability. Placed before the `p_macro > 0` branch so it also catches a nonzero block under a zero macro transition. Previously a non-product-form input came back as non-stochastic matrices silently, and a downstream `probs / probs.sum()` would have produced a wrong distribution rather than a crash.

Both are `ValueError` rather than `assert`, since asserts vanish under `python -O`. Two tests added, and they fail against the unvalidated version (3 failed, verified by reverting only the source file). Round trips re-confirmed exact with validation in place; zero-macro blocks still accepted and still round-trip exactly. 19 tests pass, `ruff@0.11.8` clean, and the `KrusellSmithEconomy` import moved to module top.

Left alone deliberately, both yours to decide: the `p_macro == 0` case returning `np.zeros((N,N))` rather than `np.full((N,N), 1.0/N)` (both round-trip identically, but zeros is the one choice that is not row-stochastic, which matters for the #1801 consumer), and the `self.MacroMrkvArray = MrkvAggArray` aliasing that makes one test assertion unfalsifiable.

**Merge order: this must land before #1801**, which consumes `extract_cond_mrkv_arrays`.

PR #1797, opened 2026-08-11
Rename the KS aggregate Markov state key: Mrkv -> MrkvAgg (breaking, KS only)

llorracc:

**Deliberately breaking (Krusell-Smith only), labeled as such:** renames the aggregate Markov state key on `KrusellSmithType` and `KrusellSmithEconomy` from `"Mrkv"` to `"MrkvAgg"` — across `shock_vars_`, `shocks`, `sow_vars`, `track_vars`, `sow_init`, and therefore `economy.history`.

Why: the hierarchical-Markov refactor that follows gives KS agents an *idiosyncratic* Markov state, which needs the conventional agent-level `"Mrkv"` name — today the aggregate state squats on it, and the collision is exactly the kind of silent-aliasing hazard we want out of the way first. `AggShockMarkovConsumerType` and the Cobb-Douglas Markov economies are deliberately untouched.

Scope discipline: exactly 7 code lines (verified against a class-boundary map so no non-KS `"Mrkv"` is touched) + the three stale lines in `docs/guides/krusell_smith.md`. In-repo tests and examples have no KS-scoped consumers of the old key (grep-verified); the KS suite asserts on `EmpNow` and economy dynamics, unaffected. Migration for downstream KS code is a one-line key swap, stated in the changelog's Major/breaking entry.

- [x] Changelog: **Major (breaking)** entry with migration note · [x] Docs updated


alanlujan91 (approved):

alanlujan91:
Reviewed as part of stack #1803 (#1797 -> #1798 -> #1801); the full review is on [#1801](https://github.com/econ-ark/HARK/pull/1801). Nothing blocking originates in this PR.

**The rename is complete and Krusell-Smith is numerically untouched.** I swept every quoted `"Mrkv"` literal repo-wide with `rg --no-ignore --hidden` including `.ipynb` JSON, since a missed reader here is a silent `KeyError` or an absent variable rather than a compile error. Every surviving hit is either a deliberately untouched class (`ConsMarkovModel`, `ConsRepAgentModel`, `AggShockMarkovConsumerType`, `CobbDouglasMarkovEconomy`) or the new per-agent combined state. `mill_rule` returns positionally so the `sow_vars` reorder stays consistent, the KS guide is updated, and the notebook only touches `MrkvNow_hist`/`MrkvInit`.

Results are bit-identical to main: `AFunc[0].slope` 1.0441663924068658 both sides, along with `AFunc[1].slope`, both intercepts, `mean(Urate)` 0.07032727272727274, `mean(Mnow)` 6.23198472436914 and `cFunc[0](10.0, MSS)` 1.2427979512517826. The only difference is the history key, which is the intended break.

One thing that is this PR's to decide: **`AggIndMarkovConsumerType` was released in 0.17.2** and the stack removes it with no alias. It is declared in the changelog and HARK is 0.x, so entirely your call, but it is a released name rather than a dev-only one.

This is the bottom of the stack, so under the repo's merge-whole convention it lands with #1798 and #1801 rather than alone. I had approved it earlier for its own content; that still stands, and the stack is now blocked only on two findings in #1801 (a zero-probability macro block handing agents uninitialized memory, and the general `[i][j]` branch having no test coverage). Fixes in progress.

PR #1799, opened 2026-08-11
Wire the dual-measure base-draw cache through IndShock/Markov get_shocks (stacked)

llorracc:
> ⚠️ **CI note:** if `build (ubuntu-latest, 3.12)` shows red here, it is the **pre-existing breakage on unmodified `main`** (scipy 1.18.0 deepcopy regression) — fix in **#1788**; merge that first.

> 🔗 **Stacked on the income_shuffle chain (#1789 → #1792) and contains #1783's `HARK.dual_measure` commit until it merges.** Merge order: #1783, #1789, #1792, then this — the diff shrinks to just the wiring as parents land.

Wires `HARK.dual_measure`'s base-draw cache through `IndShockConsumerType.get_shocks` and `MarkovConsumerType.get_shocks`. Under the opt-in `_cache_base_shock_draws` flag (set by `DualMeasureMixin` users), the income-shock uniforms are recorded for Q-CDF inversion — keyed by the raw `t_cycle` value / `(t, j)` / `"newborn"`, matching the module's consumption protocol. Without the wiring the module already degrades gracefully (its Q side draws its own events); with it, the Q measure reuses the P-pipeline's exact uniforms, which is what the variance-reduction guarantees assume.

**Non-disruption, proven two ways:**
1. The cache arm draws the *same* uniforms and applies the *same* `searchsorted` inversion as the default `draw_events` path — so even with the flag **on**, the P-stream is bit-identical (tested: identical histories flag-on vs flag-off; no attributes appear on the default path at all).
2. An end-to-end equality proof: with a degenerate permanent shock (Q ≡ P distribution), the Q side inverting the recorded uniforms reproduces the P shocks **exactly**.

Deliberate deviations from the long-running source branch: recording is fully conditional (the source assigned unconditionally); the unused `_newborn_mask` is dropped; shuffle-arm `None`-markers are omitted (missing key and `None` are treated identically by the module); and the wiring is re-derived against main's post-#1760 loop structure — the source branch's stale loop-header revert is not carried. Known module-side limitation flagged for a follow-up: the Markov Q-newborn path uses pre-#1760 newborn semantics internally (independent of this wiring).

Full `test_IndShockConsumerType` + `test_ConsMarkovModel` suites green (37).

- [x] Tests (P-stream invariance + end-to-end equality + module suite) · [x] Documented · [x] CHANGELOG.md updated


alanlujan91:
Chris, reviewing this stack as a unit (#1789 then #1792 then #1799), evaluated at #1799's head merged with current main. Default-off behaviour is clean and the wiring is correct where it runs. Three things block, and the first is the one I would want to know about.

## The cache is never switched on

`setup_Q_measure()` is the only enable path. It sets `IncShkDstn_Q` and `dual_measure = True`, but it never sets `_cache_base_shock_draws`. Nothing under `HARK/` sets it; the only assignments in the tree are in `tests/test_dual_measure.py:127,148`. So `_draw_Q_shocks_indshock` falls through to `IncShkDstnQ.draw_events(N)` and the two measures draw independently, which is the situation this PR exists to fix.

`IncShkDstn_Q` is also absent from `self.distributions`, which holds `['IncShkDstn','PermShkDstn','TranShkDstn','kNrmInitDstn','pLvlInitDstn']`, so `reset_rng` never rewinds it. Measured `corr(pLvl_P, pLvl_Q)` across repeated `initialize_sim()`, AgentCount=300:

| run | as shipped | with `_cache_base_shock_draws=True` set by hand |
|---|---|---|
| 0 | +0.9686 | +0.9686 |
| 1 | -0.1232 | +0.9686 |
| 2 | -0.0426 | +0.9686 |
| 3 | -0.0437 | +0.9686 |
| 4 | -0.0071 | +0.9686 |

The first run looks coupled because `make_Q_measure_dstn` clones the seed, so both streams start in lockstep; from the second run onward P rewinds and Q does not. That is the same decay pattern found in #1783's review, and this PR is the fix for it. One line in `setup_Q_measure` closes it.

## `income_shuffle=True` silently voids the cache

The dispatch is `if income_shuffle: ... elif _cache: ...`, so with both enabled `base_draws_dict` stays empty and `self._base_shock_draws = {}`. Measured `corr(PermShk_P, PermShk_Q) = -0.0000`, with no warning. Since #1789 and #1799 are in the same stack, a user has every reason to expect both to work together. Either raise on the combination or record the shuffle path's implied uniforms.

## Nothing tests `income_shuffle`

Replacing both `if getattr(self, "income_shuffle", False):` guards with `if False:`, i.e. disconnecting the feature entirely:

| mutation | result |
|---|---|
| `income_shuffle` off, `ConsIndShockModel` | **60 passed, 0 failed** |
| `income_shuffle` off, `ConsMarkovModel` | **60 passed, 0 failed** |
| `_cache_base_shock_draws` off, both | 58 passed, **2 failed** |

So #1799's half is genuinely covered and #1789's and #1792's half is not. The cause is a tolerance rather than a missing test: `test_shuffle_empirical_frequencies` uses `atol=0.02` at N=5000, but the measured maximum per-atom frequency deviation over 60 seeds is 0.00946 under iid draws and 0.00106 under shuffle. The tolerance sits 2.1 times above the noise it is meant to detect, so it cannot fail whichever code path runs. `atol=0.003` separates them cleanly. The remaining three tests are shape-and-positivity smoke.

## Non-blocking

**`_base_shock_draws` is never cleared.** It survives `initialize_sim()` and survives the flag being turned off; measured keys `{0, 'newborn'}` still present after an uncached run. The reader then consumes stale uniforms from an earlier period. If `AgentCount` changed between runs it raises (`cannot assign 300 input values to the 120 output values`), and if it did not, it is silent. Clearing it in `initialize_sim` fixes both.

**Markov newborns are asymmetric.** IndShock caches newborns under the key `"newborn"`; `ConsMarkovModel.py:1062-1078` has no `_cache` branch at all, and the measured keys are `{(0,0),(0,1)}` with no `'newborn'`. Nothing breaks, because the Q-reader treats Markov newborns deterministically, but the two models now differ in whether newborn shocks are P/Q coupled, which is surprising in a stack whose point is that they behave alike.

**The docstring overstates the small-N case.** "Eliminating cross-sectional sampling noise" holds once N comfortably exceeds the atom count, not before. 3000 repetitions against a 56-atom `IncShkDstn`:

| N | max z on E[count] | SD(mean TranShk) shuffle/iid | SD(mean PermShk) shuffle/iid |
|---|---|---|---|
| 5 | 2.69 | 0.795 | 0.504 |
| 25 | 2.09 | 0.362 | 0.233 |
| 100 | 1.34 | 0.061 | 0.111 |
| 1000 | 0.15 | 0.019 | 0.032 |

Unbiased at every N, which is the important part, but the variance gain is only about 20% at N=5. Markov makes small groups routine: at `AgentCount=20` the smaller Markov group measured 4 agents against 56 atoms.

**Period indexing.** `t_key = t - 1 if self.cycles == 1 else t` in `_draw_Q_shocks_indshock` disagrees with the writer's unconditional `t - 1`. Both branches run at `T_cycle=1` because they land on the same index, so nothing currently fails, but a `T_cycle>1` lifecycle model with `cycles=0` would give P and Q different periods. I did not construct one.

## Verified correct

**#1799's written symbol matches #1783's reader exactly**: `self._base_shock_draws` at `ConsIndShockModel.py:2298` and `ConsMarkovModel.py:1078` against `getattr(self, "_base_shock_draws", {})` at `dual_measure.py:286,334`. No near-miss, which was worth confirming because a one-character mismatch would leave #1783 permanently inert with no error anywhere.

**Only two copies of the remainder-allocation logic exist in this tree**, and both are correct: `discrete.py:243` is the #1808 systematic-sampling fix, verified unbiased by measurement, and `ConsIndShockModel.py:1444` in `_sim_death_shuffled` is a two-outcome Bernoulli remainder that is correct by construction. There is no `MarkovProcess._draw_shuffled` and no `eps = 1.0 /` anywhere here, so the copy found in #1786 is not reachable from this stack.

**Default-off bit identity against main: max absolute difference 0** across twelve fingerprints, covering IndShock and Markov `mNrm`/`cNrm`/`pLvl`/`PermShk`/`TranShk` histories plus post-run positions of both `self.RNG` and `IncShkDstn._rng`.

**Cache on/off P-stream identity: 0.000e+00** on `PermShk`, `TranShk`, `mNrm`, `cNrm`, with identical RNG position, so the CHANGELOG's claim that the P pipeline is untouched holds.

**The #1808 fix is intact here.** `draw(N, shuffle=True)` over 40000 repetitions at `pmv=[.1,.2,.3,.4]`, max `|E[count] - N*pmv|`: 0.0000 at N=0, 0.0065 at N=1, 0.0012 at N=2, 0.0022 at N=3, 0.0043 at N=7, 0.0008 at N=13, against a Monte Carlo standard error of about 0.0025.

Full suite after merging main: 790 passed, 0 failed.

## Project rules

Non-ASCII characters on new lines (excluding the pre-existing Greek and thorn notation): `ConsIndShockModel.py:2208,2209,2247,2258,2286`; `ConsMarkovModel.py:1032,1045,1071`; `dual_measure.py:275`; `test_dual_measure.py:4,5,6,9,12,141`; `test_IndShockConsumerType.py:1028,1056,1061,1070,1083,1093,1096,1103,1109,1110,1137,1138,1178,1186`; `docs/CHANGELOG.md:31,32`. A function-level import at `test_IndShockConsumerType.py:1004`.

## On #1794

`income_shuffle` matches exactly what `core.py:1851` reads, so merging #1789 makes half of #1794 live. `markov_shuffle` does not exist anywhere in this tree, so the other half stays inert until #1793 lands.

Attribution: the cache-never-armed issue, the shuffle interaction, the stale cache and the Markov newborn gap are #1799. The missing `income_shuffle` coverage is #1789 and #1792.

alanlujan91:
Fixes pushed as `40e3862c` (fast-forward over `5f9c0e5d`, no force), with the companion test fix on `shuffle-b1-income-markov` as `c92a7740`, which this tip merges.

**The cache now arms itself.** `setup_Q_measure()` sets `_cache_base_shock_draws` and registers `IncShkDstn_Q` in `self.distributions`. Both were needed and each is caught by a separate test: arming alone gives a coupled first run that still decays, because `reset_rng` was not rewinding the Q distribution. Re-measured `corr(pLvl_P, pLvl_Q)` across repeated `initialize_sim()`, nothing set by hand:

| run | before | after |
|---|---|---|
| 0 | +0.9686 | +0.9686 |
| 1 | -0.1232 | +0.9686 |
| 2 | -0.0426 | +0.9686 |
| 3 | -0.0437 | +0.9686 |
| 4 | -0.0071 | +0.9686 |

**Worth calling out explicitly**: this is a behaviour change for anyone currently getting the uncoupled Q draws, even though it restores what the module docstring already claimed. `_cache_base_shock_draws = False` is now documented as the way back.

**`income_shuffle=True` alongside the cache** emits a RuntimeWarning naming both flags, on IndShock and Markov alike, rather than silently leaving `_base_shock_draws = {}`.

**`_base_shock_draws` is cleared in `initialize_sim`.** It previously survived both the reset and the flag going False, so the reader consumed stale uniforms; the AgentCount 300 -> 120 case that raised `ValueError: cannot assign 300 input values to the 120 output values` now runs.

Mutation results, each verified to have actually applied before the run:

| mutation | failures |
|---|---|
| `_cache_base_shock_draws = True` -> `False` in `setup_Q_measure` | 2 |
| drop the `IncShkDstn_Q` registration only | 1 |
| remove both shuffle-conflict warnings | 1 |
| remove the `_base_shock_draws` clear in `initialize_sim` | 2 |
| `income_shuffle` dispatch -> `if False:`, IndShock | 1 |
| same, Markov | 1 |

That last pair was **0 and 0** before. The cause was a tolerance rather than a missing test: `test_shuffle_empirical_frequencies` used `atol=0.02` while measured iid noise is 0.01306 and shuffle is 0.00214, so it could not fail whichever path ran. Now `atol=0.005`, plus a new Markov test that compares the shuffled and iid paths against each other within each state rather than against a fixed bound, since a fixed bound is what made the old one blind.

Default-off bit identity against main is unchanged: all 12 fingerprints identical, checked separately on both commits. Full suite 794 passed, 0 failed. `ruff@0.11.8` `format --check` and `check` clean, which also fixes a real format failure that was on this branch as pushed, independent of any merge.

One naming question for you: `warn_if_shuffle_voids_base_draw_cache` is a new module-scope name in `ConsIndShockModel`, imported by `ConsMarkovModel`, so the warning lives in one place rather than two. Happy to make it private or move it.

Three things left deliberately for you, all judgement rather than defect: the **Markov newborn asymmetry** (IndShock caches newborns under `"newborn"`, `ConsMarkovModel.py:1062-1078` has no `_cache` branch, so Markov newborns are not P/Q coupled); the **small-N docstring claim**, since "eliminating cross-sectional sampling noise" holds once N exceeds the atom count, which is 56 in the default calibration, and the measured variance gain at N=5 is only about 20%; and the **`t_key = t - 1 if self.cycles == 1 else t`** indexing, which disagrees with the writer's unconditional `t - 1` and is unreachable at T_cycle=1 because both collapse to index 0. I did not construct a failing T_cycle>1 case, so I left it.

alanlujan91 (approved):

PR #1798, opened 2026-08-11
Hierarchical-Markov rewrite: AggIndMrkvConsumerType(MarkovConsumerType); KS reparents (breaking)

llorracc:
> ⚠️ **CI note:** if `build (ubuntu-latest, 3.12)` shows red here, it is the **pre-existing breakage on unmodified `main`** (scipy 1.18.0 deepcopy regression) — root cause and one-line fix in **#1788**; merge that first.

> 🔗 **Stacked on #1797** (the `MrkvAgg` rename) and contains #1796's commit (the hierarchical helpers) until those merge — the diff shrinks to just the rewrite as its parents land. Merge order: #1796 → #1797 → this.

**The structural piece of the hierarchical-Markov refactor, deliberately labeled breaking:**

- `ConsAggIndMarkovModel` is rewritten: **`AggIndMrkvConsumerType(MarkovConsumerType)`** replaces the former `AggIndMarkovConsumerType(AgentType)`. Hierarchical macro+micro Markov states are carried in `shocks["Mrkv"]` (combined index `= N * macro + micro`); `get_macro_markov_states` resolves `EconomyMrkvNow` → `shocks["MrkvAgg"]` → derived-from-combined; `get_micro_markov_states` is the overridable per-cell draw; with the hierarchical counts unset the class degrades to a pure `MarkovConsumerType`.
- **The old name is removed, not aliased.** The old class's contract (`MrkvCombined`, `AgentType` base, different shock-key semantics) differed enough that a compatibility alias would produce *silent* breakage in downstream code written against the old contract; a loud `ImportError` with a changelog migration note is the honest failure mode.
- **`KrusellSmithType` reparents onto the new class**, in the same commit as its no-op `sim_death` override — the two are inseparable: without the override, the `MarkovConsumerType` base would consume an extra RNG draw per period and shift the KS stream.
- The `from_combined` helpers preserve the scalar path (`int` in → `int` out).

**Non-disruption evidence:** the untouched, seeded `tests/ConsumptionSaving/test_ConsAggShockModel.py` passes in full on this branch (18 with the new unit tests) — the KS default simulation path is unchanged. Deliberately *not* included (deferred to their own PRs per our risk assessments): the `markov_shuffle` branch of `get_micro_markov_states` (stacks on the shuffle family), a +372-line "temporary reference implementation" from the source branch, and a script-style example test.

- [x] Tests (full KS suite + hierarchical unit tests) · [x] numpydoc · [x] CHANGELOG.md (Major/breaking entries with migration notes)

PR #1801, opened 2026-08-11
Add the markov_shuffle branch to hierarchical get_micro_markov_states (stacked; completes the series)

llorracc:
> ⚠️ **CI note:** if `build (ubuntu-latest, 3.12)` shows red here, it is the **pre-existing breakage on unmodified `main`** (scipy 1.18.0 deepcopy regression) — fix in **#1788**; merge that first.

> 🔗 **Stacked on #1798** (the hierarchical rewrite) and carries the commits of **#1786**, **#1793**, and **#1794** until they merge — the diff shrinks to just this wiring as parents land. This is the **last construction PR of the series**: with it, every piece of the source branch's machinery has an owned, review-scoped path to main.

Restores the deliberately-deferred `markov_shuffle` branch to `AggIndMrkvConsumerType.get_micro_markov_states`: quota-exact micro-state transitions per (macro, source-micro) cell via `MarkovProcess.draw(shuffle=True)`, supporting both the simple and the general `[i][j]` conditional formats, with `balanced_transitions` (systematic sampling by pLvl within each cell — pLvl deliberately, never wealth). The default path — iid `RNG.choice` per cell — is textually unchanged.

Tests: the dedicated suite written alongside the original implementation (hierarchical transition counts under shuffle, CRN across conditional arrays, `make_shock_history` flag toggle/restore), passing as-is.

- [x] Tests · [x] numpydoc · [x] CHANGELOG.md updated


alanlujan91:
Note for whoever merges this one: **#1796 has landed** (`25eb03c39`), so `extract_cond_mrkv_arrays`, which this PR consumes, is now on main. That dependency is satisfied.

It was worth stating because #1796 was registered as standalone with base `main` while being a producer for this PR, so nothing in the stack map or the merge box connected them. The direction was the harmless one, since merging standalone PRs first satisfies a producer-below edge for free, but the constraint was real and undeclared in either PR body.

Note also that #1796 gained input validation on that helper: it now raises `ValueError` if `MrkvIndArray.shape != (M*N, M*N)`, or if any block's row sums do not match `MacroMrkvArray[i,j]`. If this PR ever calls it with a hierarchical array that is not exactly of product form, that is now a loud failure rather than silently non-stochastic output.

One thing left open there that bears on this PR specifically: when a macro transition has probability zero, the helper returns `np.zeros((N,N))`. It round-trips correctly, but zeros is not a row-stochastic matrix, so a consumer here that normalizes `probs / probs.sum()` gets `nan` rather than a distribution. `np.full((N,N), 1.0/N)` round-trips identically and avoids that. Worth checking whether this PR's `get_micro_markov_states` path can reach a zero-probability macro transition.

alanlujan91:
Chris, reviewing this stack as a unit (#1797 then #1798 then #1801, evaluated at #1801's head merged with current main), since the rename and the reparent are only meaningful together. The two things I most expected to break are clean; the weakness is in the feature #1801 exists to add.

## Krusell-Smith is bit-identical to main

This is the check that mattered most for a breaking rename plus a reparent, and it passes to the last bit:

| | origin/main | this stack |
|---|---|---|
| `AFunc[0].slope` | 1.0441663924068658 | 1.0441663924068658 |
| `AFunc[1].slope` | 1.0498010085000264 | 1.0498010085000264 |
| `mean(Urate)` | 0.07032727272727274 | 0.07032727272727274 |
| `mean(Mnow)` | 6.23198472436914 | 6.23198472436914 |
| `cFunc[0](10.0, MSS)` | 1.2427979512517826 | 1.2427979512517826 |

Both intercepts and `mean(Aprev)` likewise. The only difference is the history key, `Mrkv` to `MrkvAgg`, which is the intended break. Full suite on the merged tree: 793 passed, 1 skipped.

**The rename is complete.** I swept every quoted `"Mrkv"` literal repo-wide with `rg --no-ignore --hidden`, including `.ipynb` JSON, since a missed reader would be a silent `KeyError` or an absent variable rather than a compile error. Every surviving hit is either deliberately untouched (`ConsMarkovModel`, `ConsRepAgentModel`, `AggShockMarkovConsumerType`, `CobbDouglasMarkovEconomy`) or the new per-agent combined state. `mill_rule` returns positionally so the `sow_vars` reorder stays consistent, the KS guide is updated, and `KrusellSmithType.ipynb` only touches `MrkvNow_hist`/`MrkvInit`.

**The reparent is inert for KS.** The MRO becomes `KrusellSmithType -> AggIndMrkvConsumerType -> MarkovConsumerType -> IndShockConsumerType -> PerfForesightConsumerType -> AgentType`. I enumerated inherited-but-not-overridden methods programmatically rather than reading the class, so nothing could hide: KS overrides all 13 hot-path methods, `pre_solve` is overridden so `check_markov_inputs` never fires, and the 15 newly-inherited methods are unreachable in the KS workflow. The bit-identical numbers are the evidence.

**The hierarchical math is right where exercised.** Known-answer check at M=3, N=2 with a structural zero in the macro chain, 4000 agents: empirical transition frequencies match `CondMrkvArrays[1][2]` to 4e-4.

## Blocking: a zero-probability macro block hands agents uninitialized memory

`ConsAggIndMarkovModel.py:287,309,335`. `new_micro = np.empty(...)` combined with `if cond[mi,:].sum() <= 0.0: continue` leaves every agent in that cell holding whatever bytes were in the allocation. Measured at M=3, N=2 with `macro[0,2] = 0`, previous macro 0 and next macro 2, 4000 agents:

```
drawn micro state min/max:  2 .. 131874259223184     (valid range [0, 1])
out of range:               4000 of 4000
unique values:              [2, 3, 1145147440, 131874259223184]
```

Those values go straight into `shocks["Mrkv"] = N*macro + micro` and index solution arrays. It is reachable whenever `(macro_prev, macro_next)` is a zero cell, and the agent cannot rule that out, because the macro state arrives from outside it (the economy's sow, or `MrkvPrbsInit` at t=0) and need not respect its own `MacroMrkvArray`.

The telling contrast is that the default branch fails **loudly** on the same input: `probs / probs.sum()` at line 367 raises `ValueError: Probabilities contain NaN`. So the shuffle branch is strictly less safe than the path it replaces.

Worth stepping back: this is the third `np.empty`-derived silent-corruption bug in this PR series. #1783 records uninitialized `bNrm`/`kNrm` into `history_Q`, and #1793 sorted on a `state_now["pLvl"]` that `_sim_period_prologue` had already blanked. In all three the wrong values look plausible, which is what makes them dangerous. `np.full(..., np.nan)` costs nothing and converts every one of them into a loud failure.

## Blocking: the general `[i][j]` branch has no test coverage

Mutation results against `test_ConsAggIndMarkovModel.py` plus `test_hierarchical_mrkv_helpers.py`, baseline 12 passed:

| mutation | tests failing |
|---|---|
| disable the whole `markov_shuffle` branch | 2 |
| delete the general-format branch (`if general:` -> `if False:`) | **0** |
| transpose the index: `CondMrkvArrays[mp_i][mn_i]` -> `[mn_i][mp_i]` | **0** |
| remove the zero-row guard at :309/:335 | **0** |
| ignore `balanced_transitions` `sort_key` at :319,:345 | **0** |
| drop the `shocks["MrkvAgg"]` read in default `get_macro_markov_states` | **0** |

The transpose is the one that worries me: it is a real composition error that would silently swap source and destination macro conditioning, and nothing catches it. The cause is that both existing shuffle tests use the **simple** format, one with `M=1` and the other with all agents in macro 0, so the general path never runs. A test needs list-of-list `CondMrkvArrays`, at least two distinct `(macro_prev, macro_next)` pairs live simultaneously, and asserted per-cell counts.

## Merge-order hazard, please do not resolve this one by taking a side

`tests/ConsumptionSaving/test_hierarchical_mrkv_helpers.py` is an add/add conflict against main, and `ConsAggIndMarkovModel.py` conflicts on `extract_cond_mrkv_arrays`. This branch carries the **pre-#1796** helper, so a "take ours" resolution silently deletes the shape check, the row-sum check, and the two `test_extract_rejects_*` tests that landed with #1796. Resolve as a union; I did, and the result is green.

## Also worth fixing

**Import placement.** The branch moves an import back down relative to main: `test_hierarchical_mrkv_helpers.py:75, 90, 107` and `test_ConsAggIndMarkovModel.py:151`. Project rule is module top.

**`balanced_transitions` degrades silently** at `:316-319` and `:342-345` when `state_now` has no `pLvl`: `sort_key` stays `None` and the run proceeds as a plain shuffle. Combined with that path being untested, a user can believe balancing is on when it is not.

**The RNG stream is data-dependent under `markov_shuffle`.** `MarkovProcess(cond, seed=int(self.RNG.integers(...)))` is constructed inside the `(pair, micro_prev)` loops, so the number of parent draws varies with the realized state distribution. That undercuts common random numbers across scenarios that shift the distribution, which is what the CRN test is for. Drawing a fixed grid of sub-seeds up front would make it distribution-invariant.

**API asymmetry.** `extract_cond_mrkv_arrays` and `KrusellSmithEconomy` both emit the general `[i][j]` format, but the default non-shuffle `get_micro_markov_states` raises `NotImplementedError` on it at `:355-360`. It fails loudly so it is not a bug, but the docstring should say the format is only consumable with `markov_shuffle=True`.

**`AggIndMarkovConsumerType` was released in 0.17.2** and is removed here with no alias. It is in the changelog and HARK is 0.x, so your call; I flag it only because it is a released name.

**`KrusellSmithType.initialize_sim` calls `AgentType.initialize_sim(self)` directly**, skipping two MRO levels. That is what preserves the bit-identical numbers today, and it will silently skip anything later added to the intermediate class.

**15 Unicode em/en dashes on added lines**, 6 in `ConsAggIndMarkovModel.py` and 4 in `tests/test_distribution.py`. Repo rule is ASCII hyphen.

## What would have caught a problem and did not

I compared the full `economy.history`, both `AFunc` slopes and intercepts, and the agent's `cFunc` value against main, so a rename that dropped a variable or shifted the sow ordering would have shown as a changed `Urate` or `Mnow` mean or a missing history key. I ran the whole suite rather than the KS subset, so a break in an unrelated `Mrkv` consumer would surface. On the reparent I enumerated the inherited method set programmatically, so a method silently picked up from `IndShockConsumerType` could not hide.

Note this branch also carries `d80be582`, `c7e5c058` and `b6c5b556` from the #1805 chain. The `balanced_transitions` and RNG-stream findings sit on the seam between the two stacks; I reviewed them as consumed here and left the distributions layer to the #1805 review, where a separate blocking issue was found.

Fixes for the two blocking items and the smaller ones are in progress and will be pushed to this branch.

alanlujan91:
## Merge-order hazard between this stack and #1805, please read before merging either

This branch carries **its own copy** of #1786's commits rather than descending from that branch (`git merge-base --is-ancestor origin/baseline-q1-distribution-draw-modes origin/hierarchical-markov-shuffle` returns false). One of those commits contains the biased leftover-slot allocator in `MarkovProcess._draw_shuffled`:

```
hierarchical-markov-shuffle   'eps = 1.0 /' present   <- biased version
shuffle-b2-markov-transitions  allocate_remainder_slots  <- fixed version
baseline-q1-distribution-draw-modes  'eps = 1.0 /' present
main                           neither
```

The fix for that allocator was applied on #1793's branch, at the top of stack #1805. It is not on this one.

**Consequence: merging this stack before #1805 lands the biased allocator on main.** Measured, E[count] over 20000 repetitions at row `[0.7, 0.2, 0.1]`, N=13: `[9.696, 2.205, 1.099]` against a target of `[9.1, 2.6, 1.3]`. That row is analytic rather than noisy, and the bias matters more under shuffle than under iid draws precisely because the variance is reduced: the long-run state-0 share bias goes from about 4% of the period-to-period standard deviation to roughly 200% of it.

**Suggested order: merge stack #1805 first**, then rebase this branch onto the resulting main and confirm `HARK/distributions/base.py` contains `allocate_remainder_slots` and no `eps = 1.0 /` before merging. If it is merged the other way round, the resolution on `base.py` decides whether the bug returns, and nothing in CI would catch it: the existing `test_shuffle` draws N=1000, where the correct and the biased allocators agree exactly.

Full details of the allocator issue are on [#1786](https://github.com/econ-ark/HARK/pull/1786) and [#1793](https://github.com/econ-ark/HARK/pull/1793).

---

Separately, the blocking items from my review of this stack are fixed and pushed as `2610e1cb`.

**Uninitialized memory.** Both shuffle branches now raise `ValueError` naming the offending `(macro_prev, macro_next)` cell instead of `continue`, and the buffer is `np.full(..., _UNSET_MICRO)` with an end-of-method check. Not `np.zeros`, which would silently place every affected agent in micro state 0; and not `np.nan`, since the buffer is integer dtype, so a negative sentinel is the integer equivalent.

**The general `[i][j]` branch now has coverage**, 7 new tests. Mutation results, before and after:

| mutation | was | now |
|---|---|---|
| transpose `CondMrkvArrays[mp_i][mn_i]` | 0 | 3 |
| delete the general-format branch | 0 | 5 |
| drop the zero-row guard (general) | 0 | 1 |
| drop the zero-row guard (simple) | 0 | 1 |
| ignore the `balanced_transitions` sort key | 0 | 1 |
| set the sentinel to 0 instead of -1 | n/a | 7 |

Plus warnings when `balanced_transitions` finds no `pLvl`, three imports moved to module top, and 15 Unicode dashes replaced on branch-added lines only.

**Krusell-Smith is still bit-identical to main**: `AFunc[0].slope` 1.0441663924068658 unchanged, along with both intercepts, `mean(Urate)`, `mean(Mnow)` and `cFunc[0](10.0, MSS)`. No golden moved, so none needed updating. Full suite 800 passed, 1 skipped.

One behaviour change worth your view: the new `ValueError` fires for anyone whose `MrkvPrbsInit` places agents in a macro state the economy then contradicts. At `t_sim == 0` those draws are discarded anyway, so such a user previously got harmless garbage and will now get a hard failure. Raising still seems right, since the alternative is inventing a distribution the model says cannot exist, but it is a real edge for a hierarchical model with a sparse macro chain.

Your rebase also absorbed #1796, so the merge hazard I flagged earlier for `extract_cond_mrkv_arrays` is gone: the validating version is present and both `test_extract_rejects_*` tests pass.

alanlujan91:
Follow-up, pushed as `2d19cdb8` (fast-forward over your `e0877637`, no force at any point).

**The dropped sort-key fix had a second half nobody had found.** `4c57e664` corrected `balanced_transitions` in `ConsMarkovModel.py`, but `AggIndMrkvConsumerType.get_micro_markov_states` reads a sort key at two more sites, and those still used `state_now`. So the same defect was live in a second file, including in the version I pushed an hour earlier.

The mechanism, measured during a live 4-period simulation by capturing at the moment `get_markov_states` runs:

| t_sim | `state_now["pLvl"]` range | `state_prev["pLvl"]` range |
|---|---|---|
| 0 | [0.0, 6.64372926891753e-310] | [1.0, 1.0] |
| 1 | [0.0, 6.6437292688859e-310] | [0.859, 1.178] |
| 2 | [0.0, **1.1780702264015428**] | [0.738, 1.388] |
| 3 | [5.42e-315, 0.02] | [0.685, 1.511] |

Row 2 is why this survives review: the uninitialized bytes are a leaked copy of the *previous period's* maximum, so the value is finite, positive and entirely plausible for a `pLvl`.

It also invalidated the guard I added earlier. I had tested `"pLvl" not in self.state_now`, but `_sim_period_prologue` replaces the *values* while keeping the keys, so that condition is False every period. The warning could never fire while the sort silently used uninitialized memory. It now tests `state_prev.get("pLvl") is None`.

**Why the original fix vanished without anyone noticing: it had no test.** I reverted it on the current tree and ran everything that plausibly covers it, `test_IndShockConsumerType.py` + `test_distribution.py` + `test_ConsMarkovModel.py`: **106 passed, 0 failed**. A one-token revert of a real correctness fix was invisible to CI, which is exactly how a rebase discards it silently. There is now `testBalancedSortKeyDuringSimulation`, which drives a real simulation and asserts the sort key handed to `MarkovProcess.draw` equals `state_prev["pLvl"]` period by period.

Final mutation table against the shipped source:

| mutation | before | now |
|---|---|---|
| `ConsMarkovModel` sort key back to `state_now` | 0 | 1 |
| hierarchical `pLvl_prev` back to `state_now` | n/a | 2 |
| transpose `CondMrkvArrays[mp][mn]` | 0 | 3 |
| delete the general-format branch | 0 | 6 |
| drop the zero-row guards | 0 | 1 / 1 |
| ignore the balanced sort key | 0 | 1 |
| sentinel -1 -> 0 | n/a | 8 |

Two caveats I would rather state than bury. A companion heuristic test that rejects non-finite or sub-1e-6 sort keys **did not catch** the `state_now` mutation, because the recycled memory in that run was finite and O(1) — the row-2 case above. Only the exact-match test killed it; the heuristic is kept for the denormal case but is not coverage. And the sentinel-fill mutation fails 0 tests by construction: with the raises in place nothing reaches the drawing loops unwritten, so that check has no live trigger. The sentinel-to-zero mutation failing 8 tests is what demonstrates the instrument is wired correctly.

**Krusell-Smith remains bit-identical to main**: `AFunc[0].slope` 1.0441663924068658 and every other figure unchanged, so no golden needed touching. Full suite 803 passed, 1 skipped. `ruff@0.11.8` clean.

Standing recommendation, since this is now the third instance in the series: `_sim_period_prologue` blanking float state arrays with `np.empty` rather than `np.nan` is the upstream cause of this entire bug class, and one line in `core.py` would convert all of it from silent to loud. That is shared simulation machinery well outside this stack, so it is untouched here; I opened [#1809](https://github.com/econ-ark/HARK/issues/1809) for it with the three measured instances.

alanlujan91 (approved):
Approving as part of stack #1803, reviewed and merged whole.

Two blockers were fixed. A zero-probability macro block handed agents uninitialized memory (measured micro states up to 1.3e14 for 4000 of 4000 agents, flowing into `shocks["Mrkv"]` and indexing solution arrays); both shuffle branches now raise `ValueError` naming the cell, with an `_UNSET_MICRO` sentinel and an end-of-method check. And the general `[i][j]` branch, the point of this PR, killed zero mutations; it now has 7 tests, with transposing the index failing 3, deleting the branch failing 6, and setting the sentinel to 0 failing 8.

The dropped `state_prev` sort-key fix is restored here and extended to the two hierarchical call sites that had the same defect, with a comment naming the mechanism so a future rebase dropping it is visibly wrong.

On merging: this branch carried a stale copy of the distributions layer including the pre-#1808 biased allocator. Resolved to main's version, verified `allocate_remainder_slots` present and `eps = 1.0 /` absent. KS remains bit-identical.

PR #1802, opened 2026-08-11
Make CubicHermiteInterp deepcopy/pickle independent of scipy spline internals

llorracc:
> **CI note:** based on post-#1788 main, so CI resolves scipy 1.17.1 and the full matrix should be green. The scipy-1.18.0 proof cannot run in CI anymore (the #1788 pin correctly refuses that version), so it is certified locally below, with a negative control.

Companion to #1788. That PR excludes scipy 1.18.0 (a permanent tombstone — that release will always be broken); this PR removes the *class* of failure so no future scipy internal change can reintroduce it.

## What

`CubicHermiteInterp` gains a `__getstate__`/`__setstate__` pair: the wrapped `scipy.interpolate.CubicHermiteSpline` (`_chs`) is excluded from serialized state and rebuilt deterministically from the stored `x_list`/`y_list`/`dydx_list` on restore. Because deepcopy routes through `__reduce_ex__` when no `__deepcopy__` is defined, the one dunder pair fixes **both** `copy.deepcopy` and `pickle`.

## Why

scipy 1.18.0 caches array-namespace **module** objects on every PPoly-family instance (`_xp`, `_xp_internal`; scipy #25489, from the Array-API adoption in scipy#25072), and module objects cannot be pickled or deepcopied. HARK hits this inside the solver itself: with `CubicBool=True`, `solve_one_period_ConsIndShock` → `MargValueFuncCRRA.__init__` → `deepcopy(cFunc)` → `TypeError: cannot pickle 'module' object`. That is exactly the 17-test failure set (`test_IndShockConsumerType`, `test_IndShockConsumerTypeFast`, `test_KinkedRconsumerType::test_cubic_and_vFunc`) seen on ubuntu CI legs before #1788.

Fixing the object rather than the call site makes every deepcopy site work at once (`MargValueFuncCRRA`, `AgentType.copy`, `distribute_params`, user code) and makes cubic solutions picklable for spawn-based multiprocessing — independent of scipy version, forever. Upstream's own fix (scipy#25500, merged, on the 1.18.1 milestone) will heal 1.18.0's successors; this PR removes HARK's dependency on scipy's private attribute layout either way.

The rebuild is deterministic: `CubicHermiteSpline(x, y, dydx)` from identical float arrays reproduces bit-identical coefficients (asserted in the new tests via `assert_array_equal` on `_chs.c` and on evaluations/derivatives across the lower-extrapolation, interior, and upper decay-extrapolation regions).

An alternative — making `_chs` a lazily rebuilt property that is never stored — was considered and rejected for this PR: it touches the runtime path, while the dunder pair is a pure serialization addition with zero behavior change for live objects.

## Validation (all local, py3.12 / numpy 2.4.6 / numba 0.66.0)

| arm | scipy | result |
|---|---|---|
| baseline: `test_interpolation.py` + `test_IndShockConsumerType.py` + `test_IndShockConsumerTypeFast.py` + `test_KinkedRconsumerType.py` | 1.17.1 | **167 passed** (+27 subtests), 77 s |
| certification: same four files | **1.18.0 (forced)** | **167 passed** (+27 subtests), 51 s |
| negative control: `git stash` the `interpolation.py` change only, rerun `test_KinkedRconsumerType.py` | 1.18.0 | **1 failed — `TypeError: cannot pickle 'module' object`** (then green again on `stash pop`) |
| full test suite `tests/` | 1.18.0 (forced) | **705 passed** (+27 subtests), 0 failed, 9 m 38 s |
| breadth audit: deepcopy+pickle of every scipy-backed object HARK persists (Normal/Lognormal/MeanOneLogNormal/Uniform/Weibull/Bernoulli frozen wrappers, `DiscreteDistribution`, `MarkovProcess`, `IndexDistribution`, unsolved agent, solved cubic agent, solved cubic `solution[0]`) | 1.18.0 | **ALL CLEAR** — the only 1.18.0-sensitive object HARK stores is the cubic spline this PR fixes |

The negative control proves the certification environment genuinely triggers the bug without the fix — the two arms differ only in the dunder pair.

New tests (run under whatever scipy CI resolves, so they guard the contract continuously):
- `tests/test_interpolation.py::testsCubicHermiteInterpSerialization` — deepcopy and pickle round-trips over three constructor variants (default NaN lower extrapolation, linear lower extrapolation, explicit intercept/slope limits), asserting bit-equality of evaluations, derivatives, and rebuilt spline coefficients, plus storage independence.
- `tests/ConsumptionSaving/test_IndShockConsumerType.py::testCubicSolutionSerialization` — end-to-end: a solved `CubicBool` agent survives deepcopy and its solution survives pickle, bit-exactly.

## Scope

Touches only `HARK/interpolation.py` (+25 lines), two test files, and one changelog bullet. No overlap with any other open PR; mergeable in any order.


alanlujan91:
Chris, the fix is correct and I have pushed one addition to this branch: it now also removes the `!=1.18.0` scipy exclusion.

## Verified

Round trip is bit-identical rather than approximately equal. Across three configurations (default, `lower_extrap=True`, and with `intercept_limit`/`slope_limit`), evaluated at 5001 probe points spanning below, inside and above the grid, `np.array_equal` holds for values, for first derivatives, and for `_chs.c`, under both `deepcopy` and `pickle`. Maximum absolute difference 0.0.

The state handling is the right shape: `__getstate__` is `self.__dict__.copy()` minus `_chs` rather than a hand-maintained key list, so nothing can silently fall out of sync as the class grows. The object's full `__dict__` is exactly `{x_list, y_list, dydx_list, n, coeffs, _chs}`, so there is nothing else to miss. Every remaining `_chs.` access is public API (`.c`, `derivative`, `antiderivative`, `integrate`, `roots`, `solve`), so the scipy-independence claim holds literally.

## Why I removed the pin here rather than in a follow-up

`requirements/base.txt` still carried `scipy>=1.10,!=1.18.0` from #1788, which was a stopgap to stop the py3.12 CI job failing on exactly the bug this PR fixes at its root. Left in place, it does more than sit there: it keeps CI from ever exercising the new tests. With scipy held below 1.18.0, `CubicHermiteSpline` has no `_xp`/`_xp_internal` module attributes at all, so `deepcopy` and `pickle` already succeed on unpatched code, and both new tests pass identically against `origin/main`'s `interpolation.py`. They would have provided no regression protection.

Checked under an actual scipy 1.18.0 environment with this branch installed:

| | result |
|---|---|
| `import HARK.utilities`, `from scipy.interpolate import interp1d` | OK, no DeprecationWarning |
| `test_interpolation`, `test_utilities`, `test_core`, `test_IndShockConsumerType` | 288 passed, 0 failed |
| `interp1d` deepcopy and pickle | both OK, evaluations match |
| `CubicBool=True, vFuncBool=True` solve then `deepcopy` of the solved agent | OK |

Positive control, same environment with `origin/main`'s unpatched `interpolation.py` swapped in: **15 failed, 114 passed**. The failures are not confined to serialization. `testBufferStock`, `testIndShockConsumerTypeExample`, `testIndShockConsumerTypeLifecycle`, `testIndShockConsumerTypeCyclical`, `testStablePoints` and `test_Harmenbergs_method` all fail with `TypeError: cannot pickle 'module' object`, traced to the unconditional `deepcopy` in `ValueFuncCRRA.__init__` at `interpolation.py:3539`. So under 1.18.0 the old code breaks ordinary `CubicBool=True` solves, not only explicit serialization, which makes the exclusion a bigger thing to be carrying quietly.

On safety: `interp1d` is the only other `scipy.interpolate` construct anywhere in HARK. Its three call sites in `utilities.py` are local variables; `kernel_regression` returns one but has no callers inside `HARK/`, so it never reaches a deepcopy or pickle path. Directly tested under 1.18.0, `interp1d` has no module-object attributes and both deepcopy and pickle succeed. `CubicHermiteSpline` is the only PPoly-family object HARK stores, and this PR fixes it.

## Non-blocking

**Pickle compatibility is one-directional, and I added a CHANGELOG note.** A pickle written by old code loads under the new code, since `__setstate__` just overwrites `_chs` with the deterministic rebuild. The reverse fails: a new-format pickle under old code gives `AttributeError: 'CubicHermiteInterp' object has no attribute '_chs'`. That is a normal consequence of the change, but HARK users pickle solved models and may unpickle them in a differently-versioned environment, so it is worth stating.

**Post-round-trip coverage gap.** `roots`, `solve`, `integrate`, `antider_interp` and `der_interp` all delegate straight to `self._chs` but are not exercised on the clone after a round trip. Low risk given `_chs.c` is verified bit-identical and the reconstruction is deterministic, but one more line would close it.

Good to merge from my side once you are happy with the pin removal riding along.

alanlujan91 (approved):
Fix verified: round trip bit-identical across 3 configurations and 5001 probe points, values and derivatives and `_chs.c`, under both deepcopy and pickle. Positive control under scipy 1.18.0 with main's unpatched `interpolation.py` fails 15 tests including ordinary `CubicBool=True` solves, so this was masking a real breakage rather than only a serialization concern.

I pushed two commits: removing the now-redundant `!=1.18.0` exclusion (which was keeping CI from ever exercising the new tests), and a merge of current main.

PR #1811, opened 2026-08-12
Declare public API of recently added modules (__all__ + normalization docs section)

llorracc:
Polish follow-up to the recently merged board PRs, applying the same public-API hygiene #1795 established for ConsAggShockModel:

- `__all__` for `HARK.simulation.normalization` (#1784) and `HARK.ConsumptionSaving.ConsAggIndMarkovModel` (#1796) — both names verified importable.
- An API-reference section for `HARK.simulation.normalization` appended to the existing Simulation tools page (the module merged without one; the toctree already lists the page, so no index change).
- One changelog bullet.

Notes: `tests/test_normalization.py` + `tests/ConsumptionSaving/test_hierarchical_mrkv_helpers.py` green locally. When #1798 (the hierarchical rewrite) merges, its branch will carry the `__all__` updated to the renamed class so the merge stays clean.


alanlujan91:
Forwarded onto current main. One thing needed fixing, and it is worth knowing about because `__all__` fails in a particular way.

**This branch's `__all__` named `AggIndMarkovConsumerType`, which no longer exists.** #1798 renamed it to `AggIndMrkvConsumerType`, and that landed on main earlier today, so `from HARK.ConsumptionSaving.ConsAggIndMarkovModel import *` would have raised `AttributeError`. You had already added the correct version to #1798 (`fc2f9e49`), so main's copy was right and this branch carried the stale one; I resolved the conflict to main's.

That is the awkward property of `__all__`: it is a list of strings, so a name that no longer exists is invisible to the import machinery, the linter and the test suite until someone does a star-import. Verified after resolving, by importing each module and checking every entry resolves:

```
ConsAggIndMarkovModel  __all__=['make_hierarchical_mrkv_array', 'extract_cond_mrkv_arrays', 'AggIndMrkvConsumerType']   missing: none
normalization          __all__=['ShockNormalizationMixin', 'PermanentIncomeNormalizationMixin']                         missing: none
```

Also added the blank line ruff 0.11.8 wants after the `__all__` block in `normalization.py`. Note CI formats the whole repo rather than the changed files, so `uvx ruff@0.11.8 format --check $(git ls-files '*.py')` is the check that matches; a current local ruff formats differently.

26 tests pass, lint clean.

alanlujan91 (approved):
Approving. Additive: `__all__` for two modules plus a docs API-reference section. Fixed the stale class name (`AggIndMarkovConsumerType` -> `AggIndMrkvConsumerType`) and verified every `__all__` entry resolves on import.

PR #1822, opened 2026-09-10
Fix the income-weighted (Harmenberg) transition matrices under mortality with permanent-income growth

llorracc:
# Fix the income-weighted (Harmenberg) transition matrices under mortality with permanent-income growth

## What was wrong

Two tools build transition matrices under the permanent-income-neutral measure: the model-file Simulator (`AgentSimulator.make_transition_matrices`, used by `find_steady_state`, `simulate_shock_by_grids` and `make_basic_SSJ`) and `NewKeynesianConsumerType.calc_transition_matrix`. Both reweight survivors' transitions by the permanent shock alone and re-inject newborns at the population death rate. That is Harmenberg's construction for a model without death; with death *and* permanent-income growth it is not the income-weighted measure: newborns arrive at a fixed level while survivors' incomes keep growing, so the stationary distribution overweights the young at every age, and every normalized aggregate built from it (long-run averages, impulse responses, sequence-space Jacobians) is biased. The construction is exact when `PermGroFac == 1` or `LivPrb == 1`, which is why the defect is invisible in most tests; the `IndShockConsumerType` defaults (`LivPrb = 0.98`, `PermGroFac = 1.01`) that the SSJ tutorial uses are inside the biased region.

The size, income-weighted consumption and assets, pre-fix minus exact:

| model | consumption | assets |
|---|---|---|
| `IndShockConsumerType` defaults (the tutorial) | −0.5 % | −3.0 % |
| two-state Markov consumer with a zero-income unemployment state, G = 1.01, L = 0.98 | −3.4 % | −7.2 % |
| a very patient Markov consumer (growth-patience factor 0.9995, growth in employment only) | −12 % | −30 % |

Ground truth for the second row is a Monte Carlo with HARK's own simulator weighting each agent by its permanent income: the corrected chain is within 0.13 % (consumption) and 0.11 % (assets) of it; the previous one is off by the table's amounts.

`NewKeynesianConsumerType.calc_transition_matrix` also omitted `PermGroFac` from the normalized transition itself (`m' = R a / psi + theta` instead of `R a / (G psi) + theta`); invisible at its defaults (`PermGroFac = 1`), wrong for any user who sets growth.

## The fix

The weight on a transition is the realized growth of the *level* being normalized by — `G = PermGroFac * PermShk` in the model files — relative to the growth newborns inherit, and newborns enter at the complement of the surviving weighted mass (the row deficit), which is what makes the stationary vector of the chain the income-weighted distribution. Concretely:

- `SimBlock.make_transition_matrices` now weights each blob's probability mass by the realized value of `norm` *after* the block has run, so `norm` may be any realized variable (`'G'`), not only a drawn shock. Weighting by `'PermShk'` gives exactly the previous numbers.
- `AgentSimulator.make_transition_matrices` gains `newborn_growth` (default 1.0): the per-period growth of the normalizing level that newborns inherit (HARK's `PermGroFacAgg`); under `norm` with mortality the replacement block uses the weighted surviving mass and its complement, stored per period in `newborn_shares`, and raises when the surviving mass exceeds one (growth outpaces mortality: no finite mean level, no stationary income-weighted distribution).
- Outcome reads under `norm` — `get_long_run_dstn`, `simulate_shock_by_grids`, `simulate_cohort_by_grids`, and the scaling of `make_basic_SSJ` / `calc_shock_response_manually` — are normalized by the mass growth within the period, so results are per unit of the period's (post-growth) level, the Harmenberg aggregate. `make_basic_SSJ` and `calc_shock_response_manually` take `newborn_growth` too, defaulting to the agent's `PermGroFacAgg`.
- `gen_tran_matrix_1D/2D` take `PermGroFac` (in the transition) and, for the 1D neutral-measure case, `surv_growth = PermGroFac / PermGroFacAgg`; `NewKeynesianConsumerType.calc_transition_matrix` supplies both and raises in the no-finite-mean case.

**The trend convention is kept as an option.** If all growth is an aggregate trend that newborns inherit, weighting by the shock alone *is* exact; that is `PermGroFacAgg = PermGroFac` (or `newborn_growth = PermGroFac`), under which the new code reproduces the previous matrices bit for bit. HARK's Monte Carlo simulator already implements that convention through `PlvlAgg`, so the grid tools are now consistent with it under either convention. The default (`PermGroFacAgg = 1`) is HARK's documented semantics: individual growth, fixed newborn income.

## What changes for users

- Results change only when survival is below one *and* growth differs from one, and only under `norm` (or `neutral_measure=True`). `norm='PermShk'` returns what it returned before; `norm='G'` is the corrected recipe and is what the SSJ notebooks now use.
- `NewKeynesianConsumerType` at its defaults is unchanged (its existing tests pass untouched).
- New: `newborn_growth` kwarg (Simulator, `make_basic_SSJ`, `calc_shock_response_manually`), `AgentSimulator.newborn_shares`, `PermGroFac`/`surv_growth` arguments on the matrix generators (defaults preserve the old signatures' behavior).

## Tests

`tests/test_simulate.py::testsForIncomeWeightedMeasure`: the newborns' share of the stationary income-weighted mass equals `1 − LivPrb·PermGroFac` (grid-independent closed form); `norm='G'` ≡ `norm='PermShk'` when `PermGroFac = 1`; `newborn_growth = PermGroFac` reproduces the shock-only weighting bit for bit; long-run outcome distributions are stochastic vectors under `norm`; the income-weighted mean matches a Monte Carlo and is closer to it than the previous construction; the no-finite-mean case raises; `make_basic_SSJ` runs under `norm='G'` and equals the `'PermShk'` result without growth.
`tests/ConsumptionSaving/test_ConsNewKeynesianModel.py::test_NeutralMeasureWithGrowth`: the NK matrix agrees with the Simulator on the same primitives with growth (consumption to 5e-4 — the residual is the class's placement of newborns at `m = 1` rather than at their first income draw, unchanged here), the trend option differs from the fix, columns sum to one, the error path raises.

## Notebooks

`SSJ-tutorial.ipynb` and `SSJ-advanced-examples.ipynb` pass `norm="G"`; the tutorial's description of `norm` and its Harmenberg section say why. In the advanced notebook the substitution `UnempDynType.IncShkDstn[0][0] = UnemployedIncDstn` is moved after `update()`, which rebuilds `IncShkDstn` from its constructor and was silently undoing it.

## Known limits, deliberately not in this PR

- A shock to the employment chain itself changes the income mass over time; under `norm` the fake-news derivatives are exact for every shock that leaves the income process alone (interest factor, discount factor, transfers, wages, taxes) and need an income-mass recurrence for chain shocks. Documented, not implemented here.
- `make_basic_SSJ` differences the shocked finite-horizon path against the steady state; for very patient households the steady-state solve's residual at the default tolerance enters the far-horizon columns. A separate PR will propose an unperturbed ghost run.
- `NewKeynesianConsumerType` places neutral-measure newborns at `m = 1`; its 2-D branch places them at their first income draw. Left as is; noted in the test.

Found while building a HAFiscal household as a `MarkovConsumerType` twin and comparing this tool's output with our own transition-matrix code, which had made the same correction earlier this summer; the two now agree to 1e-4 on both a median and a growth-impatience-cap household.

---

https://claude.ai/code/session_01A2RFCLdEBybCs4hRCae1HT


alanlujan91:
Pushed 3a415e4c on top of this branch, a cleanup pass with no intended change in results.

- Removed the draw-time Harmenberg path that moving the weighting after the quasi-sim left unused: `RandomEvent._apply_harmenberg` and the `norm` argument of every `quasi_run` and of `SimBlock.run_quasi_sim`. Nothing passes `norm` to them any more.
- Merged the weighted and unweighted death/newborn mixing in `AgentSimulator.make_transition_matrices` into one path (this also drops a K x K `np.tile`), and removed `self.newborn_growth`, which nothing reads.
- The `newborn_growth is None -> PermGroFacAgg` default now lives once, in `_prepare_ssj_computation`, instead of in both `make_basic_SSJ_matrices` and `calc_shock_response_manually`.
- `gen_tran_matrix_1D` takes the surviving mass through `LivPrb` (the NK caller passes `LivPrb * surv_growth`), so the extra `surv_growth` argument is gone. `gen_tran_matrix_2D` computes `mNext_ij` once per `i` rather than once per `(i, j)`.
- `_neutral_measure_surv_growth` reads `self.neutral_measure` directly, since the default parameters always set it.
- The two new test classes import at module top and solve their agents once per class.

Checked against the previous head d0e23fd7: the transition matrices are bit-identical (simulator with no `norm`, `norm='PermShk'`, `norm='G'`, and fake-news timing; NK 1D and 2D kernels). Steady states, long-run averages, the SSJ and `calc_impulse_response_manually` agree to within the run-to-run noise of `find_steady_state` itself. Two runs of identical code differ by up to about 2e-14 (about 3e-10 for the impulse response). Full test suite passes (896).

Two things for you:

1. #1823, #1824, #1826, #1827 and #1828 each carry this PR's original commits, so each will need a rebase onto the new head.
2. `make_flat_LC_SSJ_matrices` still calls `make_transition_matrices(grids, norm)` without `newborn_growth`, reads `periods[t].matrices` without the mass-growth division that `get_long_run_dstn`, the grid simulators and `make_basic_SSJ_matrices` now apply, and keeps the old `norm` docstring alongside its own `trend` factor. Is the lifecycle SSJ under the new weighting meant to be covered by this PR, or left for a follow-up?

llorracc:
Thanks for the cleanup, Alan. Both points are being implemented now.

1. The five stacked PRs (#1823, #1824, #1826, #1827, #1828) will be rebased onto this branch's new head once point 2 lands here, so they pick up both commits at once.

2. `make_flat_LC_SSJ_matrices` is being brought under the same weighting inside this PR: each birth cohort's income-weighted mass is propagated age by age with the per-state surviving mass (`dead[:, 0]`, as `make_transition_matrices` reads it), the shocked masses enter the news term so a shock to the income process moves the levels, `newborn_growth` is passed through, every period is built with `for_t` so the cohort model skips the population-level replacement (and its stationarity check, which a life-cycle period with `LivPrb * PermGroFac > 1` trips), and the Jacobians are expressed per unit of the newborn cohort's first-period level. With that, `norm='G'` with `trend=None` and `norm='PermShk'` with `trend='PermGroFac'` give the same Jacobians for shocks that leave the income process alone (4e-12 on a short life cycle), and the latter is unchanged against the pinned 0.17.2 to 2e-12. Tests and the changelog line follow in the commit.

One choice to flag for your reaction: in the life-cycle helper `newborn_growth` defaults to 1.0 rather than the agent's `PermGroFacAgg`, because with flat demographics and `prod_gro` not yet implemented that is the consistent value, and `init_lifecycle` carries `PermGroFacAgg = 1.016` as a leftover. Happy to change it if you prefer the same resolution rule as the infinite-horizon functions.

llorracc:
Both points are done.

Point 2 landed here as a19e6b3a (on top of your 3a415e4c): `make_flat_LC_SSJ_matrices` propagates each cohort with the per-state surviving mass, uses the shocked masses in the news term, passes `newborn_growth` (default 1.0, see the previous comment), builds every period with `for_t` so the cohort model skips the population replacement and its stationarity check, and expresses the Jacobians per unit of the newborn cohort's first-period level. Three tests in `tests/test_simulate.py` (`testsForLifeCycleWeightedMeasure`): cohort masses match their closed form under both weightings; `norm='G'` alone agrees with `norm='PermShk'` plus `trend='PermGroFac'` on an Rfree shock; a PermGroFac shock moves the masses so the two differ. `norm='PermShk'` with the trend is unchanged against 0.17.2 to 2e-12, and the lifecycle notebook uses that pair, so its outputs stand. Changelog line extended.

Point 1: the five stacked branches are rebased onto a19e6b3a and force-pushed: #1823 1fcb7651, #1824 70609037, #1826 4e613914 (on #1823), #1827 7b2ddf0a, #1828 4166be2c (on #1826). Each branch's own tests plus `tests/test_simulate.py` pass on the new base (39 to 48 tests per branch); the only conflicts were the additive kind (a parameter, a docstring entry and a test class appended at the same spots) and were resolved by keeping both sides. CI is running on all six.

alanlujan91 (approved):

alanlujan91:
Thanks, Chris, both points are in. Against main, the life-cycle Jacobians from a19e6b3a under `norm='PermShk'` with `trend='PermGroFac'`, and under `norm=None`, agree to within 2e-12. `tests/test_simulate.py` passes.

Pushed 8cdd5b5b on top, a cleanup pass with no intended change in results:

- `make_flat_LC_SSJ_matrices` asks `simulate_cohort_by_grids` for `dead` only. `SS_dstn` now comes from `_lc_cohort_dstns`, so the per-outcome distributions (`calc_dstn=True`) were computed and discarded.
- The expectation-vector loop scales the product by the surviving mass instead of forming `LR_surv[a][:, None] * LR_trans[a]`, a K x K temporary per step.
- Three comments trimmed.

Against a19e6b3a the life-cycle Jacobians (`norm` None, `'PermShk'` with the trend, and `'G'`; `Rfree` and `PermGroFac` shocks) differ by at most 2.2e-16. Full suite passes (904).

On `newborn_growth`: 1.0 is the right default, since `make_flat_LC_SSJ_matrices` already refuses `prod_gro != 1`, and that is the same cohort trend. One correction to the reasoning, though: `init_lifecycle`'s `PermGroFacAgg = 1.016` is not a leftover. It is the calibrated yearly trend (`IncomeTools` sets it from `YearGroFac`), and `sim_birth` applies it to newborns through `PlvlAgg`, so the Monte Carlo simulation of the same agent does have a cohort trend, and this Jacobian silently assumes it away. As a follow-up, could the helper warn when the agent's `PermGroFacAgg != 1` and `newborn_growth` was not passed explicitly? Then anyone comparing it with a simulation of the same agent knows the two differ.

Two smaller things for later, if you agree. `_lc_surviving_mass` reads the column layout of `matrices["dead"]`, which `make_transition_matrices` could expose directly as the per-period surviving mass. The age-0 normalizer at the end takes its value from the first outcome array, which works only because every outcome array has the same row sums.

Approving and merging this one.

PR #1831, opened 2026-09-11
Refactor the dual-measure, pLvl normalization and hierarchical Markov code

alanlujan91:
This refactors `HARK/dual_measure.py`, `HARK/simulation/normalization.py` and the hierarchical branch of `AggIndMrkvConsumerType.get_markov_states`, and fixes `Q_degenerate_periods` for Markov agents. Simulated output does not change. I hashed every `history` and `history_Q` array against `main` (160c3b65) for nine setups: dual IndShock (infinite horizon and lifecycle), dual Markov, three normalization runs (shocks alone, then shocks with pLvl in both infinite horizon and lifecycle), and `AggIndMrkvConsumerType` with shuffled, iid and global Markov draws. All nine match, and shifting one Q-shock index changes the dual hashes, so the comparison sees a one-element change.

### `dual_measure.py`

- `_draw_Q_shocks_indshock` and `_draw_Q_shocks_markov` each wrote the same cell draw twice, once in the main loop and once for newborns. One helper, `_fill_Q_cell`, now does all four. The loops keep their order, so each distribution's `draw_events` calls and the RNG stream are unchanged.
- The mixin overrode `sim_one_period` to put the Q step before time advances, which meant copying the base pipeline. It now overrides `_sim_period_epilogue`, where `t_age` and `t_cycle` still hold the values the P step used. It also drops the `post_state_hook` override that only called `super()`.
- The `_cdf_invert` alias existed only for one test import, which now takes `cdf_invert` from `HARK.distributions.discrete`. The module also drops two guards that could never fail (`N > 0` and `np.any(idx)` on masks built from `np.unique(self.t_cycle)`) and a `None` check that the `isinstance` beside it already covers. `clear_history_Q`, `_record_Q_history` and the `IncShkDstn_Q` builder are shorter, and `_first_scalar` replaces the period-0 lookup repeated in `compute_mean_pLvl` and `compute_pLvl_factor`.
- `setup_Q_measure` never recorded a Markov period in `Q_degenerate_periods`. It tested `p is q` on each period's entry. For a Markov agent that entry is a list of per-state distributions, which the Q side rebuilds as a new list. A Markov period now counts when every state's distribution is degenerate. `_Q_equals_P` holds that rule, and `_any_reweighting_happened` now reads the resulting list instead of repeating the test.
- Because the Q step now lives in the epilogue, a subclass whose own `sim_one_period` never calls `_sim_period_epilogue` would run dual mode with `history_Q` all NaN (llorracc's review). Dual-mode `simulate` clears a flag before each period, the epilogue sets it, and `simulate` raises `NotImplementedError` on the first period where it is still clear. This is a runtime check, where the review suggested a static `co_names` check at setup, because a check that only reads names would refuse an override that delegates through `super().sim_one_period()`, and would miss a call made through a helper. It does not touch the draw order.

### `simulation/normalization.py`

- `_analytical_log_pLvl_moments` walked each age's income history from birth, for every age present, in every period, so the cost of normalizing grew with the cube of the simulation's length. It now caches each age's result until `initialize_sim` clears the cache.
- `_log_moments` replaces the log-moment computation written twice in `_permshk_log_moments`. The module also drops the `idx=None` fallback nothing reached, the redundant `getattr` reads and the repeated `T_cycle` expression. A comment that referred to a function that does not exist now refers to `_refuse_normalization`.

### `ConsAggIndMarkovModel.py`

The hierarchical branch of `get_markov_states` combines the macro and micro states once, keeps the states of newborns (and of everyone in the first period) with one mask, and sets `MacroMrkvNow` and `MicroMrkvNow` through `_split_mrkv`. I left alone the part of the file that #1815 rewrites (`get_micro_markov_states` and the `CondMrkvArrays` format checks).

### Tests and changelog

`test_markov_degenerate_periods_are_recorded` builds a three-period Markov agent whose periods have both, one or neither state degenerate, and requires `Q_degenerate_periods == [0]`. It fails on `main`, which returns `[]`. `test_dual_simulate_refuses_a_pipeline_that_skips_the_epilogue` requires the new error from an override that skips the epilogue, and finite `history_Q` from one that delegates through `super()`; it fails with the guard removed. Imports in `test_dual_measure.py` move to the top of the module, `test_normalization.py` imports `DualMeasureMixin` directly in place of `pytest.importorskip`, and two stale docstrings in that file now match the code. The changelog entry about `sim_one_period` now describes the epilogue override, the entry on `Q_degenerate_periods` states the Markov rule, and a new entry covers the pLvl moment cache and the new epilogue guard.

The branch is merged with `main` at 251b4807 (#1832). The full suite passes (891 tests), and pre-commit is clean.


llorracc (approved):
Read the whole diff and ran it here. Approving; one optional suggestion and one note on our side.

**By reading.** Each rewrite keeps the arithmetic and the order of the random draws. `_fill_Q_cell` issues the same `cdf_invert` / `draw_events` calls in the same loop order as the four blocks it replaces (the `np.any` guards survive where a cell can be empty and were redundant where `np.unique` supplies the keys). The hierarchical `get_markov_states` builds the same combined state and keeps the same agents' previous state: newborns, and everyone at `t_sim == 0`. `_Q_equals_P` gives `_any_reweighting_happened` its old answer and makes `Q_degenerate_periods` count a Markov period only when every state is unchanged, which is the fix the changelog describes. The age cache in `_analytical_log_pLvl_moments` is keyed on the only thing the walk depends on and is cleared where the step cache already was. The epilogue override is equivalent for every class that reaches `AgentType.sim_one_period`; the one in-tree `AgentType` with a pipeline of its own, `RiskyContribConsumerType`, is already refused by `setup_Q_measure` through the `get_Rport` check.

**By running.**
- Full suite at d9733aaf, Python 3.12 / numpy 2.5.3: 889 passed.
- The three touched test files also pass on Python 3.11 / numpy 1.26, the floor HAFiscal's fork lineage keeps: 65 passed.
- `test_markov_degenerate_periods_are_recorded` fails on the merge base 160c3b65 with `[] == [0]` and passes here, so it catches what it claims.
- ruff 0.11.8 check and format are clean on the five files.
- HAFiscal's own suites for the dual measure, the normalization mixin and the hierarchical state encodings (48 tests; HAFiscal's classes composed on these mixins), run in HAFiscal's environment under 160c3b65 and under d9733aaf: 48 passed under each.
- A bit-identity probe of HAFiscal's four agent classes (the production `AggFiscalType`, `DualAggFiscalType`, and both compositions with the normalization mixin), each solved and simulated under 160c3b65 and under d9733aaf: all 42 P and Q history digests identical. Nothing HAFiscal's paths produce moves.

**Optional.** With the Q step in `_sim_period_epilogue`, an outside subclass that overrides `sim_one_period` without calling the epilogue now runs with `dual_measure=True` and a `history_Q` full of NaN, where before the mixin's own `sim_one_period` silently replaced that subclass's pipeline. Both fail quietly; the new one is the easier to make loud: a check in `setup_Q_measure` in the style of `_warn_if_hook_unreachable` (`"_sim_period_epilogue" in type(self).sim_one_period.__code__.co_names`) would refuse at setup. Nothing in-tree needs it today.

**On our side.** HAFiscal's dual-MC path imported the private `_cdf_invert` alias; it now imports `HARK.distributions.discrete.cdf_invert`, which the pin already has. No change requested here.

PR #1835, opened 2026-09-13
Add check_flow_budget and aggregate_SSJs beside make_basic_SSJ: the household flow-budget identity as the acceptance test, and aggregation over types

llorracc:
**Stacked on #1823** (whose chain is #1822 -> #1823; review the last commit only). It adds the two utilities the ghost run's users need next: the identity that says whether a set of Jacobians is right, and the aggregator over types.

A household block's sequence space Jacobians have to obey the household's own budget, and nothing else in the toolkit says whether they do. This PR adds the check beside `make_basic_SSJ`, and the aggregator that a model with several types needs next.

**`check_flow_budget`** (with `flow_budget_residuals`, which returns the residual matrix). In the units `make_basic_SSJ` reports (responses per unit of the period's normalizing level), the consumption, end-of-period asset and labor income responses to a perturbation dated `s` satisfy, at every date `t`,

```
dC[t, s] + dA[t, s] - rho * dA[t-1, s] - dY[t, s] = cash[t, s],    rho = Rfree * LivPrb / newborn_growth,
```

with `dA[-1, s] = 0`: survivors bring `Rfree` times their end-of-period assets into the next period, the dead bring nothing (their replacements' initial assets belong to the steady state and do not respond), and the normalizing level grows by `newborn_growth` per period (1.0 by default, `PermGroFac` under the trend convention). The right side is what the perturbation itself injects: nothing for a perturbation that reaches the household through labor income (with `dY` counted) or through preferences; for a perturbation of the return factor, the return on the assets brought into the period, `LivPrb * A_ss / newborn_growth`, on the date the perturbed return is paid, and nothing elsewhere. The check tests every column and raises a `ValueError` naming the first date and column that fail. Without the income Jacobian it falls back to the weaker check that needs no oracle at all: every column delivers only on its own date, and the same amount in every column (the same experiment, moved in time).

Two things make it more than a unit test. It is the acceptance test for *any* construction of these Jacobians, HARK's or a user's, since it appeals to no construction as the reference. And it settles dating questions by measurement: the interest-factor Jacobians from `make_basic_SSJ` with `offset=True` pass only with the return term on row `s` of column `s`, so the column dated `s` is the return paid at the start of period `s`; put the term one row later and the check names row 0 of column 0 at once. The factor `rho` is measured too, not assumed: with `PermGroFac = 1.01` the identity closes to 1e-11 with `Rfree * LivPrb` under the default convention and with `Rfree * LivPrb / 1.01` under the trend convention (`newborn_growth = 1.01`), and fails at 3e-2 with either factor in the other convention.

**`aggregate_SSJs`** adds Jacobians over types as a weighted sum, in whatever structure `make_basic_SSJ` returned them (an array, a list or a dict per type), with the weights' meaning stated: each type's share of the population's normalizing level, which is its population share when the types' newborns start at the same level and share the same survival and growth, and its population share times its relative mean level otherwise. The identity is linear, so the aggregate obeys it with the same `rho` and the aggregated `A_ss`; the tests check that.

**Measured.** On the HANK income process (`wage` a perturbable parameter), tolerance 1e-10, `T_max = 60`: the wage identity closes to 6e-12 of the Jacobians' scale with the ghost run and without it (it is an accounting identity, so the long run solve's residual does not enter), and the income Jacobian is exactly the unit delivery on the diagonal; the interest-factor column delivers `LivPrb * A_ss` to 1.5e-7, the discretization of the assets-to-capital transition (the residual equals the assets brought in, `sum(D * kNrm)`, to 1e-12); a patient type (`DiscFac 0.97`) closes to 2e-7 at the far end of column 0, the fake-news assembly's own linearization error at the finite-difference step, which is why the check's default tolerance is 1e-6 and the return-factor case is documented to need a tolerance for the grid.

**Tests** (`tests/test_simulate.py`, six): constructed matrices with growth and a trend (1e-13; parameter and shape validation); the wage perturbation closes with and without the ghost run, its income Jacobian is the delivery, the pattern check agrees; the interest-factor perturbation delivers the assets brought in on its own date, the exact check without the return term refuses it and the pattern check needs neither; the survivor factor under growth with and without the trend, and the wrong factor refused; a broken column is named; aggregation over two types in list, dict and array forms, linear in the identity for both shocks, with the length, structure and shape checks. Full suite 909 passed. **Notebook:** `SSJ-advanced-examples.ipynb` gains a section after the HANK income-process example, four types (two risk groups, two discount-factor atoms each) aggregated with population weights, the checks shown type by type and on the aggregate, the omitted return term caught; re-executed (the new cells take about twenty seconds).

This is the E2 item of the HAFiscal project's Jacobian card, whose Step 4 runs this identity as a gate on every column of its household block (it caught a mis-dated column 0 there); the check is written so that it applies to `make_basic_SSJ`'s output as it is.


llorracc:
The examples job on this PR has failed six times out of six since it was opened, each time about eight minutes into `Run examples` with exit code 143 and "The runner has received a shutdown signal", which is the hosted VM running out of memory. Main's own runs and the other open PRs pass. I traced it. The PR's own cells are not the cause, but the failure is real and deterministic, so here is what I found.

**Mechanism.** `pytest-xdist` 3.8.0's `loadscope` scheduler sorts its work queue by descending number of tests per scope (`xdist/scheduler/loadscope.py`, `schedule()`: `sorted(..., key=lambda item: -len(item[1]))`), and for nbval every cell is a test. So the two workers take the notebooks with the most cells first: LabeledModels (46), Advanced-Intro (38), Quick_start_with_solution (38), Transition_Matrix_Example (35), HANKFiscal_example (31), Simulation-Intro (28), and then this PR's 27-cell SSJ-advanced-examples, where main's 23-cell version comes twelfth. The four added cells move the notebook from twelfth to seventh in the queue, into the window in which Transition_Matrix_Example is still running on the other worker.

**Measured** with the workflow's own command (`pytest --nbval-lax --nbval-current-env --dist loadscope -n 2 examples/`) inside a memory-capped cgroup on a 60 GB Linux box, Python 3.12, a memory sample every 5 s:

| tree | peak memory of the two-worker job | running at the peak | result |
|---|---|---|---|
| this PR (ab15b868) | 21.5 GiB | SSJ-advanced-examples beside Transition_Matrix_Example | 855 passed in 18 min given 26 GiB; OOM-killed under a 15 GiB cap at 213 cells (CI dies at 227-232) |
| main (d741d378) | 16.8 GiB | KS-HARK-presentation beside Transition_Matrix_Example | 851 passed in 18 min; also OOM-killed under 15 GiB, later, at 270 cells |

Single notebooks, one at a time (max RSS): Transition_Matrix_Example 15.8 GiB; SSJ-advanced-examples 6.8 GiB on both trees, this PR's four cells adding about 35 MB at the peak; HANKFiscal_example 3.7 GiB; KS-HARK-presentation 0.7 GiB. `ubuntu-latest` has 16 GB of RAM and a 4 GB swap file, so main's 16.8 GiB window survives on swap and this PR's 21.5 GiB window cannot. Main is one scheduling shift from the same failure: any PR that adds cells to a large notebook, or adds a notebook of 24 or more cells, changes the pairing, and the transition-matrix notebook alone already needs the swap.

Two observations on where the memory is, in case they help. In SSJ-advanced-examples about 5 GiB of what stays resident after the heavy cells is freed memory that glibc keeps: with `MALLOC_MMAP_THRESHOLD_=131072` the footprint after the Markov cell drops from 5.5 GiB to 0.5 GiB, at the price of the heavy cells running 2.5 to 5 times slower. The live transient of that Markov cell is 6.6 GiB either way: the `T_max` dense transition arrays the fake-news construction holds at once (300 x 802 x 802 doubles per set).

**Remedies**, all outside this PR's cells, your call which: (a) run the `SequenceSpaceJacobians` notebooks serially in `examples.yml` (a second job, or `-n 1` for that directory), which removes the pairing for good at the cost of a few minutes of wall time; (b) lighter heavy cells in Transition_Matrix_Example and in the advanced notebook's Markov and portfolio sections (smaller `T_max` or grids; the plots show t <= 150 anyway); (c) sparse or streamed finite-horizon transition arrays in `make_basic_SSJ`, the structural fix and its own PR. I can push (a) as a commit here or as a separate small PR, whichever you prefer; nothing in this PR has been changed for it.

